PR #17, opened 2025-07-01
feat: document tracking choice - appsync or ddb

rpawlaszek:
*Description of changes:*
This PR modified `idp_common` in that it now has a factory for using a service to track document processing progress. Before the only option was to use `appsync` module which wrote to the DynamoDB TrackingTable indirectly. With this PR there are two modules `appsync` and `dynamodb` that expose service classes which share the same usage interface. With that, there is added a factory that by default creates an `appsync` module originating service to track document processing progress, but can also create `dynamodb` module originating service. The `dynamodb` choice requires `TRACKING_TABLE` environment variable to be present to instrument the service where to write updates.

This PR does not change the solution behavior, but introduces the change to `idp_common` as well as all AWS Lambda function code to rely on the factory rather than specific (i.e. `appsync`) service implementation. This is a building block that is to be used to allow users a choice for writing document processing progress information.

NOTE: if `dynamodb` is to be selected this means that the change is not going through an AppSync/GraphQL mutation which means that the subscription in the UI would not be triggered. Hence, if `dynamodb` module was selected - the UI will either not have real-time notifications or the real-time notification would need to be recreated.

### DIFF
--- a/lib/idp_common_pkg/idp_common/__init__.py
+++ b/lib/idp_common_pkg/idp_common/__init__.py
@@ -13,6 +13,9 @@ def __getattr__(name):
     if name in [
         "bedrock",
         "s3",
+        "dynamodb",
+        "appsync",
+        "docs_service",
         "metrics",
         "image",
         "utils",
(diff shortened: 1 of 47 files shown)


rstrahan:
Thanks Rafal - merging it now to develop branch, and it'll be in the next release.

PR #31, opened 2025-07-30
New groupConsecutivePages config and raise lambda memory

markjayne-amzn:
… 

*Issue #, if available:*

*Description of changes:*
add new groupConsecutivePages config option that defaults to true and increase lambda memory to avoid fail.  If groupConsecutivePages is false and classificationMethod == multimodalPageLevelClassification then every page will get its own section.

### DIFF
--- a/lib/idp_common_pkg/idp_common/classification/service.py
+++ b/lib/idp_common_pkg/idp_common/classification/service.py
@@ -129,6 +129,14 @@ class ClassificationService:
             "classificationMethod", self.MULTIMODAL_PAGE_LEVEL
         )
 
+        # Get grouping behavior from config (defaults to True for backward compatibility)
+        # Handle both boolean and string values due to UI serialization issues
+        group_pages = classification_config.get("groupConsecutivePages", True)
+        if isinstance(group_pages, str):
+            self.group_consecutive_pages = group_pages.lower() not in ('false', '0', 'no', 'off')
+        else:
+            self.group_consecutive_pages = bool(group_pages)
+
         # Log classification method
         if self.classification_method == self.TEXTBASED_HOLISTIC:
             logger.info("Using textbased holistic packet classification method")
@@ -1360,7 +1368,7 @@ class ClassificationService:
                 current_pages = [sorted_results[0]]
 
                 for result in sorted_results[1:]:
-                    if result.classification.doc_type == current_type:
+                    if result.classification.doc_type == current_type and self.group_consecutive_pages:
                         current_pages.append(result)
                     else:
                         # Create a new section with the current group of pages
--- a/patterns/pattern-2/template.yaml
+++ b/patterns/pattern-2/template.yaml
@@ -851,7 +851,7 @@ Resources:
       Handler: index.handler
       Runtime: python3.12
       Timeout: 900
-      MemorySize: 4096
+      MemorySize: 8192
       Environment:
         Variables:
           METRIC_NAMESPACE: !Ref StackName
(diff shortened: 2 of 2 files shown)


rstrahan (changes requested):
Thanks for the PR!

This seems to address two separate (unrelated?) issues... (1) setting to control grouping of consecutive pages of the same class for the pageLevelClassification method, and (2) Lambda memory settings.

Ideally please submit these as two separate PRs, as they are unrelated.  I think it's OK this time as the changes are small.

Additional comments / requests:

### (1) setting to control grouping of consecutive pages of the same class for the pageLevelClassification method

1. The PR should include related documentation updates.. please add for review.

2. The PR should include this new setting in the `UpdateSchemaConfig` resource in the pattern-2 template.yaml. Needs to be added to enable setting in the View/Edit Configuration UI page.

3. The new setting `groupConsecutivePages` should be `groupConsecutivePagesOfSameClass` or something similar that reflects the behavior more precisely.

4. This setting is applicable only if `classificationMethod` is set to `multimodalPageLevelClassification` - When you add it to the schema config, make it conditional on the `classificationMethod` value. See, for example, the Textract 'features' property which renders only when ocr backend value is textract - [code here](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/main/patterns/pattern-2/template.yaml#L225-L243)  

### (2) Lambda memory settings.

Please provide evidence in the form of logs, or a link to a reported issue that reflects details of the bug that you are attempting to fix with these changes.   Also, please include why the memory needs to change for each of the modified lambda functions - even the functions that were already set to 4096 - to address any concern that this is overkill and could be addressed by a more targeted approach.

Thanks!

rstrahan:
@mmiakashs  - Can you review the part 1 of this PR - My earlier comments / requests were:

### (1) setting to control grouping of consecutive pages of the same class for the pageLevelClassification method
The PR should include related documentation updates.. please add for review.

The PR should include this new setting in the UpdateSchemaConfig resource in the pattern-2 template.yaml. Needs to be added to enable setting in the View/Edit Configuration UI page.

The new setting groupConsecutivePages should be groupConsecutivePagesOfSameClass or something similar that reflects the behavior more precisely.

This setting is applicable only if classificationMethod is set to multimodalPageLevelClassification - When you add it to the schema config, make it conditional on the classificationMethod value. See, for example, the Textract 'features' property which renders only when ocr backend value is textract - [code here](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/main/patterns/pattern-2/template.yaml#L225-L243)

***However, as Mark has noted -*** 
> @rstrahan sorry for the delay, was on vacation and busy with 7 projects/customers. I also just noticed that you released a new experimental feature multimodalPageBoundaryClassification, does that make groupConsecutivePagesOfSameClass redundant?

We need to discuss and close on how to best integrate the new `multimodalPageBoundaryClassification` this week, but can you review this and see if Mark's PR is now redundant?

Tx.

rstrahan:
@markjayne-amzn  - please create a new - separate - issue for the OOM and the InvalidParameterException errors..  and separate it out from the consecutive pages topic - it's a completely separate issue that we need to troubleshoot separately.  
In that issue, please include the troubleshooting you shared above, and also the 'Steps to reproduce' the error. 
I'm reluctant to simply double the memory of all lambdas without first understanding the root cause of the problem.. but if you create an issue and show me how to reproduce it, I will troubleshoot and fix. Tx.

markjayne-amzn:
@rstrahan here's a new issue, thanks for looking into this :) 
https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/35

PR #48, opened 2025-09-04
feat(agentic idp): First version of agentic IDP using strands

kazmer97:
## Summary

• Add agentic extraction module using Strands agents for structured data output without JSON parsing errors • Enhance
extraction service with dynamic Pydantic model generation from configuration attributes • Maintain full backward
compatibility - traditional extraction remains default behavior

## Changes Made

• New: lib/idp_common_pkg/idp_common/extraction/agentic_idp.py - Core agentic functionality with tool-based
extraction
• Modified: lib/idp_common_pkg/idp_common/extraction/service.py - Added dynamic model generation and agentic
integration
• Updated: lib/idp_common_pkg/idp_common/extraction/README.md - Comprehensive documentation with usage examples

## Benefits

• Eliminates JSON parsing errors through validated Pydantic models
• Improves extraction accuracy with self-correcting agent tools
• Provides dual extraction paths (traditional vs agentic) based on configuration
• Zero breaking changes - existing workflows continue unchanged

This enhancement significantly improves extraction reliability while preserving all existing functionality and
configuration patterns.

Screenshot 2025-09-18 at 12 45 55
Screenshot 2025-09-18 at 12 45 44

Screenshot 2025-09-18 at 12 46 20


lib/idp_common_pkg/idp_common/extraction/agentic_idp.py
```diff
@@ -0,0 +1,753 @@
...
+    def extraction_tool(
+        extraction: model_class,  # pyright: ignore[reportInvalidTypeForm]
+        agent: Agent,  # pyright: ignore[reportInvalidTypeForm]
+    ) -> dict:  # pyright: ignore[reportInvalidTypeForm]
+        logger.debug(f"extraction_tool called with data: {extraction}")
+        extraction_model = model_class(**extraction)  # pyright: ignore[reportAssignmentType]
+        extraction_dict = extraction_model.model_dump()
+        agent.state.set(key="current_extraction", value=extraction_dict)
+        logger.debug(f"Successfully stored extraction in state: {extraction_dict}")
+        return extraction_dict
```
kazmer97: this could be replaced with just a success message

template.yaml
```diff
@@ -138,6 +138,11 @@ Parameters:
     Default: ""
     Description: Specify Custom Fine Tuned Extraction Model ARN. Leave blank to use pretrained model specified in the selected configuration.
 
+  Pattern2ECRRepositoryUri:
```
rstrahan: Does this need to be a parameter of the main stack?  I feel that there is not a use case for user to ever change it, and if it's a parameter, but we change the default value when we publish, stack updates will keep the old value.
I think the <PATTERN2_ECR_REPO_URI> token should be embedded in the Resources where needed to pass it as a parameter to nested stack.

patterns/pattern-2/template.yaml
```diff
@@ -559,6 +570,27 @@ Resources:
               - system_prompt
               - task_prompt
             properties:
+              agentic:
```
rstrahan, line 827: How do prompts work..  Does the agent still use the existing System prompt and Task prompt?  (I had imagined the agent would need different prompt variants, but if not, that's great!)

publish.py
```diff
@@ -1646,6 +1697,312 @@ def _validate_python_syntax(self, directory):
                         return False
         return True
 
+    def _fix_pattern2_image_uris(self, template_path):
```
rstrahan: I thought SAM could figure all this out automatically.. no?
See https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-imageuri:~:text=Lambda%3A%3AFunction%20resource.-,ImageUri,-The%20URI%20of

Seems to imply ImageUri can be omitted if you do this:

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-build.html#build-container-image

Can you check?  Would be so much simpler to maintain.

notebooks/examples/step3_extraction_using_agent_idp.ipynb
rstrahan, line 1: Love this notebook, but it failed for me.. screen shot attached..

image

template.yaml
kazmer97: **Changes made:**
1. Kept the parameter for passing to nested stack (required for CloudFormation)
2. Token replacement happens during publish (line 1348 in publish.py)
3. Removed the problematic `_fix_pattern2_image_uris()` function that was manipulating ImageUri values

The token `<PATTERN2_ECR_REPO_URI>` is now only replaced in the main template during publish, and SAM handles the actual ImageUri values automatically via `sam package --image-repository`.

This maintains the parameter for nested stack usage while avoiding the stack update issue you identified.

patterns/pattern-2/template.yaml
kazmer97: How Prompts Work with Agentic Extraction
The agentic extraction uses the existing System and Task prompts, but in a different way than traditional extraction:

Traditional Extraction
System prompt: Sent directly to Bedrock as system message
Task prompt: Sent as user message
Model responds with JSON text that needs parsing
Agentic Extraction
System prompt → Passed via custom_instruction parameter and appended to agentic system prompt
Task prompt → Sent as user message (content blocks with text/images)
Uses Strands agent with tools for structured output
No JSON parsing needed - returns validated Pydantic model
Key Difference
The agentic system prompt (in agentic_idp.py) provides extraction guidelines and tool usage instructions. Your existing system/task prompts are incorporated as custom instructions to guide the extraction without changing the core agent behavior.

lib/idp_common_pkg/tests/conftest.py
```diff
@@ -11,16 +11,22 @@
...
 # These mocks need to be set up before any imports that might use these packages
 
-# Mock strands modules for agent functionality
-sys.modules["strands"] = MagicMock()
-sys.modules["strands.models"] = MagicMock()
-sys.modules["strands.hooks"] = MagicMock()
-sys.modules["strands.hooks.events"] = MagicMock()
+# Mock strands modules for agent functionality only if not already installed
+try:
+    import strands
```
rstrahan: this fails to get past the 'ruff lint' in the publish script..  or in 'make lint'

~/projects/genaiic-idp-accelerator (review/pr-48-agentic-idp)$ make lint
ruff check --fix
F401 `strands` imported but unused; consider using `importlib.util.find_spec` to test for availability
  --> lib/idp_common_pkg/tests/conftest.py:16:12
   |
14 | # Mock strands modules for agent functionality only if not already installed
15 | try:
16 |     import strands
   |            ^^^^^^^
17 | except ImportError:
18 |     sys.modules["strands"] = MagicMock()
   |
help: Remove unused import: `strands`

F401 `bedrock_agentcore` imported but unused; consider using `importlib.util.find_spec` to test for availability
  --> lib/idp_common_pkg/tests/conftest.py:25:12
   |
23 | # Mock bedrock_agentcore modules for secure code execution only if not already installed
24 | try:
25 |     import bedrock_agentcore
   |            ^^^^^^^^^^^^^^^^^
26 | except ImportError:
27 |     sys.modules["bedrock_agentcore"] = MagicMock()
   |
help: Remove unused import: `bedrock_agentcore`

patterns/pattern-2/template.yaml
```diff
@@ -563,6 +574,27 @@ Resources:
...
+                    type: boolean
+                    description: Enable Agentic Extraction
+                    order: 0
+                    default: false
+                  review_agent:
+                    type: boolean
+                    description: This introduces a second agent to review the first agents work. Only use with highly complex workflows as it increases token usage.
+                    order: 1
+                    default: false
+                  enable_caching:
```
rstrahan: We should be able to automate this..  already in idp_common bedrock client we have a list of models that support caching, so use that to auto detect and cache if model supports it.

config_library/pattern-2/lending-package-sample/config.yaml
```diff
@@ -976,6 +976,9 @@ classification:
     4. Outputting in the exact JSON format specified in <output-format>
     </final-instructions>
 extraction:
+  agentic:
+    enabled: true
```
rstrahan: I think we should not enable agentic extraction by default..  that raises the stakes too much, I think, till we get a bit more experience with it..  Maybe later when it's lower risk.

docs/extraction.md
```diff
@@ -9,6 +9,64 @@ Information extraction is a central capability of the GenAIIDP solution, transfo
 
 Configure extraction behavior through several components:
 
+### Agentic Extraction (Recommended for Production)
```
rstrahan: Is it too soon to say this..  Better to say (Preview) or (Experimental) for now, IMHO, till we have more testing and ironed out bugs..

lib/idp_common_pkg/tests/conftest.py
kazmer97: fixed

docs/extraction.md
```diff
@@ -9,6 +9,64 @@ Information extraction is a central capability of the GenAIIDP solution, transfo
...
+
+The extraction service supports two modes: **traditional** and **agentic**. Agentic extraction provides superior accuracy and consistency, especially for complex documents with nested structures or strict schema requirements.
+
+#### When to Enable Agentic Extraction
+
+Enable agentic extraction when you need:
```
rstrahan, line 20: Can we say something about this enabling future extensions involving MCP servers to integrate validation or enrichment into the extraction process... (eg validate names, addresses, etc from external services)

docs/extraction.md
```diff
@@ -32,44 +90,73 @@ classes:
...
   # Prompts for extraction
   system_prompt: |
     You are an expert in extracting structured information from documents.
     Focus on accuracy in identifying key fields based on their descriptions.
     For each field, look for both the field label and the associated value.
     Pay attention to formatting patterns common in business documents.
     When a field is not present, indicate this explicitly rather than guessing.
-    
+
   task_prompt: |
```
rstrahan, line 137: Can you explain how the system_prompt and task_prompts apply to the agentic extraction..  Does the agent use the same prompt as 'traditional' method?

rstrahan (changes requested):

lib/idp_common_pkg/idp_common/extraction/agentic_idp.py
rstrahan, line 1: See slack comment on throttle catching..

it failed with
Execution Failed
Error: ValueError: Agent invocation failed: An error occurred (ThrottlingException) when calling the ConverseStream operation (reached max retries: 4): Too many tokens, please wait before trying again.
We need to keep retrying... for much much longer..  [@kaznb](https://amazon.enterprise.slack.com/team/U05T7TR53DM) can you check
idp_common bedrock client for exponential backoff/retry logic
state machine for retries at lambda invocation level..
We should never fail due to throttling.. at least not for several hours..
Note - I seem to get this error every time with Sonnet 4.5 in my isengard account when processing just 1 lending_package.pdf (Isengard has really low quota), but, whatever the customer quota is, we still need to defend against throttling when volumes are large and demand exceeds quota

config_library/pattern-2/lending-package-sample/config.yaml
kazmer97: moved to false by default

template.yaml
rstrahan: But you can pass the value as a parameter to nested stacks without exposing it as a parameter in the main template..  
Just use the TOKEN as the parameter value.. ie instead of
```
937        ECRRepositoryUri: !Ref Pattern2ECRRepositoryUri
938        ImageVersion: !Ref Pattern2ImageVersion
```
do
```
937        ECRRepositoryUri: "<PATTERN2_ECR_REPO_URI>"
938        ImageVersion: "<IMAGE_VERSION>"
```
The publish script will still replace them, and we avoid confusing users with unnecessary visible parameters.

template.yaml
```diff
@@ -923,6 +934,8 @@ Resources:
           - !GetAtt WorkforceURLResource.PortalURL
           - ""
         PermissionsBoundaryArn: !Ref PermissionsBoundaryArn
+        ECRRepositoryUri: !Ref Pattern2ECRRepositoryUri
```
rstrahan: See above comment.. 
Do 
```
937        ECRRepositoryUri: "<PATTERN2_ECR_REPO_URI>"
938        ImageVersion: "<IMAGE_VERSION>"
```
instead, and remove the params..

kazmer97: when we do this SAM defaults the tag of the image to latest, which means on subsequent local builds and pushes cdk doesn't update the lambdas functions as the image tag doesn't change. The alternative to allow SAM to build the images and specify the tag that way slows it down significantly as SAM doesn't parallelise builds quite as well for docker

publish.py
```diff
@@ -226,7 +232,7 @@ def print_usage(self):
         """Print usage information with Rich formatting"""
         self.console.print("\n[bold cyan]Usage:[/bold cyan]")
         self.console.print(
-            "  python3 publish.py <cfn_bucket_basename> <cfn_prefix> <region> [public] [--max-workers N] [--verbose] [--no-validate] [--clean-build] [--lint on|off]"
```
rstrahan, line 229: Why remove --clean-build ?

template.yaml
```diff
@@ -138,6 +136,13 @@ Parameters:
     Default: ""
     Description: Specify Custom Fine Tuned Extraction Model ARN. Leave blank to use pretrained model specified in the selected configuration.
 
+  Pattern2ImageVersion:
```
rstrahan: do not add main template parameter for Pattern2ImageVersion..  this is never something a user would change..   (Same comment I made before when you had ECR URI as a paramerter of the main stack..)

OK to make it a parameter of the nested PATTERN2STACK  (since user never sees that), but not a parameter of the main stack.

template.yaml
```diff
@@ -913,16 +942,29 @@ Resources:
...
         SageMakerA2IReviewPortalURL: !If
           - IsHITLEnabled
           - !GetAtt WorkforceURLResource.PortalURL
           - ""
         PermissionsBoundaryArn: !Ref PermissionsBoundaryArn
+        ECRRepositoryUri: !GetAtt Pattern2ECRRepository.RepositoryUri
+        # ImageVersion uses content-based hash from publish process unless overridden
+        # This ensures images are only rebuilt when code changes, while allowing manual overrides
+        ImageVersion:
+          !If
```
rstrahan: I'm confused by this conditions.. Surely there will *always* be an image version, since we create it in the publish script.

template.yaml
```diff
@@ -6707,6 +6983,127 @@ Resources:
       # previous version if build fails.
       CodeLocation: !Sub "arn:${AWS::Partition}:s3:::<ARTIFACT_BUCKET_TOKEN>/<ARTIFACT_PREFIX_TOKEN>/<WEBUI_ZIPFILE_TOKEN>"
 
+  # ==================== Pattern-2 Docker Build Resources ====================
+
+  Pattern2ECRRepository:
```
rstrahan: Lose the 'Pattern2' from the resource name..    We'll eventually use the same EcrRepository for all patterns

template.yaml
```diff
@@ -6707,6 +6983,127 @@ Resources:
...
+    Type: AWS::ECR::Repository
+    Condition: IsPattern2
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
+    Properties:
+      # ECR names must be lowercase - use unique stack ID portion for uniqueness per stack
+      RepositoryName: !Sub
```
rstrahan: I don't think RepositoryName is required, so leave it out, and let CloudFormation give it a name.. this is much more robust and easier to defend against name conflicts with existing resources.

template.yaml
```diff
@@ -6707,6 +6983,127 @@ Resources:
...
+        - "idp-pattern2-${StackGuid}"
+        - StackGuid: !Select [2, !Split ['/', !Ref 'AWS::StackId']]
+      ImageScanningConfiguration:
+        ScanOnPush: true
+      EncryptionConfiguration:
+        EncryptionType: AES256
```
rstrahan: use the CustomerManagedEncryptionKey for the encryption

PR #52, opened 2025-09-08
Feature/s3 vectors backend

HatmanStack:
# Add Dual Backend Knowledge Base Support: OpenSearch + S3 Vectors

## 🎯 Overview

This PR introduces **dual backend support** for knowledge bases, expanding from the single OpenSearch option to support both **OpenSearch** (real-time, higher performance) and **S3 Vectors** (serverless, cost-effective) backends. Users can now choose their knowledge base infrastructure based on performance requirements and cost considerations.

## ✨ Key Features Added

- **🔄 Dual Backend Architecture**: Choose between OpenSearch and S3 Vectors
- **💰 Cost Optimization**: S3 Vectors provides pay-per-query pricing vs. always-on OpenSearch clusters  
- **🚀 Serverless Scaling**: S3 Vectors automatically scales without infrastructure management
- **🧠 Intelligent Filtering**: LLM-powered metadata filtering for improved search relevance
- **📊 Dynamic UI**: Configuration interface adapts based on selected backend
- **🔗 Unified API**: Same GraphQL interface regardless of backend choice

### Backend Selection Flow
```
User Selects Backend → CloudFormation Deployment → Lambda Routing → Query Processing
     ↓                        ↓                         ↓               ↓
OpenSearch/S3Vectors    Conditional Resources    KB_TYPE Environment   Backend-Specific Logic
```

#### **ConfigurationLayout.jsx**
- ✅ Added `KnowledgeBaseStatus` component displaying active backend type and performance characteristics
- ✅ Implemented feature gating - S3 Vectors configuration only appears when backend is enabled
- ✅ Dynamic UI adaptation based on `schemaToUse` computed property

#### **KnowledgeBaseStatus.jsx** *(NEW)*
- 🆕 Real-time backend status indicator (OpenSearch/S3 Vectors/None)
- 🆕 Performance descriptions and cost implications
- 🆕 Integrated with settings context for automatic updates

#### **template.yaml - Major Overhaul**

**Parameter Updates:**
```yaml
# Before
DocumentKnowledgeBase: ["DISABLED", "BEDROCK_KNOWLEDGE_BASE (Create)"]

# After  
DocumentKnowledgeBase: ["OpenSearch", "S3 Vectors", "Disabled"]
```

**New S3 Vectors Parameters:**
- `VectorSimilarityMeasure`: Cosine/Euclidean distance metrics
- `KnowledgeBaseEmbeddingModelId`: Embedding model selection
- `KnowledgeBaseVectorIndexName`: Custom index naming
- `S3VectorsIngestionScheduleExpression`: Automated ingestion scheduling

**New Resources:**
- `S3VectorsCatalogTable`: DynamoDB metadata catalog
- `DOCUMENTS3VECTORSKB`: Complete S3 Vectors CloudFormation stack
- `EmbeddingModelDimensions`: Mapping for vector dimensions by model

**New UI Build Flow:**
- switched to yarn and installing from frozen file

**Enhanced Conditions:**
- `ShouldCreateS3VectorsKnowledgeBase`: S3 Vectors creation logic
- `ShouldUseAnyKnowledgeBase`: Unified backend detection

#### **1. S3 Vectors Provisioner** (`src/s3_vectors_provisioner/`)
```python
# Key capabilities
- provision_s3_vector_resources()  # CloudFormation custom resource
- get_s3vectors_client()          # Service client initialization  
- send_cfn_response()             # CF lifecycle management
```

#### **2. S3 Vectors Ingestion** (`src/s3_vectors_ingestion/`)
```python
# Document processing pipeline
- DocumentReader class           # Text extraction & chunking
- process_document_folder()      # Individual document processing
- store_vectors_in_batches()     # Batch vector uploads
- classify_chunk_metadata()      # Filterable metadata generation
- update_dynamodb_catalog()      # Catalog maintenance
```

#### **3. S3 Vectors Query** (`src/s3_vectors_query/`)
```python
# Advanced query processing
- process_query_pipeline()       # Main query workflow
- query_s3_vectors()            # Vector similarity search
- enhance_query_with_filter()   # Intelligent metadata filtering
- rerank_candidates()           # LLM-based result re-ranking
- generate_response()           # Citation-rich responses
```

#### **4. S3 Metadata Filter Analyzer** (`src/s3_metadata_filter_analyzer/`)
```python
# Intelligent filtering system  
- analyze_metadata()            # Vector metadata sampling
- generate_filter_examples()    # LLM filter template creation
- store_filter_examples()       # S3 template storage
- update_lambda_environment()   # Dynamic configuration
```

#### **Knowledge Base Query Resolver** (`src/lambda/query_knowledgebase_resolver/`)
- ✅ **Backend Detection**: Routes queries based on `KB_TYPE` environment variable
- ✅ **Unified Interface**: Same GraphQL response format for both backends
- ✅ **Backward Compatibility**: Existing OpenSearch functionality unchanged
- 🆕 `get_s3_vectors_response()`: Direct S3 Vectors Lambda invocation
- 🔧 Enhanced `extract_document_id()`: Supports both backend metadata formats

#### **Pattern 2 & Pattern 3 Templates**
```yaml
# New S3 Vectors configuration section
s3Vectors:
  filterableMetadataKeys: []  # Up to 6 custom metadata fields
  # Built-in keys: document_id, document_type, section_type, confidence
```

#### **S3 Vectors Helper Package** (`lib/idp_common_pkg/idp_common/s3vectors/`)
```python
# Reusable S3 Vectors client
class S3VectorsClient:
    - create_bucket()    # Storage provisioning
    - create_index()     # Vector index setup  
    - put_vectors()      # Batch vector uploads
    - query_vectors()    # Similarity searches
```

#### **Build Optimization** (`publish.py`)
- ✅ Excluded `build/` and `node_modules/` from deployment artifacts
- ✅ Added `uv` support for enhanced Python package management
- ✅ Reduced deployment size and upload time

#### **Dependency Management** (`lib/idp_common_pkg/pyproject.toml`)
```toml
[project.optional-dependencies]
all = [
    "PyYAML==6.0.1",     # Configuration file support
    "rich==13.7.1",      # Enhanced terminal output  
    "setuptools==70.1.1" # Build requirements
]
```

## 🔄 Migration & Compatibility

- ✅ **Zero Breaking Changes**: Existing OpenSearch deployments continue working
- ✅ **Feature Parity**: Both backends support the same query interface
- ✅ **Configuration Preservation**: Existing settings remain compatible

## 🧪 Testing Considerations

- **Unit Tests**: Verify backend routing logic in query resolver
- **Integration Tests**: Test both OpenSearch and S3 Vectors query flows
- **UI Tests**: Validate dynamic configuration display
- **Performance Tests**: Compare query latency between backends
- **Cost Analysis**: Validate S3 Vectors pricing model

## 📖 Future Feature Enhancements 

- [ ] Add Ability for Multi-Index set ups
- [ ] Create Metrics Integration for Tracking 
- [ ] Add Query and Seperate Index for Sections Portion of Output Bucket To Speed Up Queries on Large Datasets
- [ ] Options for ReEmbedding and Model Change
- [ ] Add performance tuning 

## 🚀 Deployment Impact

**New CloudFormation Parameters Required:**
- Existing deployments: No action needed (defaults to OpenSearch)
- New deployments: Choose between OpenSearch/S3 Vectors
- S3 Vectors users: Configure embedding model, similarity measure, and ingestion schedule

**Resource Creation:**
- S3 Vectors: Creates DynamoDB catalog table + Lambda functions + S3 Vectors Bucket & Index
- OpenSearch: Unchanged resource creation pattern

---

This implementation provides users with flexible, cost-effective knowledge base options while maintaining full backward compatibility and a consistent user experience.


options/s3-vectors-kb/src/s3_vectors_ingestion/index.py
tanimath, line 1: @HatmanStack thanks for the PR. This is much better as yesterday version was failing for pattern 1, however I still see issues related to confidence assessment for pattern 1.

# S3 Vectors Ingestion Analysis: Pattern 1 BDA Compatibility

## Executive Summary
The S3 vectors ingestion code **runs without errors** on Pattern 1 BDA output but **processes 0 vectors** due to missing confidence files. All pages are effectively skipped.

## Current Behavior Analysis

### ✅ No Lambda Errors
- Code has proper exception handling for missing files
- Continues processing when `textConfidence.json` is missing
- Logs errors but doesn't crash

### ❌ Zero Vector Processing Confirmed
- All pages are skipped due to missing `textConfidence.json`
- `_get_preferred_text()` returns empty list `[]`
- `extract_text_chunks()` receives empty list and returns `[]`
- Result: 0 vectors created and stored

## Code Flow Analysis

```python
def _get_preferred_text(self):
    page_texts = []  # Starts empty
    for page_num in pages:
        try:
            # Get result.json ✅ Works
            result_data = json.loads(...)
            
            # Get textConfidence.json ❌ Fails for BDA
            confidence_data = json.loads(...)  # Exception thrown here
            
        except Exception as e:
            print(f"Error reading page {page_num}: {e}")
            continue  # Skip this page, page_texts remains empty
    
    return page_texts  # Returns empty list []

def extract_text_chunks(self):
    pages = self._get_preferred_text()  # Gets []
    if not pages:  # True for empty list
        return []  # Returns empty, no processing
```

## BDA Output Structure

### What Exists in Pattern 1 BDA
```
lending_package.pdf/
├── pages/0-5/
│   ├── image.jpg                  ✅ Present
│   ├── parsedResult.json          ✅ Present {"text": "markdown"}
│   ├── result.json                ✅ Present (BDA page data)
│   └── textConfidence.json        ❌ Missing (OCR confidence)
├── sections/0-5/
│   └── result.json                ✅ Present (explainability_info)
```

### What Code Expects vs Reality
| File | Code Expects | BDA Provides | Result |
|------|-------------|--------------|---------|
| `textConfidence.json` | OCR confidence | Missing | Exception → Skip page |
| `result.json` | Page data | BDA page data | Available but unused |
| `parsedResult.json` | Not used | Text content | Available but unused |

## BDA Confidence Data Location

**Available in:** `sections/{id}/result.json`
```json
{
  "explainability_info": [
    {
      "FIELD_NAME": {
        "confidence": 0.92578125,
        "confidence_threshold": 0.8,
        "geometry": [{"page": 2, "boundingBox": {...}}]
      }
    }
  ]
}
```

## Solution Required

### Modify `_get_preferred_text()` Method
1. **Remove dependency** on `textConfidence.json`
2. **Extract confidence** from BDA sections' `explainability_info`
3. **Map section confidence** to pages using geometry data
4. **Use text** from existing `parsedResult.json` or `result.json`

### Implementation Approach
```python
# Instead of reading textConfidence.json:
try:
    result_data = json.loads(...)  # ✅ Works
    
    # Get confidence from BDA sections (new logic needed)
    confidence_score = self._get_bda_confidence_for_page(page_id)
    
    if result_data and confidence_score:
        page_texts.append(...)
```

## Impact Assessment

### Current State
- ✅ No crashes or errors
- ❌ 0 vectors processed
- ❌ No searchable content
- ❌ Knowledge base empty

### After Fix
- ✅ Vectors created with BDA confidence data
- ✅ Content searchable
- ✅ Knowledge base functional
- ✅ Proper confidence metadata preserved

## Files Requiring Changes
- `options/s3-vectors-kb/src/s3_vectors_ingestion/index.py`
  - Method: `DocumentReader._get_preferred_text()` (lines ~235-245)
  - Add: BDA section confidence extraction logic
  - Add: Page-to-section confidence mapping

HatmanStack: Thanks for the feedback!  I had meant to gate pattern-1 as using BDA seemed out of scope for this MVP.  Happy to take a look and add it in.  Should be no problem.  What do you think about some type of error handling to not fail the ingestion?  Maybe a placeholder number?  I only worry about our banded query search slice in `fetch_candidates` of s3_vectors_query.  I think the other slices will still pick up the vector that didn't get the confidence ingested but we could run into problems when we go for optimization and tuning.  I'll make a note.  Would be interested in how you're thinking about it.

tanimath: Thanks @HatmanStack  for talking a look into this. I checked the latest code and it works well for pattern 1 where textConfidence.json file is missing, it now uses a default confidence score of 50 which seems fair as fall back approach. I don't think there is any impact for vector search as long as we are using single pattern deployment, for mixed pattern there could be unfair disadvantage to pattern 1 results as pattern 2/3 will use actual confidence score whereas pattern1 will get the default 50. May be we should change default to 80 so that it get included in the compose filter? I am not sure on this. For now I think 50 confidence score will work correctly for single pattern deployment.

tanimath:
@rstrahan I checked bedrockkb implementation and its independent of any pattern specific implementation as it doesn't include confidence score in the vector meta in bedrockkb. The key difference with s3-vector-kb is that it saves textract confidence score from textconfidence.json file in s3 vector em-bedding's and it factors in while evaluating kb query response. 

I have captured below for comparison between bedrockkb and s3-vector-kb

## Confidence Score Logic Differences:

### **bedrockkb (Bedrock Knowledge Base):**
• **No stored confidence scores** - vectors don't have confidence metadata
• **Dynamic similarity scores** calculated at query time using vector similarity (cosine/euclidean)
• **Single score type**: relevance score (0.0-1.0) based purely on semantic similarity
• **No filtering by confidence** - all vectors are treated equally

### **s3-vectors-kb (S3 Vectors Knowledge Base):**

1. Stored Confidence Scores:
• **OCR confidence scores** extracted from textConfidence.json files
• Calculates average_confidence, min_confidence, and count per document chunk
• Stores confidence as metadata with each vector (renamed from average_confidence to confidence)
• **Default confidence of 50** when textConfidence.json is unavailable

2. Multi-layered Scoring System:
• **Similarity score**: Distance converted to similarity using score = 1.0 / (1.0 + distance)
• **Rerank score**: LLM-based relevance scoring (0.0-1.0)
• **Final score**: Weighted combination: (0.5 * similarity_score) + (0.5 * rerank_score)

3. Confidence-based Filtering:
• Uses confidence scores for query optimization
• **Multi-slice retrieval** with confidence filters:
  • Unfiltered slice (all documents)
  • Base filter slice
  • **High confidence slice** (confidence >= 0.8)
• Improves recall by retrieving from multiple confidence bands

Key Advantage of s3-vectors-kb:
The confidence scores enable quality-aware retrieval - you can prioritize high-confidence OCR results while 
still falling back to lower-confidence content when needed, leading to more reliable document processing 
results.

agents.json
rstrahan, line 1: This looks like a AI helper file - should not be part of the PR..  I will remove.

docs/s3-vectors.md
```diff
@@ -0,0 +1,309 @@
...
+
+| Parameter | Values | Description |
+|-----------|--------|-------------|
+| `DocumentKnowledgeBase` | `OpenSearch` \| `S3 Vectors` \| `Disabled` | Selects knowledge base backend or disables it entirely |
+| `KnowledgeBaseModelId` | Model ARN | Foundational model for knowledge base chat (when enabled) |
+| `DocumentAnalysisAgentModelId` | Model ARN | Model for document analytics queries (only for OpenSearch) |
+
+#### S3 Vectors Specific Parameters
+
+These parameters only appear when S3 Vectors is selected:
```
rstrahan, line 48: How is this possible?  Cfn does not have 'conditional' parameters to my knowledge.

docs/s3-vectors.md
```diff
@@ -0,0 +1,309 @@
...
+
+When deploying the solution, you configure the knowledge base backend through these parameters:
+
+#### Primary Configuration
+
+| Parameter | Values | Description |
+|-----------|--------|-------------|
+| `DocumentKnowledgeBase` | `OpenSearch` \| `S3 Vectors` \| `Disabled` | Selects knowledge base backend or disables it entirely |
+| `KnowledgeBaseModelId` | Model ARN | Foundational model for knowledge base chat (when enabled) |
+| `DocumentAnalysisAgentModelId` | Model ARN | Model for document analytics queries (only for OpenSearch) |
```
rstrahan, line 44: Why is this only for OpenSearch?  We need to use a model for RAG inferences, whether we're using OpenSearch or S3 as the vectorstore, right?

docs/s3-vectors.md
```diff
@@ -0,0 +1,309 @@
...
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `VectorSimilarityMeasure` | `cosine` | Distance metric (`cosine` \| `euclidean`) |
+| `KnowledgeBaseEmbeddingModelId` | `amazon.titan-embed-text-v2:0` | Embedding model for vectorization |
+| `KnowledgeBaseVectorIndexName` | `idp-vectors-index` | Name for the vector index |
+| `S3VectorsIngestionScheduleExpression` | `rate(30 minutes)` | Automated ingestion frequency |
+
+## S3 Vectors Configuration
+
+### Embedding Models and Dimensions
```
rstrahan, line 59: This seems like too much to expose.. most users won't know/care - they just want it to work. Can we go with workable defaults?

docs/s3-vectors.md
```diff
@@ -0,0 +1,309 @@
...
+### Vector Similarity Measures
+
+Choose the appropriate similarity measure for your use case:
+
+- **Cosine Similarity** (Recommended): Measures angle between vectors, normalized for document length
+- **Euclidean Distance**: Measures absolute distance, sensitive to vector magnitude
+
+### Filterable Metadata Configuration
+
+S3 Vectors supports custom metadata filtering for improved search precision. You can specify up to 6 additional filterable metadata keys beyond the built-in fields:
```
rstrahan, line 79: I like the idea of filtering.. But don't we also need UI filter controls to enable this to be used?

lib/idp_common_pkg/pyproject.toml
```diff
@@ -129,6 +129,9 @@ all = [
     "openpyxl==3.1.5",
     "python-docx==1.2.0",
     "strands-agents>=1.0.0",
+    "PyYAML==6.0.1",
```
rstrahan, line 132: I don't believe we need these packages in the idp_common_pkg - they are needed only by the publish script, not for runtime document processing.

lib/idp_common_pkg/setup.py
```diff
@@ -62,6 +62,10 @@
     "docs_service": [
         "requests==2.32.4",
     ],
+    # S3 Vectors module dependencies
```
rstrahan, line 65: Since it's empty, we don't need it.

lib/idp_common_pkg/idp_common/s3vectors/client.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def create_bucket(self, vector_bucket_name: str) -> None:
+        """
+        Idempotently creates an S3 Vector bucket.
+
+        Args:
+            vector_bucket_name: The name for the vector bucket.
+        """
+        try:
+            self.s3vectors.create_vector_bucket(vectorBucketName=vector_bucket_name)
```
rstrahan, line 45: Will require KMS Encryption to be complaint with our security requirements - encryptionConfiguration

options/s3-vectors-kb/src/s3_vectors_provisioner/index.py
```diff
@@ -0,0 +1,119 @@
...
+        dimension: The dimensionality of the vectors that will be stored.
+        distance_metric: The metric used to measure vector similarity (e.g., 'cosine').
+
+    Returns:
+        A dictionary containing the names of the created bucket and index.
+    """
+    logger.info(f"Provisioning resources: bucket='{vector_bucket_name}', index='{vector_index_name}'")
+    client = get_s3vectors_client()
+     
+    client.create_bucket(vector_bucket_name)
```
rstrahan, line 47: Encryption needed.

options/s3-vectors-kb/src/s3_vectors_provisioner/index.py
```diff
@@ -0,0 +1,119 @@
...
+        vector_bucket_name,
+        vector_index_name,
+        dimension,
+        distance_metric
+        )
+    
+    
+    return {'bucket': vector_bucket_name, 'index': vector_index_name}
+
+def send_cfn_response(event: Dict[str, Any], context: Any, status: str, data: Dict, physical_resource_id: str = None):
```
rstrahan, line 61: use standard cfnresponse package instead

patterns/pattern-1/template.yaml
```diff
@@ -362,7 +362,7 @@ Resources:
                   model:
                     type: string
                     description: Bedrock model ID
-                    enum: ["us.amazon.nova-lite-v1:0", "us.amazon.nova-pro-v1:0", "us.amazon.nova-premier-v1:0", "us.anthropic.claude-3-haiku-20240307-v1:0", "us.anthropic.claude-3-5-sonnet-20241022-v2:0", "us.anthropic.claude-3-7-sonnet-20250219-v1:0", "us.anthropic.claude-sonnet-4-20250514-v1:0", "us.anthropic.claude-sonnet-4-20250514-v1:0:1m", "us.anthropic.claude-opus-4-20250514-v1:0", "us.anthropic.claude-opus-4-1-20250805-v1:0"]
+                    enum: ["amazon.nova-lite-v1:0", "amazon.nova-pro-v1:0", "amazon.nova-premier-v1:0", "anthropic.claude-3-haiku-20240307-v1:0", "anthropic.claude-3-5-sonnet-20241022-v2:0", "anthropic.claude-3-5-haiku-20241022-v1:0"]
```
rstrahan, line 365: Looks like a regression in the list of models.. we recently added sonnet and opus 4x.. this removes them.

src/ui/src/components/document-kb-query-layout/DocumentsQueryLayout.jsx
rstrahan, line 1: To make use of metadata filters, we'd need to enhance UI to expose the filters, right?

rstrahan: Or are you using filters exclusively through conversion from natural language?

agents.json
HatmanStack: added to .gitignore

docs/s3-vectors.md
HatmanStack: completed

docs/s3-vectors.md
HatmanStack: completed

docs/s3-vectors.md
HatmanStack: If we're going with full BKB offload I completely agree.  It makes since only if we want to include things that core AWS doesn't offer.

docs/s3-vectors.md
HatmanStack: Included in the patterns' template.yaml 

``` 
s3Vectors:
            order: 9
            type: object
            sectionLabel: S3 Vectors
            properties:
              filterableMetadataKeys:
                type: array
                description: "Add up to 6 metadata keys for filtering.  These 2 are always included: document_id, confidence."
                listLabel: "Filterable metadata keys"
                itemLabel: "Metadata key"
                maxItems: 6
                items:
                  type: string
```

and gated in the ConfigurationLayout.jsx

```
  const schemaToUse = React.useMemo(() => {
    if (!schema) return null;
    if (isS3VectorsEnabled) return schema;
    if (!schema.properties || !schema.properties.s3Vectors) return schema;
    const gated = { ...schema, properties: { ...schema.properties } };
    delete gated.properties.s3Vectors;
    return gated;
  }, [schema, isS3VectorsEnabled]);
```

lib/idp_common_pkg/idp_common/s3vectors/client.py
HatmanStack: completed

options/s3-vectors-kb/src/s3_vectors_provisioner/index.py
HatmanStack: Not sure here, the docs from boto3: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3vectors/client/query_vectors.html  


disregard this.  Was thinking query not bucket create.

options/s3-vectors-kb/src/s3_vectors_provisioner/index.py
HatmanStack: completed

patterns/pattern-1/template.yaml
HatmanStack: completed

src/ui/src/components/document-kb-query-layout/DocumentsQueryLayout.jsx
HatmanStack: They're auto selected using a Nova micro call once they are inputed in our webui configuration.  We build them at ingestion time based on the current filterable keys and the text chunk and then build the metadata-filter at query time from the examples (metadata filter analyzer), current filterable keys and the query.  We're only using them in one slice of our three pronged search. Not sure yet how this will change with BKB.

rstrahan:
Thanks..  Yes, please let's keep this implementation to be just using S3 Vectors as a data source for BKB.  If this limits how we handle chunking or filtering or anything else just now, so be it.. I expect these limits will get resolved over time, and it's more strategic for us stick with BKB than to try to roll our own.  
We can add more features over time..  
If you can get me a new PR by tomorrow I can still try to test and get in included this week.  Many thanks!

PR #62, opened 2025-09-23
Feature/expanded text ingestion

HatmanStack:
# Text Extraction Enhancement

## What Changed

Added intelligent PDF routing to avoid unnecessary OCR costs for text-native PDFs.

### TextExtractionService
- Located in `idp_common/text_extraction`
- Uses PyMuPDF to inspect PDFs and detect if they contain extractable text
- Extracts text directly from digital PDFs without OCR

### Enhanced OcrService
- Now checks if PDFs are text-native before sending to OCR
- Routes text-native PDFs to direct extraction
- Sends scanned PDFs and images to OCR as before
- Returns same Document object format regardless of path

## Processing Flow

```
PDF Upload → PDF Inspection → Text-native? 
                                  ↓
                               Yes → Direct text extraction
                                No → OCR (existing flow)
```

## Benefits

- **Cost Reduction**: No Textract calls for digital PDFs
- **Faster Processing**: Direct text extraction is much faster than OCR
- **Same Output**: Document object structure unchanged

## Tasks

- [ ] Test processing with diverse data
- [ ]  Filter out non-relevant images from processing pipeline
- [ ]  Include processed image metrics in reporting dashboard

### DIFF
--- a/lib/idp_common_pkg/idp_common/ocr/service.py
+++ b/lib/idp_common_pkg/idp_common/ocr/service.py
@@ -486,10 +486,11 @@ class OcrService:
                         document.errors.append(f"{error_msg} (see logs for full trace)")
             elif (
                 file_type == "pdf"
+                and self.backend != "textract"
                 and self.text_extraction_service.is_pdf_text_native(file_content)
             ):
-                # New path for text-native PDFs
-                logger.info("Processing as text-native PDF.")
+                # New path for text-native PDFs (only when not using Textract backend)
+                logger.info("Processing as text-native PDF with hybrid extraction.")
                 pdf_document = fitz.open(stream=file_content, filetype="pdf")
                 document.num_pages = len(pdf_document)
 
@@ -610,9 +611,12 @@ class OcrService:
 
                 pdf_document.close()
             else:
-                # Existing path for scanned PDFs and image files
+                # Standard OCR path: Used for Textract backend (all PDFs) or scanned PDFs with other backends
                 if file_type == "pdf":
-                    logger.info("Processing as scanned PDF/image-based PDF.")
+                    if self.backend == "textract":
+                        logger.info("Processing PDF with Textract backend (skipping hybrid extraction).")
+                    else:
+                        logger.info("Processing as scanned PDF/image-based PDF.")
                 else:
                     logger.info(f"Processing as image file: {file_type}")
 


lib/idp_common_pkg/idp_common/ocr/README.md
```diff
@@ -9,6 +9,26 @@ This module provides OCR (Optical Character Recognition) capabilities for proces
...
+
+To improve efficiency and reduce costs, the OCR service now includes an intelligent routing mechanism that distinguishes between text-native PDFs and scanned (image-based) PDFs. This avoids sending documents that already contain machine-readable text to an expensive OCR service.
+
+This logic is handled by the `TextExtractionService`, which is used internally by the `OcrService`.
+
+### How It Works
+
+1.  **PDF Inspection**: When a PDF document is processed, it's first inspected by the `TextExtractionService`.
+2.  **Content-based Routing**:
+    -   **Text-Native PDFs**: If the service detects a significant amount of selectable text, it extracts the text directly using `PyMuPDF`. This bypasses the OCR backend (Textract or Bedrock) entirely. All required document artifacts (page images, parsed text) are still generated for compatibility with the rest of the IDP pipeline.
```
rstrahan, line 22: What happens if it's a mixture - i.e. a Text PDF with embedded images that are important to OCR and extract?

lib/idp_common_pkg/idp_common/ocr/service.py
```diff
@@ -410,52 +567,7 @@ def process_document(self, document: Document) -> Document:
                         ): i
                         for i in range(num_pages)
                     }
-
-                    # Start memory monitoring in background thread
-                    memory_monitor_shutdown = self._start_memory_monitoring()
```
rstrahan, line 415: Did you mean to remove the memory monitoring?  Seems orthogonal to your feature and possibly an AI overreach?

lib/idp_common_pkg/idp_common/text_extraction/service.py
rstrahan, line 1: Does this need to be a new class, or can we keep it neatly nested and abstracted inside the existing OCR class?

patterns/pattern-2/template.yaml
```diff
@@ -632,11 +632,6 @@ Resources:
             type: object
             sectionLabel: Assessment Inference
             properties:
-              enabled:
```
rstrahan, line 635: Another AI snafuu?  Seems unintentional?  
Can you carefully review all the changes as we go fwd to catch any changes that aren't directly related to the feature you are building.. Say no to AI overreach and GitHub lineage problems :)

HatmanStack: Son of a gun.  I had a feature branch init issue that I was dealing with late last night and from the looks got turned around here.  reverting.  As an aside, the two toggles feels bad and something I don't remember from previous iterations.  Any chance to simplify?

lib/idp_common_pkg/idp_common/ocr/README.md
HatmanStack: you are right on this.  will revisit.

lib/idp_common_pkg/idp_common/text_extraction/service.py
HatmanStack: was thinking about the expansion to other text formats.  Right now we have: ["txt", "csv", "xlsx", "docx"].  I'd like to expand to be something more comprehensive:  [ ".docx", ".txt", ".rtf", ".odt", ".doc", ".pages", ".html", ".htm", ".xml", ".md", ".markdown", ".json", ".yaml", ".yml", ".csv", ".tsv", ".rst", ".asciidoc", ".adoc", ".epub", ".mobi", ".tex", ".py", ".js", ".java", ".sql", ".ini", ".conf", ".config", ".log", ".env", "xlsx"] I haven't done the work yet to understand what each will take and thought a new module might make sense for the future of the feature?

patterns/pattern-2/template.yaml
rstrahan: > As an aside, the two toggles feels bad and something I don't remember from previous iterations. 
I don't follow - can you clarify this?

lib/idp_common_pkg/idp_common/ocr/service.py
HatmanStack: It's been moved to the _process_futures helper function. Is that not right?

rstrahan: AH, I hadn't spotted a new location for it..  Just saw the removal..  If it's only been refactored for the better, and not removed, then no problems!

patterns/pattern-2/template.yaml
HatmanStack: Meant those two toggles in Assessment Inference—feels a bit cluttered. Maybe I missed why pattern2 vs pattern3 for assessment_validator? Is there a reason to keep both? Could we default to validation and just have passthrough for edge cases? Not a huge deal, just prefer a simpler UI.

rstrahan: Ah, i see.. Yeah, I kinda agree..  One of the team members recently added better error checking, and part of that was to add validation logic on top of the assessment output, to check that the assessment class was creating properly structured explainability info including confidence score and geometry for each attribute..  
BUT, we discovered it was sometimes missing deeply nested list attributes. Rather than risk causing document FAILED states after upgrading to the version with this new validation logic, I opted to add the option to enable/disable the validation.. It's disabled by default (best effort assessment), but can be enabled if assessment is considered critical and must be right for the document to be considered successfully processed.

Open to suggestions to make it all less cluttered / confusing..   The best option would be to have the assessment never fail to properly assess every attribute correctly, so the validation just stays enabled always  - Fancy helping dig into that?  :)

HatmanStack: I wouldn't say it bothers me That much

lib/idp_common_pkg/idp_common/ocr/README.md
rstrahan: Any update on this?

lib/idp_common_pkg/idp_common/text_extraction/service.py
rstrahan: I'd rather keep it all abstracted behind the OCR class..  I feel this is a feature of OCR - not a new top level class.

patterns/pattern-2/template.yaml
rstrahan: OK, let's resolve this

rstrahan:
What's the latest status on this @HatmanStack - are you still tinkering with it or do you feel it's solid and ready to merge?  Maybe add some notes on your (new feature + regression) testing to provide confidence.  Tx!

HatmanStack:
Going to take time to properly test and feel confident about not losing context with the hybrid text extraction.  Would also like to add some logic to filter non-pertinent images, logos etc...  Maybe by the weekend.

PR #63, opened 2025-09-23
Optional Classification and Extraction Steps

HatmanStack:
## Summary
Introduces conditional execution of classification and extraction steps for pattern2 and pattern3, enabling cost and performance optimizations for workflows that don't require full processing.

## Changes
- Implemented conditional routing in pattern2 and pattern3 
- Updated UI to display toggles for easy enable/disable 
- Updated documents to reflect changes

## Use Case
**Invoice Processing**: Process large volumes for archival with OCR-only, then selectively enable extraction ( reprocess ) for payment processing. Result: costs only incurred when detailed data extraction is needed.

## Benefits
- **Backward compatible**: Existing workflows unchanged. Assessed downstream compatibility - when extraction is skipped, returning `{"section_id": ..., "document": ...}` maintains the ProcessResultsFunction's data contract (only extracts "document" key). When classification is skipped, returning `{"document": ...OCR document...}` with empty sections array allows ProcessSections Map state to iterate over empty array and produce empty ExtractionResults, preserving workflow integrity.
- **Cost savings**: Skip expensive ML services when not needed  
- **Performance**: Reduce processing time from minutes to seconds for OCR-only workflows

## Tasks
- [X] Decouple summarization from main processing flow
- [X]  Decouple extraction from main processing flow
- [X]  Decouple classification from main processing flow
- [X]  Decouple ocr from main processing flow
- [X]  Decouple assessment from main processing flow


rstrahan:
What the status on this @HatmanStack  - did you resolve / answer all the questions - do you feel it's ready?   If so I can try to make some time today to inspect and possibly merge if there are no issues.

HatmanStack:
It feels close, I haven't checked on assessment interactions.  Would also like to add a few more notes on exactly what I was trying to accomplish on some edits.  Always worry about edge cases with this sort of thing.  Maybe this afternoon?

rstrahan:
Thanks.. no rush.. Might be best, since tomorrow is release day, to target merging it early next week to target next friday, rather than rush and risk breaking.  Anyway, let me know clearly here when you feel you've done due diligence (i.e. tried to anticipate all the ways it might break :) ) and yoy feel it's ready..  Tx!

rstrahan:
Great stuff! Will look on Monday.

Did you notice in yesterday's release I made classification and extraction optional in the context of the new 'edit sections' feature.  Consider if that condition check could simply be expanded to check your new enabled flag as well.. That could be quite elegant I think.

Assessment data is added to the same result.json add extracted data, in a nested object called explainability_info.

Cheers
Bob



Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

HatmanStack:
I folded everything in to the updated approach pulling all of the checks out of the services into patterns.  This does cause unecessary processing in some cases.  For instance when a user wants something like an OCR -> Summarize or OCR with a knowledge-base.  We're still processing classification and extraction steps at least once.  [ef94dab](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/pull/63/commits/ef94dab1f7d3d21ac1184314a45ac1e70d0f01d1) created isolated steps but the changes didn't fit well with your update.  I can take another run at isolating the steps for these use cases in a future PR.  For now this gets us to a tuneable state after the first run and preserves the intelligent processing logic.  I also added a check for a summary in the summarization service so it's always displayed in the UI.  Dig the npm lint check added to publish.  All the best.

lib/idp_common_pkg/idp_common/extraction/service.py
```diff
@@ -16,6 +16,7 @@
 
 from idp_common import bedrock, image, metrics, s3, utils
 from idp_common.models import Document
+from idp_common.models import Document
```
rstrahan: looks like a duplicate line added??

lib/idp_common_pkg/idp_common/assessment/granular_service.py
```diff
@@ -2018,4 +2018,4 @@ def _convert_error_list_to_string(self, errors) -> str:
             return f"{first_errors} and {len(errors) - 1} more errors"
         else:
             # Few errors - show all
-            return "; ".join(errors)
```
rstrahan, line 2021: Seems to be a lot of line removed / identical line added - which makes the PR harder to review..

lib/idp_common_pkg/pyproject.toml
```diff
@@ -65,6 +65,7 @@ extraction = [
 # Assessment module dependencies
 assessment = [
     "Pillow==11.2.1",  # For image handling
+    "PyMuPDF==1.25.5", # Need fitz for creating text confidence from raw text uri
```
rstrahan, line 68: Not sure I understand this addition.. can you clarify?

lib/idp_common_pkg/setup.py
```diff
@@ -44,6 +44,7 @@
     # Assessment module dependencies
     "assessment": [
         "Pillow==11.2.1",  # For image handling
+        "PyMuPDF==1.25.5", # Need fitz for creating text confidence from raw text uri
```
rstrahan, line 47: ditto

patterns/pattern-2/src/classification_function/index.py
```diff
@@ -34,6 +35,12 @@ def handler(event, context):
     # Use default=str to handle Decimal and other non-serializable types
     logger.info(f"Config: {json.dumps(config, default=str)}")
     
+    # Extract document from the OCR result - handle both compressed and uncompressed
```
rstrahan: Adding duplicated code here..    Please carefully check all the changes for errors like this.

image

patterns/pattern-2/src/classification_function/index.py
```diff
@@ -45,30 +52,34 @@ def handler(event, context):
...
-        document_service.update_document(document)
-        
-        # Prepare output with existing document data
-        response = {
-            "document": document.serialize_document(working_bucket, "classification_skip", logger)
-        }
-        
-        logger.info(f"Classification skipped - Response: {json.dumps(response, default=str)}")
-        return response
+        if not normalize_boolean_value(classification_config.get('tuning', False)):
```
rstrahan, line 62: Conditional logic looks incorrect..

```
    if pages_with_classification == len(document.pages) and len(document.pages) > 0:
        if not normalize_boolean_value(classification_config.get('tuning', False)):
            logger.info(f"Skipping classification for document {document.id} - all {len(document.pages)} pages already classified")
```

This will skip classification *only* if (a) all pages are already classified **AND** (b) `classification.tuning` is False..

1. What is `classification.tuning` - shouldn't it be `classification.enabled`?
2. You want to skip if either condition is True

patterns/pattern-2/src/classification_function/index.py
```diff
@@ -118,14 +123,15 @@ def handler(event, context):
     
     # Check if document processing completely failed or has critical page failures
     if document.status == Status.FAILED or failed_page_exceptions:
+        error_details = '; '.join(str(e) for e in getattr(document, 'errors', []) if e)
```
rstrahan: Clarify this change.. Looks unrelated to the Optional Classification and Extraction.

patterns/pattern-2/src/extraction_function/index.py
```diff
@@ -60,22 +61,24 @@ def handler(event, context):
...
     # Intelligent Extraction detection: Skip if section already has extraction data
     if section.extraction_result_uri and section.extraction_result_uri.strip():
-        logger.info(f"Skipping extraction for section {section_id} - already has extraction data: {section.extraction_result_uri}")
-        
-        # Return the section without processing
-        response = {
-            "section_id": section_id,
-            "document": full_document.serialize_document(working_bucket, f"extraction_skip_{section_id}", logger)
-        }
+        if not normalize_boolean_value(extraction_config.get('enabled', False)):
```
rstrahan: Again - this logic looks wrong.. Please reexamine..  It should skip if either condition holds, not both. 

```
    if section.extraction_result_uri and section.extraction_result_uri.strip():
        if not normalize_boolean_value(extraction_config.get('enabled', False)):
            logger.info(f"Skipping extraction for section {section_id} - already has extraction data: {section.extraction_result_uri}")         
```

patterns/pattern-2/src/ocr_function/index.py
```diff
@@ -43,32 +44,36 @@ def handler(event, context):
...
-        
-        # Prepare output with existing document data
-        working_bucket = os.environ.get('WORKING_BUCKET')
-        response = {
-            "document": document.serialize_document(working_bucket, "ocr_skip", logger)
-        }
-        
-        logger.info(f"OCR skipped - Response: {json.dumps(response, default=str)}")
-        return response
+        if not normalize_boolean_value(ocr_config.get('tuning', False)):
```
rstrahan, line 62: What this `tuning` config properly again.. Should be `enabled`?
And as above, please check this logic.. Needs to be an OR condition.  This logic seems very confusing to me..
If AI is helping you (as I suspect) please Do Not Trust it..  carefully review the code and ensure it makes sense.

patterns/pattern-2/template.yaml
```diff
@@ -150,6 +150,11 @@ Resources:
             required:
               - features
             properties:
+              tuning:
```
rstrahan, line 153: Confused by 'tuning'  - seems disconnected from enable true | false which is supposed to be the focus of this PR.  Please remove all this (looks like hallucination or scope creep) and ensure PR is laser focused on just the classification/extraction enable/disable toggle.

patterns/pattern-3/template.yaml
```diff
@@ -991,6 +1011,7 @@ Resources:
           GUARDRAIL_ID_AND_VERSION: !If [HasGuardrailConfig, !Sub "${BedrockGuardrailId}:${BedrockGuardrailVersion}", ""]
           LOG_LEVEL: !Ref LogLevel
           WORKING_BUCKET: !Ref WorkingBucket
+          OUTPUT_BUCKET: !Ref OutputBucket
```
rstrahan: Why are you adding OUTPUT_BUCKET to each function here?  I see no related changes that require it, nor does it seem related to the scope of the PR.

lib/idp_common_pkg/idp_common/summarization/service.py
```diff
@@ -405,6 +405,13 @@ def process_document(
             # Update document status to completed if not already failed
             if document.status != Status.FAILED:
                 document.status = Status.COMPLETED
+            try:
+                # Perserving Document Summary for Frontend Across DeCoupled Architecture
```
rstrahan: Please clarify the purpose of this new logic.. the comment doesn't explain it clearly to me.

lib/idp_common_pkg/idp_common/extraction/service.py
HatmanStack: may have been duplicates from a interactive rebase.  The only service that should be changed now comes is summarization.  With the added check for an existing summary.  All others are from main.

patterns/pattern-3/template.yaml
HatmanStack: Only needed in summarization changed

lib/idp_common_pkg/idp_common/summarization/service.py
HatmanStack: reworked comment

patterns/pattern-2/src/ocr_function/index.py
HatmanStack: enabled didn't seem like the right fit for describing this feature as by default it's always enabled and then checked for existence on new runs.  The PR allows us to decouple from the checks.

patterns/pattern-2/src/extraction_function/index.py
HatmanStack: Thanks for finding this!

lib/idp_common_pkg/pyproject.toml
HatmanStack: Line 1220 of granular service is a fallback for creating an OCR service for textConfidence from raw text uri.  We don't have access to PyMuPDF for the OCR service call.

HatmanStack:
@rstrahan Ready for review. This should be a fast review. Planning to revisit and turn off these features after your changes have settled.

docs/pattern-2.md
```diff
@@ -225,6 +225,9 @@ The pattern exports these outputs to the parent stack:
 
 **Stack Deployment Parameters:**
 - `ClassificationMethod`: Classification methodology to use (options: 'multimodalPageLevelClassification' or 'textbasedHolisticClassification')
+- **OCR**: Control the ability to tune ocr via configuration file `ocr.tuning` property.
```
rstrahan, line 228: I'm afraid I am confused now by the scope of this PR, including the introduction of this 'tuning' concept.  I had thought this would be a simple PR, a quick win, to introduce only an 'enable' property for classification and extraction, with the simple logic in the associated lambdas to skip the associated processing if enabled is False.   (Like we have already for Summarization).  
And so I am confused by why this would require introduction of new stack deployment parameters and these .tuning properties.

lib/idp_common_pkg/idp_common/classification/service.py
```diff
@@ -2147,4 +2147,4 @@ def holistic_classify_document(self, document: Document) -> Document:
             )
             raise
 
-        return document
```
rstrahan, line 2150: Try to get rid of all the 'noisy' diffs..  makes it much harder to review the important changes.

lib/idp_common_pkg/idp_common/summarization/service.py
```diff
@@ -398,13 +398,22 @@ def process_document(
...
+        output_bucket = os.environ["OUTPUT_BUCKET"]
         if not enabled:
             logger.info(
                 f"Summarization is disabled in configuration for document {document.id}, skipping processing"
             )
             # Update document status to completed if not already failed
             if document.status != Status.FAILED:
                 document.status = Status.COMPLETED
+            try:
+                # When we Reprocess a document and don't enable Summarization we lose the summary in the UI
```
rstrahan, line 410: I do not know how 'reprocessing a document' has anything to do with the scope of the PR, which was simply (as the title suggests) to make Classification and Extraction optional.. Simple, safe, uncomplicated..    
This seems to have evolved into something much more, which, while it may be great (if I understood it :) ) , is riskier and not the 'quick win' I'd hoped for.

lib/idp_common_pkg/pyproject.toml
rstrahan: I still don't understand it..   Or what it has to do with the scope of this PR.

patterns/pattern-2/src/classification_function/index.py
rstrahan: Unresolved..  Logic needs fixed, and let's get rid of 'tuning' and revert to the simple 'enabled' flag (same as we have for assessment and summarization)

patterns/pattern-2/src/ocr_function/index.py
rstrahan: See my earlier comments.. I think we have unfortunately diverged in understanding the scope of this PR..  Can we get back to scoping it to simply allow classification and extraction to be optionally disabled through an 'enable' flag in the config.

patterns/pattern-2/template.yaml
rstrahan: Not resolved - reopening.

rstrahan (changes requested):
I can't move this forward as is, mostly because the scope seems to be very different that I had initially expected.. See my comments. 
I cannot prioritize the time to fully understand and test the concepts, logic, risk, benefits of everything you are attempting in this expanded PR.
Sorry @HatmanStack .. If  you can pare it right back to the simple enable/disable behavior, it can be a low risk, quick win..

PR #93, opened 2025-10-22
Fix Agentic config reading

kazmer97:
Config Can return boolean values as strings, we need to handle the case
for bool and string values

### DIFF
--- a/lib/idp_common_pkg/idp_common/extraction/service.py
+++ b/lib/idp_common_pkg/idp_common/extraction/service.py
@@ -31,9 +31,6 @@ from idp_common.utils import extract_json_from_text
 logger = logging.getLogger(__name__)
 
 
-logger = logging.getLogger(__name__)
-
-
 class ExtractionService:
     """Service for extracting fields from documents using LLMs."""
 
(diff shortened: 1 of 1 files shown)


lib/idp_common_pkg/idp_common/extraction/service.py
```diff
@@ -1324,11 +1321,29 @@ def process_document_section(self, document: Document, section_id: str) -> Docum
...
+
+            agentic_enabled = (
+                isinstance(agentic_enabled, str) and agentic_enabled.lower() == "true"
+            ) or (isinstance(agentic_enabled, bool) and agentic_enabled)
+
+            agentic_review_agent_enabled = self.config.get("extraction", {}).get(
+                "review_agent", False
+            )
+
+            agentic_review_agent_enabled = (
```
rstrahan, line 1338: Use utils function - normalize_boolean_value..
https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/main/lib/idp_common_pkg/idp_common/utils/__init__.py#L255

PR #137, opened 2025-11-18
new assesment

kazmer97:
*Issue #, if available:*

*Description of changes:*


config_library/pattern-2/bank-statement-sample/config.yaml
```diff
@@ -404,11 +404,7 @@ assessment:
   image:
     target_height: ""
     target_width: ""
-  granular:
```
rstrahan, line 408: This doesn't look backward compatible. Is it?  We do not want to break the existing granular assessment. Or do you propose that we replace 'granular assessment' (our current default) with 'agentic assessment'?

config_library/pattern-2/bank-statement-sample/config.yaml
```diff
@@ -455,107 +451,6 @@ assessment:
     - Provide tight, accurate bounding boxes around the actual text
     </assessment-guidelines>
 
-    <spatial-localization-guidelines>
```
rstrahan, line 459: Again, won't removing this break the current assessment implementation?

lib/idp_common_pkg/idp_common/assessment/__init__.py
```diff
@@ -7,9 +7,8 @@
 This module provides services for assessing the confidence and accuracy of
 extraction results by analyzing them against source documents using LLMs.
 
-The module supports both:
-1. Original approach: Single inference for all attributes in a section
-2. Granular approach: Multiple focused inferences with caching and parallelization
+All assessment now uses the granular approach with Strands agents for
```
rstrahan, line 10: Ah, this answers my question above.. you *are* replacing granular assessment with agentic assessment..   High stakes - we have customers (like Deluxe) relying on this.. if the new method consistently works better and doesn't affect cost, then i agree this is cleaner, more maintainable, approach.. But you'll need to do tests to prove that.

src/ui/src/components/document-viewer/VisualEditorModal.jsx
rstrahan, line 1: Can you clarify the changes to the UI?  I was not expecting any UI changes, and hard to assess the nature/purpose from code read.  
Mostly, can you confirm that changes are backward compatible with processed docs using the previous method, and especially with pattern-1 BDA output..  Or put differently, can you confirm that the JSON structure for `explainability_info` in the results.json is unchanged?

rstrahan:
Closing this in favor of new branch that pulled from this one..

Issue #136, opened 2025-11-15
Bug: page_indices in extraction results always starts at [0] instead of preserving document position


mmiakashs:
## Problem
The `page_indices` field in extraction service output (`split_document.page_indices`) incorrectly always starts at `[0, 1, 2, ...]` for each section, regardless of which pages from the original document packet the section contains.

## Current Behavior
- Section with pages [5, 6, 7] outputs `page_indices: [0, 1, 2]` ❌
- Section with pages [10, 11, 12] outputs `page_indices: [0, 1, 2]` ❌

## Expected Behavior
- Section with pages [5, 6, 7] should output `page_indices: [4, 5, 6]` ✓
- Section with pages [10, 11, 12] should output `page_indices: [9, 10, 11]` ✓

<img width="904" height="901" alt="Image" src="https://github.com/user-attachments/assets/f47647ae-407e-4596-989e-0240f19886e2" />

PR #138, opened 2025-11-19
Agentic extraction better retry

kazmer97:
- **bedrock utils for retry**
- **invoke extraction with long retries**

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/idp_common_pkg/tests/unit/test_bedrock_utils.py
+++ b/lib/idp_common_pkg/tests/unit/test_bedrock_utils.py
@@ -38,8 +38,8 @@ class MockEventStreamError(MockClientError):
     pass
 
 
-# Now import the module under test
-from idp_common.utils.bedrock_utils import (
+# Now import the module under test (must be after mock classes are defined)
+from idp_common.utils.bedrock_utils import (  # noqa: E402
     async_exponential_backoff_retry,
     exponential_backoff_retry,
 )


patterns/pattern-2/template.yaml
```diff
@@ -925,6 +924,10 @@ Resources:
                     description: This introduces a second agent to review the first agents work. Only use with highly complex workflows as it increases token usage.
                     order: 1
                     default: false
+                  review_agent_model:
+                    type: string
+                    description: Model to review the initial extraction agents work and correct it if needed, if not specified will default to the same as the extraction model.
+                    default: Null
```
rstrahan: This is invalid - CloudFormation fails with: 
```
[/Resources/UpdateSchemaConfig/Type/Schema/properties/extraction/properties/agentic/properties/review_agent_model/default] 'null' values are not allowed in templates
```

patterns/pattern-2/template.yaml
rstrahan, line 1: Please include pattern-3 as well.. these fixes apply to both patterns.

lib/idp_common_pkg/idp_common/extraction/service.py
rstrahan, line 1: Getting merge conflicts in this file

rstrahan: Also huge volume of changes that seem unrelated to 'better retries' - I can't review them..

rstrahan (changes requested):
Three main comments

1. There's a new conflict with the develop branch after latest merge of fix for #136   

2. When I deploy this branch and process a document in the stack, using agentic extraction, I get error below.. *Can you do a clean publish and deploy, and see if you also get this error after enabling agentic extraction in the config.*
```
Error: ImportError: Agentic extraction requires Python 3.10+ and strands-agents dependencies. Install with: pip install 'idp_common[agents]' or use agentic=False
```


3. I'm overwhelmed by the volume of changes in this PR - many of which seem formatter related  and unrelated to the 'better retry' focus of the PR, but the effect is to hinder my ability to review the changes.   
*Can you clean up the changes to include only the changes related to the retry improvements, tx.*

Issue #43, opened 2025-08-28
Add few-shot prompting with RAG and LLM pattern


ArlindNocaj:
**Is your feature request related to a problem? Please describe.**
Fine-tuning and customizing blueprints requires a lot of effort. Few-Shot prompting could allow to improve results based on historic documents.

**Describe the solution you'd like**
Add few-shot prompting approach in which we utilize a set of historic document/json groundtruth pairs for learning on the fly how to perform the extraction. 

Main Tasks:
* implement either knowledge_base or other similarity function to identify a most similar document from a corpus
  * a) S3 vector store with normal embeddings, or multi-modal embeddings (or multi-vector embeddings like ColQwen)
  * b) Bedrock Knowledge base 
  * c) FAISS in memory database
  * d) simple tf-idf metrics using the Textract or BDA standard output results

* Implement new Pattern 4: Few-shot context engineering with LLM
* When extracting information search for other relevant documents and take the top3 for few-shot prompting
* Run Image to JSON using structured output of LLM, e.g. Nova model with constrained decoding
* Add the top3 docs to the context of the inference.

* We can also allow a hook in the lambda to have custom way of identifying relevant documents.

<img width="879" height="470" alt="Image" src="https://github.com/user-attachments/assets/95a4cbd3-5f08-4969-8faf-624550c4d010" />

**Describe alternatives you've considered**
Fine-tuning approach like: Pattern 3, or https://github.com/aws-samples/sample-for-multi-modal-document-to-json-with-sagemaker-ai

**Additional context**

rstrahan:
@ArlindNocaj Do you have any data or examples that demonstrate the efficacy of this approach, or specific use cases that would raise the priority of implementing it?

lorchda:
Hi @rstrahan, implementation is still ongoing.

PR #149, opened 2025-12-04
Dynamic Few-Shot Prompting with RAG and LLM using Amazon S3 Vectors

lorchda:
*Issue #43*

*Description of changes:*

- Extended `class BedrockClient` to generate embeddings for Amazon Titan Multimodal Embeddings G1 and Amazon Nova Multimodal Embeddings 
- Added an example Lambda to return few-shot examples from a lookup to Amazon S3 Vectors using the existing Custom Prompt Lambda interface
- Added a notebook to ingest the fcc_invoices dataset (REALKIE) along with ground truth data into Amazon S3 Vectors

### DIFF
--- a/plugins/dynamic-few-shot-lambda/template.yml
+++ b/plugins/dynamic-few-shot-lambda/template.yml
@@ -385,3 +385,11 @@ Outputs:
     Description: Direct link to CloudWatch logs for this function
     Value: !Sub |
       https://console.aws.amazon.com/cloudwatch/home?region=${AWS::Region}#logsV2:log-groups/log-group/$252Faws$252Flambda$252F${LambdaFunctionName}
+
+  ModelId:
+    Description: Vector embedding model to use to create meaningful vector representations of documents
+    Value: !Ref ModelId
+
+  VectorDimensions:
+    Description: Vector embedding length to use, as defined by the embedding model in use
+    Value: !Ref VectorDimensions


plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+```bash
+aws cloudformation describe-stacks \
+    --stack-name GENAIIDP-dynamic-few-shot-stack \
+    --query 'Stacks[0].Outputs[?OutputKey==`DynamicFewShotFunctionArn`].OutputValue' \
+    --output text
+```
+
+### Step 3: Populate the Examples Dataset
+
+Use the [fewshot_dataset_import.ipynb](notebooks/fewshot_dataset_import.ipynb) notebook to import a dataset into S3 Vectors, or manually upload your example documents and metadata to the S3 bucket and vector index created by the stack.
```
rstrahan, line 102: Looks like this is fcc_invoices_dataset_import.ipynb now

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+    end
+
+    H -.-> X
+    H -.-> Y
+    H -.-> Z
+```
+
+## Quick Start
+
+### Step 1: Deploy the Dynamic-few shot Stack
```
rstrahan, line 79: State prerequisites, since user must clone repo and have tools (sam, aws cli) deployed.

plugins/dynamic-few-shot-lambda/README.md
rstrahan: Should you have more guidance for user to prepare and load their own dataset?

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+    --output text
+```
+
+### Step 3: Populate the Examples Dataset
+
+Use the [fewshot_dataset_import.ipynb](notebooks/fewshot_dataset_import.ipynb) notebook to import a dataset into S3 Vectors, or manually upload your example documents and metadata to the S3 bucket and vector index created by the stack.
+
+### Step 4: Configure IDP to Use Dynamic Few-Shot
+
+Add the Lambda ARN to your IDP extraction configuration:
```
rstrahan, line 106: Show how to configure the Lambda ARN in the UI (screenshot)

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+
+```yaml
+extraction:
+  custom_prompt_lambda_arn: "arn:aws:lambda:region:account:function:GENAIIDP-dynamic-few-shot"
+```
+
+**Important**: Your extraction task prompt must include the `{FEW_SHOT_EXAMPLES}` placeholder where you want the dynamic examples to be inserted.
```
rstrahan, line 113: I think we need to be more prescriptive here.  How do we make this easy / foolproof?
Eg: Should all our extraction prompt examples contain the `{FEW_SHOT_EXAMPLES}` placeholder?  Does it do no harm if there are no examples?

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+
+### Step 5: Run the Demo Notebook
+
+0. Run `notebooks/examples` steps 0, 1, 2
+1. Open `plugins/dynamic-few-shot-lambda/notebooks/step3_extraction_with_custom_lambda.ipynb`
```
rstrahan: This notebook is missing.  Maybe you forgot to commit it?
Is it meant to integrate with steps 0-2 from <ROOT>/notebooks/examples ?  (these create temp files in that directory).. Might be better to create a standalone demo notebook for this plugin?

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+    "pages": {...},
+    "sections": [...],
+    ...
+  }
+}
+```
+
+### Output Payload Structure
+
+The Lambda returns modified prompt content with dynamic few-shot examples:
```
rstrahan, line 157: Can you make your demo notebook display the images / text of the chosen examples, and compare to the image / text for the document being processed?  
It will be powerful, and helpfully, to be able to visually anchor the selected examples.. Do this for two or three visually different invoices so it's easy to see that the embeddings are selecting similar examples.
You can use this in your demo as well to highlight how the example selection works.. visually.
Include screenshots of these examples (original docs + selected examples for each) in this README, for impact.

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+The Lambda integrates with IDP's custom prompt system by:
+- Receiving the full extraction context and configuration
+- Processing the `{FEW_SHOT_EXAMPLES}` placeholder in task prompts
+- Returning modified prompt content with dynamically retrieved examples
+
+### 2. Vector Similarity Search
+
+The Lambda uses Amazon Nova multimodal embeddings to find similar examples:
+
+```python
```
rstrahan, line 184: Is there any value including code snippets in the README?  IMHO this doesn't enhance the readability / understanding, and the code snippets become out of date as we maintain the code..  Suggest remove code from README and focus it on explaining, illustrating the feature, and how and when to use it.

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+- **Metadata Configuration**: Stores `classPrompt`, `attributesPrompt`, and `imagePath` as non-filterable metadata keys
+
+## Monitoring and Troubleshooting
+
+### CloudWatch Logs
+
+Monitor the Lambda function logs:
+- `/aws/lambda/GENAIIDP-dynamic-few-shot` - Dynamic few-shot Lambda logs
+
+### Key Log Messages
```
rstrahan, line 263: This seems very perfunctory / not really useful in the README.. is it?  Try to make it a bit more enlightening / actionable.. or remove.

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+=== EXTRACTION CONFIG ===
+Model: anthropic.claude-3-5-sonnet-20241022-v2:0
+=== HANDLE INPUT DOCUMENT ===
+=== OUTPUT ANALYSIS ===
+Output keys: ['system_prompt', 'task_prompt_content']
+Task prompt content items: 5
+=== DYNAMIC FEW-SHOT LAMBDA COMPLETED ===
+```
+
+**Error Conditions:**
```
rstrahan, line 277: Same comment as above

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+
+### Performance Monitoring
+
+Key metrics to monitor:
+- **Lambda Duration**: Time to retrieve and process examples
+- **S3 Vectors Query Time**: Vector similarity search performance
+- **Example Count**: Number of examples returned per request
+- **Error Rate**: Failed example retrievals
+
+## Example Dataset Structure
```
rstrahan, line 293: This feels like an important section, but it needs fleshing out..  Include clear instructive and illustrative test to guide a user through the steps.. OR point them to an example notebook that already does it, and ensure that the notebook is well documented so user could clone and use for their own datasets.

plugins/dynamic-few-shot-lambda/README.md
```diff
@@ -0,0 +1,454 @@
...
+
+### Vector Metadata Format
+
+Each vector in the S3 Vectors index should have metadata:
```
rstrahan, line 297: This in confusing.. how is user supposed to understand and create this metadata?

plugins/dynamic-few-shot-lambda/README.md
rstrahan, line 1: I think this README needs to be reworked. It seems very fragmented, no clear structure, too many perfunctory bullet lists.. It needs to be a clear and engaging document - suggested structure:
- Include Table of Contents at the top
- Introduction to the feature / concepts
   - clarify that it focuses on extraction only, for now..  address if it can (or will in future) also apply to classification?
- When to use it and when not
- How to set it up with the built in example dataset (clear step by step)
- How to adapt it to your own data set (clear step by step)
- How to measure accuracy and costs (using new Test Studio)
  - how to assess additional costs incurred by the custom lambda / embeddings model?
     - should we have custom lambda return metering info that our extraction service integrates into overall metering/costs?
- How to maintain your 'examples' vector store over time
   - Is there a way to automatically use high confidence predictions from the accelerator to populate the 'examples'? (Future enhancement?)
- Troubleshooting problems
  - how to check if it's selecting relevant / useful examples?
  - how to find and troubleshoot failures (lambda logs, CW Dashboard?)

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
```diff
@@ -0,0 +1,416 @@
...
+        # Handle input document
+        result = _handle_input_document(placeholders, default_system_prompt, default_task_prompt)
+
+        # Log complete output structure
+        logger.info(f"=== OUTPUT ANALYSIS ===")
+        logger.info(f"Output keys: {list(result.keys())}")
+        logger.info(f"System prompt length: {len(result.get('system_prompt', ''))}")
+        logger.info(f"System prompt (first 200 chars): {result.get('system_prompt', '')[:200]}...")
+        
+        task_content = result.get('task_prompt_content', [])
```
rstrahan, line 94: Will first three items of the task prompt identify the selected few shot examples?  I suspect not, but this is the important thing to log for troubleshooting.. Please make sure it logs the specific few_shot examples that were inserted.

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
```diff
@@ -0,0 +1,416 @@
...
+    if "{FEW_SHOT_EXAMPLES}" in prompt_template:
+        parts = prompt_template.split("{FEW_SHOT_EXAMPLES}")
+        if len(parts) == 2:
+            # Process before examples
+            content.extend(
+                _build_text_and_image_content(parts[0], substitutions, image_content)
+            )
+
+            # Add few-shot examples
+            content.extend(_build_few_shot_examples_content(image_content))
```
rstrahan, line 173: Hmm.. Are you locating few shot examples based only on the image.. not the image + the document text?  
Maybe that's fine, but can you clarify since intuitively I'd have expected text content to be a major component of the semantic match.

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
```diff
@@ -0,0 +1,416 @@
...
+
+            # Process after examples (only pass images if not already used)
+            image_for_after = (
```
rstrahan, line 176: Why is this needed?  Seems you could just pass `image_content` and it'll be ignored unless there's a placeholder..  Why treat images different than other placeholders here?

plugins/dynamic-few-shot-lambda/src/GENAIIDP-dynamic-few-shot.py
```diff
@@ -0,0 +1,416 @@
...
+            f"Input event keys: {list(event.keys()) if 'event' in locals() else 'Unknown'}"
+        )
+        # In demo, we'll fail gracefully with detailed error info
+        raise Exception(f"Dynamic few-shot Lambda failed: {str(e)}")
+
+def _handle_input_document(placeholders, default_system_prompt, default_task_prompt):
+    """
+    Handle input request and return custom system_prompt and task_prompt_content
+    """
+    substitutions = {
```
rstrahan: Why not include DOCUMENT_IMAGE in substitutions?  Not sure why images are any different than text in the substitutions?  The code would be cleaner if you treat them the same.

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+  PermissionsBoundaryArn:
+    Type: String
+    Default: ""
+    Description: >-
+      (Optional) ARN of an existing IAM Permissions Boundary policy to attach to the Lambda execution role.
+      Leave blank if no Permissions Boundary is required.
+    AllowedPattern: "^(|arn:aws[a-z-]*::iam::[0-9]{12}:policy/.+)$"
+    ConstraintDescription: Must be empty or a valid IAM policy ARN
+
+  VectorBucketName:
```
rstrahan: It's unclear to me where/how this VectorBucket gets created..  Is it presumed to pre-exist before deploying the plugin?  Where are the user instructions for how to create it? 
Shouldn't your stack create it, if it's not pre-created?  (Eg see how our main stack handles EvaluationBaselineBucket or ReportingBucket - user can optionally provide a precreated bucket, but if not, stack creates one automatically)

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+    Type: String
+    Default: "genaiidp-dynamic-few-shot"
```
rstrahan: We can't use a default bucket name.. Bucket names need to be globally unique so this is guaranteed to fail if deployed an second time (by anybody anywhere)

See how we handle ReportingBucket, EvaluationBaseline bucket in main template.yaml..  User can optionally provide their own bucket (which allows sharing across multiple stacks), but if left blank, then stack creates a new bucket with a dynamically allocated unique name.

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - S3ReadPolicy:
+            BucketName: !Ref GenAIIDPS3OutputBucketName
+        - S3ReadPolicy:
+            BucketName: !Ref GenAIIDPS3WorkingBucketName
+        - Statement:
+            - Effect: Allow
+              Action: cloudwatch:PutMetricData
+              Resource: "*"
```
rstrahan, line 186: Need to add cfn_nag suppression to avoid scan complaining about the * for CW.

          - id: W11
            reason: "Cloudwatch does not support resource-level permissions"

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+                - kms:GenerateDataKey*
+                - kms:DescribeKey
+              Resource:
+                - !Ref GenAIIDPCustomerManagedEncryptionKeyArn
+
+  DynamicFewShotLogGroup:
+    Type: AWS::Logs::LogGroup
+    Metadata:
+      cfn_nag:
+        rules_to_suppress:
```
rstrahan: Since you already have parameter `GenAIIDPCustomerManagedEncryptionKeyArn` use it for this Log Group, and remove this cfn_nag suppression.

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+          - id: W84
+            reason: "Demo function - KMS CMK not required, but can be added by customer for production use cases"
+    # checkov:skip=CKV_AWS_158: "Demo function - KMS CMK not required, but can be added by customer for production use cases"
+    Properties:
+      LogGroupName: !Sub "/aws/lambda/${LambdaFunctionName}"
+      RetentionInDays: 7  # Short retention for demo purposes
```
rstrahan: Use stack param - !Ref LogRetentionDays - for this.. don't hardcode it..

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+
+  DynamicFewShotVectorBucket:
+    Type: AWS::S3Vectors::VectorBucket
+    Metadata:
+      cfn_nag:
+        rules_to_suppress:
+          - id: W84
+            reason: "Demo function - KMS CMK not required, but can be added by customer for production use cases"
```
rstrahan: Since you already have parameter GenAIIDPCustomerManagedEncryptionKeyArn use it here

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+    # checkov:skip=CKV_AWS_158: "Demo function - KMS CMK not required, but can be added by customer for production use cases"
+    Properties:
+      VectorBucketName: !Ref VectorBucketName
```
rstrahan: See my comment above about bucketname..   
- we should use customer provided existing bucket (if provided) OR create a new bucket with dynamic name.

plugins/dynamic-few-shot-lambda/template.yml
```diff
@@ -0,0 +1,216 @@
...
+    Value: !Ref DynamicFewShotDatasetBucket
+
+  UsageInstructions:
+    Description: How to use this Lambda in your IDP configuration
+    Value: !Sub |
+      Add this ARN to your extraction config:
+      extraction:
+        dynamic_few_shot_lambda_arn: "${DynamicFewShotFunction.Arn}"
+        
+  MonitoringLink:
```
rstrahan, line 384: Nice touch..  A CW Dashboard could be even better, to track number of inferences, successful matches, throttles/retries, etc.

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
```diff
@@ -0,0 +1,416 @@
...
+            )
+        else:
+            filtered_result.append(example)
+
+    return filtered_result
+
+
+def _s3vectors_find_similar_items_from_image(page_image):
+    """Search for similar items using image query"""
+    embedding = bedrock_client.generate_embedding(
```
rstrahan, line 348: This will incur token usage and costs
Can you minimally capture and log token use?
Even better, can you create metering data structure and return in lambda response.. and enhance extraction function to merge this metering into overall metering.. It can be a new 'context' (eg 'FewShotExamplesPlugin') - that way it will show up in the document Cost Estimates and Benchmark results for accurate cost transparency.

plugins/dynamic-few-shot-lambda/README.md
lorchda: In the Lambda, I currently have:
```
logger.warn("Missing {FEW_SHOT_EXAMPLES} placeholder in prompt template")
```

What I could do is to replace this by:
```
raise ValueError("Missing {FEW_SHOT_EXAMPLES} placeholder in prompt template")
```

This will stop the extraction process and the error will propagate to the UI if `{FEW_SHOT_EXAMPLES}` is missing.

As for adding `{FEW_SHOT_EXAMPLES}` to all templates - yes, I think that could be done. It should not do any harm, if there are no examples available, the placeholder will just be empty.

plugins/dynamic-few-shot-lambda/README.md
lorchda: Thanks for spotting, it was a wrong path in the README. The notebook `notebooks/examples/step3_extraction_with_custom_lambda.ipynb` already has a `DEMO_LAMBDA_ARN` configuration option that can be leveraged. It's fixed.

plugins/dynamic-few-shot-lambda/template.yml
lorchda: For general-purpose S3 buckets no, but the the S3 vector buckets namespace is scoped to account and region. You could have the same S3 vectors bucket name in the same AWS account in two different regions. 

If you leave out the `VectorBucketName` in the `AWS::S3Vectors::VectorBucket` resource, then a randomly chosen name will be provided. But then you cannot `!GetAtt MyVectorBucket.VectorBucketName` because that results in an `Requested attribute VectorBucketName must be a readonly property in schema for AWS::S3Vectors::VectorBucket` error (a current limitation of the service). You *could* still retrieve the vector bucket name by using `!Split` on the ARN as a workaround.

Right now, when running through `sam deploy --guided`, the above will be presented as a suggestion to the user. It can then be overridden by changing the name.

plugins/dynamic-few-shot-lambda/template.yml
lorchda: When running `sam deploy --guided`, the customer can choose a name. This is then stored in `samconfig.toml` as a `parameter_overrides` key/value pair. The above name is only a suggestion presented to the user.

plugins/dynamic-few-shot-lambda/template.yml
lorchda: Good point, fixed now:
```
% cfn_nag_scan --input-path plugins/dynamic-few-shot-lambda/template.yml
------------------------------------------------------------
plugins/dynamic-few-shot-lambda/template.yml
------------------------------------------------------------
Failures count: 0
Warnings count: 0
```

plugins/dynamic-few-shot-lambda/template.yml
lorchda: Fixed

plugins/dynamic-few-shot-lambda/template.yml
lorchda: Good point, fixed.

plugins/dynamic-few-shot-lambda/template.yml
lorchda: That will only work if the condition `IsS3VectorsVectorStore` is true in the main template. This will add the S3 vectors service principal and required actions to the KMS key:

https://github.com/lorchda/accelerated-intelligent-document-processing-on-aws/blob/87feb024848cc2e13a30e29a0e05e353a26b917f/template.yaml#L1477-L1490

plugins/dynamic-few-shot-lambda/README.md
rstrahan: No, I don't think we should enforce the placeholder be there.. I'm just suggesting that we should add it to the default extraction task_prompt in our configs.

plugins/dynamic-few-shot-lambda/template.yml
rstrahan: Good point. We could resolve by removing the condition in the main stack do that the S3 vectors service principal is always added?    Or (maybe better) you can create your own CMK in your stack for use with your own storage resources - vector store, Log Groups, etc.

plugins/dynamic-few-shot-lambda/template.yml
rstrahan: Right, but your template will always create a new bucket, and will fail if the bucket name is already in use.
I'm suggesting that you allow user to provide an existing bucket (with index already in there) so that few shot example vectorstore can be shared by multiple IDP stacks, as a convenience.   OR, they can leave buckename blank to have stack create one.
We should not require user to enter a bucketname for a new bucket, imho.. better to auto-gen the bucketname,

plugins/dynamic-few-shot-lambda/template.yml
lorchda: I have ended up creating my own KMS key, which is used for CW log groups, the dataset bucket (new) and the S3 vectors bucket.

plugins/dynamic-few-shot-lambda/template.yml
lorchda: Understood, this is implemented now - the user can leave `VectorBucketName`, `VectorIndexName` and `DatasetBucketName` blank and the stack will auto-create them. If provided, the stack will re-use existing resources.

plugins/dynamic-few-shot-lambda/README.md
lorchda: This is how a sample output looks like from the "7. Verify Import with Similarity Search" section

Screenshot 2025-12-16 at 17 46 28

I will add more examples to the README.

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
lorchda: That's a good question. This is based on existing program logic here:

https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/87feb024848cc2e13a30e29a0e05e353a26b917f/lib/idp_common_pkg/idp_common/extraction/service.py#L279-L298

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
lorchda: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html used for generating embeddings does not seem to return token usage.

An approach could be to run a second call to https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/count_tokens.html to retrieve token usage. I will look into it.

plugins/dynamic-few-shot-lambda/template.yml
lorchda: This is implemented now - the user can leave VectorBucketName, VectorIndexName and DatasetBucketName blank and the stack will auto-create them. If provided, the stack will re-use existing resources.

plugins/dynamic-few-shot-lambda/src/IDP-dynamic-few-shot.py
lorchda: Yes, I've only used the image for similarity search. My intuition is that it would capture similar layouts (vendors), as opposed to the text elements, which would keep changing (different company name, different line items).

I have added some examples here in "7. Verify Import with Similarity Search" for review: https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/a3a709544b147e301d3ee2d0ecb70de407dfc80a/plugins/dynamic-few-shot-lambda/notebooks/fcc_invoices_dataset_import.ipynb

plugins/dynamic-few-shot-lambda/README.md
lorchda: Added here in "7. Verify Import with Similarity Search" for review: https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/a3a709544b147e301d3ee2d0ecb70de407dfc80a/plugins/dynamic-few-shot-lambda/notebooks/fcc_invoices_dataset_import.ipynb

The same images will be added to the README.

Issue #157, opened 2025-12-11
Add support to specify service_tier keyword for supported models


khanrubd:
**Is your feature request related to a problem? Please describe.**
Currently, token consumption is on-demand at the default tier, and users cannot specify flex (slower, cheaper), priority (faster, pricier) or provisioned tiers. The request is to add an option to specify service tiers.

**Describe the solution you'd like**
Ideally, end users should be able to select from the UI (or as cli option) the service tier they want to use. This can also be be turned into a deployment time selection, although that would run into issues around not all models being compatible with all service tiers. 

**Describe alternatives you've considered**
Looked into whether this can be set at Bedrock service level or at AWS account level (e.g., all token consumption defaults to flex tier where available). Does not look like it is viable yet. Only docs I could find talk about limiting access to certain tiers etc. (https://docs.aws.amazon.com/bedrock/latest/userguide/security_iam_id-based-policy-examples-agent.html#security_iam_id-based-policy-examples-service-tiers)

**Additional context**
Amazon Bedrock runtime APIs now support specifying am optional service_tier keyword (https://aws.amazon.com/about-aws/whats-new/2025/11/amazon-bedrock-priority-flex-inference-service-tiers/). Specifically for batch mode IDP jobs, the "Flex Tier" (https://docs.aws.amazon.com/bedrock/latest/userguide/service-tiers-inference.html#w2aac21c29c11)" may be an attractive option, and it makes a real difference. For example, Nova 2 Pro at the default (standard) tier is pricier than Nova 1 Pro, but is cheaper at the Flex Tier. Likewise, Nova 2 Lite price varies by 2x (https://aws.amazon.com/nova/pricing/) across standard and flex tiers.

(So far I can tell, service_tier is an optional keyword for the Bedrock runtime SDKs / APIs, unsure about BDA though.)

PR #158, opened 2025-12-11
Feature/service tiers

khanrubd:
*Issue #157*

*Added support to specify service_tier keyword for supported models*

This is a work in progress and has not been deploy-tested yet.


khanrubd:
@rstrahan Thank you for the feedback! The <model_id>:<tier> approach _can_ be a viable option.

However, this will create model choice drop-down menu sprawl throughout the config UI. Each serviceTier supported model will end up with 3 entries: no-suffix (standard), flex-suffix, priority-suffix. Just for the three Nova2 models (Lite, Pro, Omni) we will end up with 9 entries. I am also guessing that over time more models will become supported. (This is not even to consider the "provisioned" tier which requires separate setup and are supported for Anthropic models etc.).

The "1m" context is a one off special case I guess, even for that Claude code does it differently("<model_id>:[1m]"). For CRIS: we can make "global", "us", "eu" etc. prefixes part of model ID (as Bedrock does) but still need to expose that optionality, since not all customers find global CRIS acceptable for data residency reasons.

Some alternative proposals:
1. Add a doc/UI disclaimer that if a non-standard serviceTier is chosen then _all_ models chosen as part of a config set has to support this, else processing will fail. OR
2. Before making relevant Bedrock API calls, the code will use a validator to check if non-standard serviceTier is allowed for the requested model, and if not then revert the serviceTier setting to standard for that call.

Looking forward to discuss :-)

PR #260, opened 2026-03-30
Feature/cli publish artifact bucket hardening

sirirako:
## Overview

Adds two optional enterprise hardening flags to `idp-cli publish` for organizations that require KMS-encrypted S3 buckets and cost-allocation tagging on AWS resources.

### `lib/idp_cli_pkg/idp_cli/cli.py`
- Added `--artifacts-bucket-kms-key-arn` option to the `publish` command — passes the ARN of a customer-managed KMS key to the SDK publish operation
- Added `--artifacts-bucket-tags` option — passes comma-separated `Key=Value` tag pairs to the SDK publish operation

### `lib/idp_sdk/idp_sdk/operations/publish.py`
- Added `artifacts_bucket_kms_key_arn` parameter to `PublishOperation.build()` — after a successful build, calls `s3.put_bucket_encryption()` with `SSEAlgorithm: aws:kms` and `KMSMasterKeyID` set to the provided CMK ARN
- Added `artifacts_bucket_tags` parameter to `PublishOperation.build()` — after a successful build, calls `s3.put_bucket_tagging()` with the parsed tag set

### `docs/idp-cli.md`
- Added `--artifacts-bucket-kms-key-arn` and `--artifacts-bucket-tags` flags to the `idp-cli publish` reference documentation with usage examples

### `docs/idp-sdk.md`
- Added `artifacts_bucket_kms_key_arn` and `artifacts_bucket_tags` parameters to the `client.publish.build()` SDK reference documentation

### `lib/idp_sdk/tests/unit/test_publish_operations.py`
- Added unit tests covering:
  - KMS encryption applied after successful build
  - Tags applied after successful build
  - Both flags used together
  - No encryption/tagging when flags are not provided (backward compatibility)

## Usage

```bash
# With KMS encryption and cost-allocation tags
idp-cli publish \
  --source-dir . \
  --bucket-basename my-artifacts \
  --prefix idp \
  --region us-east-1 \
  --artifacts-bucket-kms-key-arn arn:aws:kms:us-east-1:123456789012:key/key-id \
  --artifacts-bucket-tags CostCenter=123,Project=IDP,Environment=production
```

## Notes

- Both flags are **optional** — existing behavior is unchanged when omitted (backward compatible)
- The KMS key must already exist and the caller's IAM role must have `kms:GenerateDataKey`, `kms:Decrypt`, and `s3:PutBucketEncryption` permissions
- Encryption and tagging are applied **after** the build completes so the bucket is hardened before artifacts are written on subsequent runs
- When deploying a stack after publishing with `--artifacts-bucket-kms-key-arn`, pass the same KMS key ARN as `ArtifactsBucketKmsKeyArn` in the deploy parameters so CodeBuild and Lambda roles receive `kms:Decrypt` permission


rstrahan:
Thanks @sirirako !

The idp-cli publish and deploy commands support use of preexisting buckets that can be pre-created by customer to address any customer specific compliance requirements such as tagging encryption, logging, bucket policies, block public access policies, etc etc.. 

See

```
  --bucket-basename TEXT  S3 bucket basename for
                          artifacts — region is
                          appended automatically
                          (auto-generated if not
                          provided)
  --prefix TEXT           S3 key prefix for
                          artifacts (default: idp-
                          cli)
```



I feel it's reasonable to ask customer to pre-create compliant buckets.  
Given this I'm reluctant to add complexity to our cli specifically to add tags and kms keys (as subset of possible customer constraints)

Does pre-creating work for your scenario?

sirirako:
So you do not want this change at all?

PR #261, opened 2026-03-30
Added restrictions to the Class Name input according to requirements …

ccl19:
…from BDA

*Issue #, if available:*

*Description of changes:*
All class names need to follow [a-zA-Z0-9-_]+ pattern to sync to BDA. Added a function which checks whether the class name is valid in Schema Builder. If not, it will spring an error to notify the user.


rstrahan:
Love the contribution @ccl19 - Will be in tomorrow's release.

PR #298, opened 2026-04-29
Update/main 0.58 clean

alig4lfe:
# feat: Headless REST API mode with VPC-secured deployment for GovCloud

**Target:** `origin/main` (aws-solutions-library-samples)
**Source:** `update/main-0.58-clean` on `nicklynberg/accelerated-intelligent-document-processing-on-aws`

## Summary

Adds a **headless REST API mode** for the IDP accelerator that lets you submit documents for processing through an authenticated `/jobs` REST API instead of (or alongside) the existing S3-drop workflow, plus a **VPC-secured deployment option** with optional bastion host that keeps the API off the public internet.

This is what enables the IDP accelerator to be deployed into **AWS GovCloud** end-to-end without the UI/AppSync stack. The existing `--headless` template transformer strips the UI layer but didn't give you a way to submit documents programmatically without direct S3 access; this PR fills that gap.

The feature set was developed and used internally against a real GovCloud deployment in `us-gov-west-1` for the past ~3 months. Both fresh-stack creation and existing-stack update paths are exercised end-to-end (see **Testing** below).

## Motivation

GovCloud customers currently have two options to get documents into IDP:

1. `aws s3 cp` directly into the `InputBucket` — requires handing out S3 write credentials scoped to the bucket, which is awkward for machine-to-machine integrations and gives no job-level visibility.
2. Stand up their own API layer in front — duplicates the headless transform's work and fragments operations.

This PR ships a first-party REST API in the headless template itself: OAuth2 bearer-token auth via Cognito User Pool (API-only, separate from the normal web UI pool), presigned S3 uploads, and a job-tracker service that lets callers poll processing status and pull results without ever touching S3 directly.

### 1. New Jobs REST API (`src/lambda/api_handler/`, `src/lambda/job_tracker/`, `src/lambda/batch_pre_processor/`)

Three new Lambdas, all behind a **Private API Gateway** (REST, VPC endpoint resource policy):

- **`api_handler`** — AWS Lambda Powertools REST API implementing:
  - `POST /jobs` → creates a job row, returns `jobId` and a presigned `POST` for the input zip
  - `GET /jobs/{jobId}` → returns status (`PENDING_UPLOAD` / `IN_PROGRESS` / `SUCCEEDED` / `FAILED`) and, on success, a presigned `GET` URL for `results.zip`
  - Full Pydantic models for requests/responses (`models.py`), 357-line test suite (`test_index.py`)
- **`job_tracker`** — watches CloudWatch Events from Step Functions and updates the DynamoDB job row. Gates `SUCCEEDED` on the `results.zip` object actually existing in the OutputBucket (avoids a race where a caller sees SUCCEEDED, tries to download, gets 404).
- **`batch_pre_processor`** — unzips uploaded batch inputs onto `InputBucket` so the normal IDP pipeline kicks off exactly as it does for a direct S3 drop. Keeps the API layer from leaking into classification/extraction code.

### 2. VPC-secured deployment option (`template.yaml`, `patterns/unified/template.yaml`)

New top-level CFN parameters:

| Parameter | Purpose |
|---|---|
| `EnableHeadless` | Turns on the Jobs REST API (new) |
| `DeployInVPC` | Puts all IDP Lambdas in customer-supplied private subnets with a customer-supplied security group |
| `VpcId`, `PrivateSubnetIds`, `ApiGatewayVpcEndpointId`, `LambdaSecurityGroupId` | Customer-supplied networking |
| `ApiStageName` | API Gateway stage name (default `beta`) |
| `DeployBastionHost` | Optional: spins up a Session-Manager-reachable EC2 for dev-time tunnel access |
| `BastionHostSubnetId`, `BastionHostSecurityGroupId` | Customer-supplied bastion networking |

All gated by CFN `!Condition` logic so non-VPC / non-headless deployments are unchanged. Template diff is ~700 lines in `template.yaml`, ~160 in `patterns/unified/template.yaml`.

### 3. New job-service layer in `idp_common_pkg`

- `lib/idp_common_pkg/idp_common/dynamodb/job_service.py` — CRUD for job rows in the tracking table (120 lines, new module)
- `lib/idp_common_pkg/idp_common/job_service.py` — thin service facade (31 lines, new)
- Additions to `dynamodb/client.py`, `dynamodb/service.py`, and `models.py` to support the new job shape
- README updates

### 4. CLI polish for headless deployments (`lib/idp_cli_pkg/idp_cli/cli.py`)

- `--headless` now auto-sets `EnableHeadless=true` as a stack parameter. Before, users had to pass both `--headless` (which strips UI/AppSync/Cognito from the template) and `EnableHeadless=true` in `--parameters` (which turns on the Jobs API). They're always used together — this just de-duplicates.
- Fresh-stack creation with `--headless` no longer requires `--admin-email`. The headless template drops the Cognito UI pool and has no `AdminEmail` parameter, so the CLI previously failed with `ValidationError: Parameters: [AdminEmail] do not exist in the template`. We now skip the check and drop the value before it reaches CFN. Non-headless new-stack behavior is unchanged (still required).

### 5. Correctness fixes in the publish pipeline (`lib/idp_sdk/idp_sdk/_core/publish.py`)

Three fixes needed to make `idp-cli deploy --from-code . --headless` succeed against GovCloud:

- **cfn-lint skips UI/AppSync templates in headless mode.** Previously `_validate_cfn_lint` always linted `.aws-sam/idp-main.yaml` and every nested template, including `nested/appsync/template.yaml`. Those contain `AWS::AppSync::*`, `AWS::CloudFront::*`, etc., which don't exist in `us-gov-*` regions — so the lint step failed with 100+ `E3006 Resource type … does not exist` even though those resources are stripped later by `HeadlessTemplateTransformer`. Now, in headless mode, we lint `idp-headless.yaml` if present and skip the commercial-only templates.
- **Correct E/W classification.** The regex that counted cfn-lint errors used `or ":E" in line`, which also matched resource prefixes like `AWS::EC2::`, inflating warning-severity lines to errors and failing otherwise-green builds. Fixed with `^E\d{4}` / `^W\d{4}` anchored at line start.
- **JobTracker target on `WorkflowStateChangeRule`** moved from a conditional `Arn` field (flagged `E3003 'Arn' is a required property` when `DeployApiGateway=false`) to a conditional full-target dict via `!If`, so it's either complete or `AWS::NoValue`.

### 6. Tooling

- `scripts/bastion.sh` — sets up a local SSH tunnel to the bastion host (when `DeployBastionHost=true`) via SSM Session Manager + EC2 Instance Connect, maps the API Gateway private DNS onto loopback for local development. Supports a `<STACK_NAME>` argument so you can have multiple deployments.
- `scripts/get_api_token.sh` — OAuth2 `client_credentials` flow helper against the API Cognito pool, prints a bearer token.
- `scripts/create_vpc_endpoints.sh` — convenience script for standing up the required VPC endpoints (S3, DynamoDB Gateway, and the API Gateway execute-api Interface endpoint) in the customer's VPC.
- `scripts/e2e_test_headless.py` — full end-to-end Python test: OAuth → POST /jobs → presigned upload → status poll → download results. Takes `<STACK_NAME>` and `<PATH_TO_FILE>` as positional args so it's reusable across deployments.

### 7. Documentation

- `docs/govcloud-deployment.md` — overhauled into a "Deployment Packages" matrix (Vanilla / Headless API / Headless + VPC-secured / Headless + VPC + Bastion) with explicit parameter references for each
- `docs/govcloud-architecture.md` **(new)** — walks through which services are retained vs. removed in GovCloud and why
- `docs/govcloud-batch-api.md` **(new)** — full REST API reference with request/response schemas and an OAuth walkthrough
- `docs/govcloud-operations.md` **(new)** — monitoring, CloudWatch dashboards, common issues
- `docs/vpc-secured-mode.md` **(new)** — VPC prerequisites, subnet/endpoint/SG requirements

## Testing

Full end-to-end testing against a live GovCloud (`us-gov-west-1`) account:

- **Deploy #1 — update existing stack.** `idp-cli deploy --stack-name my-idp-govcloud-stack --from-code . --headless --wait --parameters "DeployInVPC=true,..."` — completed in ~3 minutes. UPDATE_COMPLETE.
- **E2E #1.** `scripts/e2e_test_headless.py my-idp-govcloud-stack samples/irs_documents.zip` — `jobId=25ec2930-b5fd-4932-8ede-eb3a803a1b2f` → `SUCCEEDED`, 7.47 MB results downloaded. ~3 min wall-clock.
- **Deploy #2 — fresh stack creation.** `idp-cli deploy --stack-name idp-govcloud-fresh-test --from-code . --headless --wait` (same VPC/subnet/SG parameters) — CREATE_COMPLETE in ~8 min.
- **E2E #2.** `scripts/e2e_test_headless.py idp-govcloud-fresh-test samples/irs_documents.zip` — `jobId=29a1272d-d1c4-435f-9bb1-34b452b7c582` → `SUCCEEDED`, 7.47 MB results. ~2:36 wall-clock.

Unit test coverage: `src/lambda/api_handler/test_index.py` (357 lines), `src/lambda/batch_pre_processor/test_index.py` (184 lines), `src/lambda/job_tracker/test_index.py` (203 lines).

## Backward compatibility

All new behavior is feature-flagged on CFN parameters that default to off:

- `EnableHeadless` defaults to `false` → no Jobs API, no new Lambdas deployed
- `DeployInVPC` defaults to `false` → no VPC config applied, Lambdas deploy as they did before
- `DeployBastionHost` defaults to `false` → no EC2 instance deployed

Existing stacks update in place with no new resources unless they opt in. The CLI fixes in (4) and (5) don't affect non-`--headless` invocations.

## Reviewer notes

- `scripts/bastion.sh` uses `sudo` for `/etc/hosts` modifications and loopback aliasing on macOS. This is a local-dev convenience tool and it should probably be annotated with the sudo requirement in the docs — flag if you'd like that cleaned up before merge. Known rough edge: when the corporate "Amazon Endpoint Client" has already written to `/etc/hosts` without a trailing newline, the bastion script's `>> /etc/hosts` can collide with the prior line; worth adding a newline-guard. Not a blocker; happy to address in this PR or a follow-up.
- The PR includes an `idp-cli deploy --headless --from-code .` path that exercises the headless template transformer end-to-end. If the upstream team would rather keep that CLI change separate from the feature work, I can split the CLI/publish fixes (commits `f4f504df`, `0117a5a0`, `a727684a`, `46431200`, `f7470c25`) into a standalone PR that lands first.
- This work was previously rebased onto a clean history branch (`update/main-0.58-clean`) after an earlier branch lost history during a bad merge. Every commit on this branch has a clean linear lineage back through the `origin/main` merge commit.

## Commit history (36 commits)

By @nicklynberg and @gardezia, Jan 21 → Apr 28. Grouped by theme:

- **Initial API layer infra (Jan 21–26)** — `a13428d7`, `e61cf9e4`, `404133d4`, `22e6214b` (and resyncs `2795e3b8`, `f02d7d3f`, `9415df94`, `e1622d06`)
- **Bastion host + VPC-secured mode (Feb 3–9)** — `eab6d0fc`, `5bfb80d7`, `6bbd0911`, `f9700c5c`, `79590868` … plus cleanups
- **Merges of feature branches (Feb 9, Feb 20)** — `f8060efe` (#5), `f2e3bc31` (#3), `f2f70781` (#8)
- **Rebuild onto clean history + `origin/main` merge (Apr 28)** — `bee44a91`
- **E2E test script + publish/CLI fixes for headless GovCloud deploy (Apr 27–28)** — `aea82775`, `f4f504df`, `0117a5a0`, `a727684a`, `f7470c25`, `46431200`


rstrahan:
Thanks for this — it's a sizable, clearly-motivated feature, and the PR description, commit hygiene (for the recent fix commits), and the live GovCloud e2e runs make the review much easier. The bones are solid. A few things to address before merge, summarized below.

## Overall verdict

__Request changes__ — mostly to tighten the auth model on `GET /jobs/{id}`, enforce the `EnableHeadless` ↔ `DeployInVPC` coupling at the CFN-parameter layer, and add a bound on uncompressed zip content. Details inline.

### What's good

- __Feature-flagged__. With defaults (`EnableHeadless=false`, `DeployInVPC=false`, `DeployBastionHost=false`), a diff of the CloudFormation changeset against `develop` is effectively empty — every new resource is gated, and the `VpcConfig: !If [DeployInVPC, …, AWS::NoValue]` additions on existing Lambdas resolve to `AWS::NoValue`. Existing stacks update in place with no new resources unless they opt in.
- __Private API Gateway resource policy__ (`Deny ≠ SourceVpce / Allow *`) is the right pattern and correctly scoped to the provided VPCE.
- __Presigned URL scoping__ is defensive: 1-hour expiry (configurable), content-type pinned to `application/zip`, `content-length-range` bound, keyed under `jobs/{uuid}/`.
- __`SUCCEEDED`/`ResultsReady` gating fix__ (commit `f7470c25`) is a real race fix that clearly came from running this in anger. 👍
- __Bastion hardening__ looks correct: `HttpTokens: required` (IMDSv2), EBS encrypted with a dedicated KMS key with rotation, SSM-only access, no inbound SSH — good defaults.
- __CLI and publish fixes__ (`--headless` implies `EnableHeadless=true`; new-stack `--headless` doesn't require `--admin-email`; cfn-lint `^E\d{4}`/`^W\d{4}` regex anchoring; conditional full-target on `WorkflowStateChangeRule`) are genuine correctness improvements and don't regress non-headless users.
- __Docs__ are above average for this repo: 4 new pages plus the GovCloud guide overhaul, with a deployment-variant matrix, API reference, and ops guide. The `vpc-secured-mode.md` prerequisites list is particularly clear.
- __Unit test coverage__ for the 3 new Lambdas is meaningful (357 / 184 / 203 lines). Assertions are real behavior checks, not "it doesn't crash."
- __Real e2e validation__ in `us-gov-west-1` against both a fresh CREATE and an UPDATE path.

## Blockers

### 🔴 H1 — `EnableHeadless=true` + `DeployInVPC=false` produces a broken stack

`ApiGateway.EndpointConfiguration.Type: PRIVATE` references `!Ref ApiGatewayVpcEndpointId`, and `ApiHandlerFunction.VpcConfig` hardcodes `!Ref PrivateSubnetIds` / `!Ref LambdaSecurityGroupId`. Those four parameters all default to `""`. A user who flips on headless without VPC mode either fails the deploy or creates an unreachable Private API GW. The CFN parameter layout implies these are independent knobs, but they aren't.

Preferred fix: add a CFN `Rules:` block so CloudFormation rejects the combo up front with a clear error, rather than failing deep in the deploy:

```yaml
Rules:
  HeadlessRequiresVPC:
    RuleCondition: !Equals [!Ref EnableHeadless, 'true']
    Assertions:
      - Assert: !Equals [!Ref DeployInVPC, 'true']
        AssertDescription: >-
          EnableHeadless=true requires DeployInVPC=true
          (the Jobs API uses a Private API Gateway).
      - Assert: !Not [!Equals [!Ref VpcId, '']]
        AssertDescription: VpcId is required when EnableHeadless=true
      - Assert: !Not [!Equals [!Join ['', !Ref PrivateSubnetIds], '']]
        AssertDescription: PrivateSubnetIds is required when EnableHeadless=true
      - Assert: !Not [!Equals [!Ref ApiGatewayVpcEndpointId, '']]
        AssertDescription: ApiGatewayVpcEndpointId is required when EnableHeadless=true
      - Assert: !Not [!Equals [!Ref LambdaSecurityGroupId, '']]
        AssertDescription: LambdaSecurityGroupId is required when EnableHeadless=true
```

### 🔴 M1 — No per-caller authorization on `GET /jobs/{id}`

`api_handler.get_job` resolves by `job_id` only. Any caller with any valid Cognito client-credentials token on the API user pool can read any job's file list and obtain a presigned download URL for its `results.zip`. UUIDv4 unguessability is the only defense in depth — that's acceptable for a strictly single-tenant deployment but surprising for a feature that ships a first-class "issue API clients" story.

Minimum fix:

1. On `POST /jobs`, record the token's `sub` / `client_id` (available in `event.requestContext.authorizer.claims`) onto the job record as `CreatedBy`.
2. On `GET /jobs/{id}`, compare the caller's principal to `CreatedBy` and return 404 (not 403, to avoid existence-leak) on mismatch.

If "all API clients share job visibility" is the intended contract, __please document it explicitly__ in `docs/govcloud-batch-api.md` under a "Multi-tenancy" heading, and note that downstream customers must not issue separate API clients to distinct tenants.

### 🔴 M2 — Zip-bomb protection missing in `batch_pre_processor`

Zip-slip is correctly mitigated via `os.path.basename(...)` before the S3 key is assembled — good. However:

- `MAX_FILE_SIZE_BYTES=5GB` is enforced on the *compressed* zip via the presigned POST, not on the inflated content.
- `zip_ref.read(file_info.filename)` loads each entry fully into memory.
- There is no cap on total uncompressed size, per-entry uncompressed size, or entry count.

A 5 GB zip of zeros inflates many orders of magnitude; even without a zip bomb, a large legitimate zip can OOM the Lambda. For an authenticated API the blast radius is "a caller can DoS their own processing pipeline," but it's still worth bounding.

Suggested fix:

```python
MAX_UNCOMPRESSED_BYTES = int(os.environ.get("MAX_UNCOMPRESSED_BYTES", 20 * 1024**3))  # 20GB
MAX_ENTRIES = int(os.environ.get("MAX_ENTRIES", 10_000))

total = 0
with zipfile.ZipFile(temp_file, "r") as zip_ref:
    entries = [i for i in zip_ref.infolist() if not i.is_dir()]
    if len(entries) > MAX_ENTRIES:
        raise ValueError(f"Zip has too many entries ({len(entries)} > {MAX_ENTRIES})")
    for file_info in entries:
        total += file_info.file_size
        if total > MAX_UNCOMPRESSED_BYTES:
            raise ValueError(f"Zip uncompressed size exceeds {MAX_UNCOMPRESSED_BYTES} bytes")
        # … stream with zip_ref.open(file_info) + s3.upload_fileobj() rather than read()
```

### 🔴 Missing CHANGELOG entry

The repo convention is to log user-facing changes in `CHANGELOG.md`. This is a substantial feature flag surface — please add an entry under the next unreleased version describing `EnableHeadless`, `DeployInVPC`, `DeployBastionHost`, the new Jobs REST API, and the `--headless` CLI behavior changes.

## Medium (preferably addressed, not strict blockers)

### M3 — Partial-failure handling in `batch_pre_processor`

The loop `raise`s on any single `put_object` exception, leaving some files in the InputBucket but the job's `Files` map unset. JobTracker then never sees them and the job hangs. Consider per-entry try/except that records failed filenames and still updates the job record (with those files marked `FAILED`), or make the whole handler idempotent and let Lambda retries finish the work.

### M4 — Cognito domain-prefix uniqueness

The API user pool domain is created with a prefix derived from the stack name. Cognito domain prefixes are __globally unique per region__ across all AWS accounts, not just the account — if the prefix isn't entropy-suffixed, a second deploy of the same stack name (or a prefix collision with another AWS customer) will hard-fail at `CREATE_IN_PROGRESS`. Please confirm the prefix includes stack-id entropy, e.g.:

```yaml
Domain: !Sub "${AWS::StackName}-${AWS::AccountId}-${AWS::Region}"
# or better:
Domain: !Select [2, !Split ["/", !Ref AWS::StackId]]
```

## Low / polish (happy to land in a follow-up)

- __L1.__ `--headless --parameters EnableHeadless=false` is silently accepted. It produces a deploy with neither UI nor Jobs API. Print a warning and require `--force` or similar.
- __L2.__ The cfn-lint headless skip list (`{"appsync"}`) is a hard-coded allowlist — maintenance item for future nested templates that are also stripped by `HeadlessTemplateTransformer`. A comment referencing the transformer's own skip list would help future-me.
- __L3.__ `scripts/bastion.sh` sudo + `/etc/hosts` newline-guard: author already flagged this in the PR description — happy to accept as a follow-up. The docs should make the sudo requirement prominent before the script is recommended.
- __L4.__ The 36-commit history is messy (two feature-branch resyncs plus a "recovery merge"). Author was candid about this and it doesn't affect the diff, but we should __squash-merge__ this PR so the commit log stays clean on `develop`.
- __L5.__ `compute_job_status` returns `"ABORTED"` when all files are `Status.ABORTED`, but the `Rules:`-block I suggested above doesn't add any validation here — worth a note: the public API reference in `docs/govcloud-batch-api.md` lists `PENDING_UPLOAD / IN_PROGRESS / SUCCEEDED / PARTIALLY_SUCCEEDED / FAILED / ABORTED`; please confirm all six are documented consistently.
- __L6.__ `create_job_record` uses `ConditionExpression` to avoid overwriting an existing `job_id` collision? Not visible from the service facade — UUIDv4 collisions aren't a realistic risk but a conditional put is cheap insurance.

## Required-before-merge checklist

- H1 — add CFN `Rules:` block coupling `EnableHeadless` → `DeployInVPC` and non-empty networking params
- M1 — scope `GET /jobs/{id}` to the creating principal, __or__ document single-tenant explicitly in `govcloud-batch-api.md`
- M2 — add uncompressed-size / entry-count bound in `batch_pre_processor`, switch to streaming
- Add `CHANGELOG.md` entry for the new feature flags and `--headless` CLI behavior
- Confirm Cognito domain prefix uniqueness (M4)
- Plan to __squash-merge__ on land

Nice-to-have (not blocking): M3, L1–L6, and the bastion newline guard already offered as a follow-up.

Thanks again — this is a meaningful unlock for GovCloud users, and the test coverage + real e2e runs show it. Happy to re-review as soon as the blockers are addressed.

alig4lfe:
> Thanks for this — it's a sizable, clearly-motivated feature, and the PR description, commit hygiene (for the recent fix commits), and the live GovCloud e2e runs make the review much easier. The bones are solid. A few things to address before merge, summarized below.
> 
> ## Overall verdict
> **Request changes** — mostly to tighten the auth model on `GET /jobs/{id}`, enforce the `EnableHeadless` ↔ `DeployInVPC` coupling at the CFN-parameter layer, and add a bound on uncompressed zip content. Details inline.
> 
> ### What's good
> * **Feature-flagged**. With defaults (`EnableHeadless=false`, `DeployInVPC=false`, `DeployBastionHost=false`), a diff of the CloudFormation changeset against `develop` is effectively empty — every new resource is gated, and the `VpcConfig: !If [DeployInVPC, …, AWS::NoValue]` additions on existing Lambdas resolve to `AWS::NoValue`. Existing stacks update in place with no new resources unless they opt in.
> * **Private API Gateway resource policy** (`Deny ≠ SourceVpce / Allow *`) is the right pattern and correctly scoped to the provided VPCE.
> * **Presigned URL scoping** is defensive: 1-hour expiry (configurable), content-type pinned to `application/zip`, `content-length-range` bound, keyed under `jobs/{uuid}/`.
> * **`SUCCEEDED`/`ResultsReady` gating fix** (commit `f7470c25`) is a real race fix that clearly came from running this in anger. 👍
> * **Bastion hardening** looks correct: `HttpTokens: required` (IMDSv2), EBS encrypted with a dedicated KMS key with rotation, SSM-only access, no inbound SSH — good defaults.
> * **CLI and publish fixes** (`--headless` implies `EnableHeadless=true`; new-stack `--headless` doesn't require `--admin-email`; cfn-lint `^E\d{4}`/`^W\d{4}` regex anchoring; conditional full-target on `WorkflowStateChangeRule`) are genuine correctness improvements and don't regress non-headless users.
> * **Docs** are above average for this repo: 4 new pages plus the GovCloud guide overhaul, with a deployment-variant matrix, API reference, and ops guide. The `vpc-secured-mode.md` prerequisites list is particularly clear.
> * **Unit test coverage** for the 3 new Lambdas is meaningful (357 / 184 / 203 lines). Assertions are real behavior checks, not "it doesn't crash."
> * **Real e2e validation** in `us-gov-west-1` against both a fresh CREATE and an UPDATE path.
> 
> ## Blockers
> ### 🔴 H1 — `EnableHeadless=true` + `DeployInVPC=false` produces a broken stack
> `ApiGateway.EndpointConfiguration.Type: PRIVATE` references `!Ref ApiGatewayVpcEndpointId`, and `ApiHandlerFunction.VpcConfig` hardcodes `!Ref PrivateSubnetIds` / `!Ref LambdaSecurityGroupId`. Those four parameters all default to `""`. A user who flips on headless without VPC mode either fails the deploy or creates an unreachable Private API GW. The CFN parameter layout implies these are independent knobs, but they aren't.
> 
> Preferred fix: add a CFN `Rules:` block so CloudFormation rejects the combo up front with a clear error, rather than failing deep in the deploy:
> 
> ```yaml
> Rules:
>   HeadlessRequiresVPC:
>     RuleCondition: !Equals [!Ref EnableHeadless, 'true']
>     Assertions:
>       - Assert: !Equals [!Ref DeployInVPC, 'true']
>         AssertDescription: >-
>           EnableHeadless=true requires DeployInVPC=true
>           (the Jobs API uses a Private API Gateway).
>       - Assert: !Not [!Equals [!Ref VpcId, '']]
>         AssertDescription: VpcId is required when EnableHeadless=true
>       - Assert: !Not [!Equals [!Join ['', !Ref PrivateSubnetIds], '']]
>         AssertDescription: PrivateSubnetIds is required when EnableHeadless=true
>       - Assert: !Not [!Equals [!Ref ApiGatewayVpcEndpointId, '']]
>         AssertDescription: ApiGatewayVpcEndpointId is required when EnableHeadless=true
>       - Assert: !Not [!Equals [!Ref LambdaSecurityGroupId, '']]
>         AssertDescription: LambdaSecurityGroupId is required when EnableHeadless=true
> ```
> 
> ### 🔴 M1 — No per-caller authorization on `GET /jobs/{id}`
> `api_handler.get_job` resolves by `job_id` only. Any caller with any valid Cognito client-credentials token on the API user pool can read any job's file list and obtain a presigned download URL for its `results.zip`. UUIDv4 unguessability is the only defense in depth — that's acceptable for a strictly single-tenant deployment but surprising for a feature that ships a first-class "issue API clients" story.
> 
> Minimum fix:
> 
> 1. On `POST /jobs`, record the token's `sub` / `client_id` (available in `event.requestContext.authorizer.claims`) onto the job record as `CreatedBy`.
> 2. On `GET /jobs/{id}`, compare the caller's principal to `CreatedBy` and return 404 (not 403, to avoid existence-leak) on mismatch.
> 
> If "all API clients share job visibility" is the intended contract, **please document it explicitly** in `docs/govcloud-batch-api.md` under a "Multi-tenancy" heading, and note that downstream customers must not issue separate API clients to distinct tenants.
> 
> ### 🔴 M2 — Zip-bomb protection missing in `batch_pre_processor`
> Zip-slip is correctly mitigated via `os.path.basename(...)` before the S3 key is assembled — good. However:
> 
> * `MAX_FILE_SIZE_BYTES=5GB` is enforced on the _compressed_ zip via the presigned POST, not on the inflated content.
> * `zip_ref.read(file_info.filename)` loads each entry fully into memory.
> * There is no cap on total uncompressed size, per-entry uncompressed size, or entry count.
> 
> A 5 GB zip of zeros inflates many orders of magnitude; even without a zip bomb, a large legitimate zip can OOM the Lambda. For an authenticated API the blast radius is "a caller can DoS their own processing pipeline," but it's still worth bounding.
> 
> Suggested fix:
> 
> ```python
> MAX_UNCOMPRESSED_BYTES = int(os.environ.get("MAX_UNCOMPRESSED_BYTES", 20 * 1024**3))  # 20GB
> MAX_ENTRIES = int(os.environ.get("MAX_ENTRIES", 10_000))
> 
> total = 0
> with zipfile.ZipFile(temp_file, "r") as zip_ref:
>     entries = [i for i in zip_ref.infolist() if not i.is_dir()]
>     if len(entries) > MAX_ENTRIES:
>         raise ValueError(f"Zip has too many entries ({len(entries)} > {MAX_ENTRIES})")
>     for file_info in entries:
>         total += file_info.file_size
>         if total > MAX_UNCOMPRESSED_BYTES:
>             raise ValueError(f"Zip uncompressed size exceeds {MAX_UNCOMPRESSED_BYTES} bytes")
>         # … stream with zip_ref.open(file_info) + s3.upload_fileobj() rather than read()
> ```
> 
> ### 🔴 Missing CHANGELOG entry
> The repo convention is to log user-facing changes in `CHANGELOG.md`. This is a substantial feature flag surface — please add an entry under the next unreleased version describing `EnableHeadless`, `DeployInVPC`, `DeployBastionHost`, the new Jobs REST API, and the `--headless` CLI behavior changes.
> 
> ## Medium (preferably addressed, not strict blockers)
> ### M3 — Partial-failure handling in `batch_pre_processor`
> The loop `raise`s on any single `put_object` exception, leaving some files in the InputBucket but the job's `Files` map unset. JobTracker then never sees them and the job hangs. Consider per-entry try/except that records failed filenames and still updates the job record (with those files marked `FAILED`), or make the whole handler idempotent and let Lambda retries finish the work.
> 
> ### M4 — Cognito domain-prefix uniqueness
> The API user pool domain is created with a prefix derived from the stack name. Cognito domain prefixes are **globally unique per region** across all AWS accounts, not just the account — if the prefix isn't entropy-suffixed, a second deploy of the same stack name (or a prefix collision with another AWS customer) will hard-fail at `CREATE_IN_PROGRESS`. Please confirm the prefix includes stack-id entropy, e.g.:
> 
> ```yaml
> Domain: !Sub "${AWS::StackName}-${AWS::AccountId}-${AWS::Region}"
> # or better:
> Domain: !Select [2, !Split ["/", !Ref AWS::StackId]]
> ```
> 
> ## Low / polish (happy to land in a follow-up)
> * **L1.** `--headless --parameters EnableHeadless=false` is silently accepted. It produces a deploy with neither UI nor Jobs API. Print a warning and require `--force` or similar.
> * **L2.** The cfn-lint headless skip list (`{"appsync"}`) is a hard-coded allowlist — maintenance item for future nested templates that are also stripped by `HeadlessTemplateTransformer`. A comment referencing the transformer's own skip list would help future-me.
> * **L3.** `scripts/bastion.sh` sudo + `/etc/hosts` newline-guard: author already flagged this in the PR description — happy to accept as a follow-up. The docs should make the sudo requirement prominent before the script is recommended.
> * **L4.** The 36-commit history is messy (two feature-branch resyncs plus a "recovery merge"). Author was candid about this and it doesn't affect the diff, but we should **squash-merge** this PR so the commit log stays clean on `develop`.
> * **L5.** `compute_job_status` returns `"ABORTED"` when all files are `Status.ABORTED`, but the `Rules:`-block I suggested above doesn't add any validation here — worth a note: the public API reference in `docs/govcloud-batch-api.md` lists `PENDING_UPLOAD / IN_PROGRESS / SUCCEEDED / PARTIALLY_SUCCEEDED / FAILED / ABORTED`; please confirm all six are documented consistently.
> * **L6.** `create_job_record` uses `ConditionExpression` to avoid overwriting an existing `job_id` collision? Not visible from the service facade — UUIDv4 collisions aren't a realistic risk but a conditional put is cheap insurance.
> 
> ## Required-before-merge checklist
> * H1 — add CFN `Rules:` block coupling `EnableHeadless` → `DeployInVPC` and non-empty networking params
> * M1 — scope `GET /jobs/{id}` to the creating principal, **or** document single-tenant explicitly in `govcloud-batch-api.md`
> * M2 — add uncompressed-size / entry-count bound in `batch_pre_processor`, switch to streaming
> * Add `CHANGELOG.md` entry for the new feature flags and `--headless` CLI behavior
> * Confirm Cognito domain prefix uniqueness (M4)
> * Plan to **squash-merge** on land
> 
> Nice-to-have (not blocking): M3, L1–L6, and the bastion newline guard already offered as a follow-up.
> 
> Thanks again — this is a meaningful unlock for GovCloud users, and the test coverage + real e2e runs show it. Happy to re-review as soon as the blockers are addressed.

# Response to review — status + game plan

Thanks for the thorough read. I've validated each blocker against the actual code, started landing fixes, and split them into **done**, **in progress**, and **refuted**. Concrete status below.

## TL;DR

| # | Blocker | Status |
|---|---|---|
| H1 | Headless ↔ VPC coupling | ✅ **Done** — CFN `Rules:` block + defense-in-depth `!If [DeployInVPC, …]` Will raise updated PR shortly |
| M1 | Per-caller auth on `GET /jobs/{id}` | ✅ **Done** — `CreatedBy` persisted + 404-on-mismatch + 6 new tests + docs Will raise updated PR shortly |
| M2 | Zip-bomb / uncompressed size bounds | ⏳ **In progress** — bounds + streaming + per-entry isolation |
| CHANGELOG | Missing entry | ⏳ **Queued** — will be the last thing added before re-request |
| M4 | Cognito domain prefix | 🟢 **Refuted with evidence** — already handled via `${AWS::AccountId}` |

All work is on the existing PR branch. No force-pushes yet; I'll push after the full set is green and re-request review.

## ✅ H1 — done

Confirmed your read. `ApiHandlerFunction` (was template.yaml line 1033) and `BatchPreProcessorFunction` (was line 1113) both had **unconditional** `VpcConfig` blocks `!Ref`-ing `LambdaSecurityGroupId` / `PrivateSubnetIds`, which default to `""`. `EnableHeadless=true` + `DeployInVPC=false` was a hard deploy failure, not just "unreachable API." `ApiGateway.EndpointConfiguration` had the same pattern on `ApiGatewayVpcEndpointId`.

**Fix applied:**

1. New `HeadlessRequiresVPC` rule added to the existing `Rules:` block in `template.yaml` (alongside the existing `MLflowTrackingURIRequired`) — fails the changeset at creation time with explicit `AssertDescription`s rather than mid-deploy:
   - `EnableHeadless=true` ⇒ `DeployInVPC=true`
   - `EnableHeadless=true` ⇒ non-empty `VpcId`, `PrivateSubnetIds`, `ApiGatewayVpcEndpointId`, `LambdaSecurityGroupId`
2. New `BastionRequiresVPC` rule (same pattern) for `DeployBastionHost=true` ⇒ `DeployInVPC=true` + non-empty bastion subnet/SG.
3. Defense in depth: the two API Lambda `VpcConfig` blocks are now wrapped in `!If [DeployInVPC, …, !Ref AWS::NoValue]`, matching the pattern already used on non-API Lambdas. The template is self-consistent even if the rule is ever relaxed.

Deliberately did **not** make `DeployInVPC` imply `EnableHeadless` — some users want VPC-only core processing without the Jobs API. Only the reverse coupling you called out.

Template re-parses clean as valid YAML after the changes.

## ✅ M1 — done

Confirmed. `get_job(job_id)` in `src/lambda/api_handler/index.py:133` did zero principal check. Went with your recommended approach (`CreatedBy` + 404-on-mismatch); I considered refuting this as "we only issue one M2M client so client == deployment boundary," but landed on implementing it because it's ~30 lines, gives us audit-trail utility even in the single-client case, and doesn't require customers to know about the pitfall if they ever add a second client for CI vs prod.

**Fix applied:**

1. **`lib/idp_common_pkg/idp_common/dynamodb/job_service.py`** — `create_job_record` gained an optional `created_by: Optional[str]` parameter that persists `CreatedBy` on the job record. Non-breaking: existing callers that don't pass it write records without the field.

2. **`src/lambda/api_handler/index.py`** — new `_caller_principal()` helper that extracts `sub` (falling back to `client_id`) from `app.current_event.raw_event["requestContext"]["authorizer"]["claims"]`, defensively. Implementation reads the raw event dict rather than Powertools' `APIGatewayEventRequestContext` wrapper because the wrapper raises `KeyError` on missing `requestContext` (matters for non-API-Gateway invocation paths and unit tests).

3. **`POST /jobs`** — calls `_caller_principal()` and passes the result as `created_by` to `create_job_record`. Emits a structured warning log if the principal can't be resolved on an authorized request (shouldn't happen in practice since API Gateway rejects unauthorized requests before Lambda is invoked, but surfaces operator misconfigurations).

4. **`GET /jobs/{job_id}`** — new check right after the job-record fetch: compares `job_record["CreatedBy"]` to the caller's principal. On mismatch, raises `aws_lambda_powertools.event_handler.exceptions.NotFoundError` → **HTTP 404** (not 403 — avoids existence-leak, per your recommendation). Jobs without `CreatedBy` (legacy records written before this change) remain readable by any authenticated caller, so in-place upgrades don't orphan in-flight jobs.

5. **`src/lambda/api_handler/test_index.py`** — 6 new tests:
   - `test_create_job_records_created_by_from_authorizer_claims` — POST persists the principal
   - `test_create_job_without_authorizer_passes_none_created_by` — no principal → None (doesn't crash)
   - `test_get_job_same_principal_visible` — creator can read their own job
   - `test_get_job_different_principal_returns_404` — cross-client read returns 404, not 403
   - `test_get_job_missing_principal_returns_404` — unresolvable caller can't read owned jobs
   - `test_get_job_legacy_record_without_created_by_is_readable` — legacy records remain accessible
   - Plus one bonus test (`test_get_job_succeeded_but_results_not_ready_reports_in_progress`) that exercises the `ResultsReady` gate — that code path had no coverage before.

6. **Bug fixes surfaced by the new tests:** two pre-existing test failures I had to correct:
   - `test_get_job_succeeded` was missing `ResultsReady: True` in its mock — it had been failing since commit `f7470c25` landed (that's the ResultsReady gating fix we cherry-picked during the branch rebuild). Updated the mock.
   - `test_get_job_not_found` was asserting **HTTP 400** for a missing job. That was a pre-existing behavior bug (the original code's `raise ValueError(f"Job {job_id} not found")` got caught by the generic `except ValueError` branch and mapped to `BadRequestError`). My change routes missing jobs through `NotFoundError` → HTTP 404, which is the correct REST behavior and matches the per-client mismatch path. Updated the assertion accordingly. This is a small user-visible change (400 → 404 on missing jobs) worth mentioning in the CHANGELOG — will flag it there.

All **28 tests pass**. `ruff check` clean on modified files.

7. **`docs/govcloud-batch-api.md`** — new "**Authorization model**" section between Authentication and API Reference, covering:
   - Per-client ownership semantics
   - Single-client deployment behavior (unchanged from before)
   - Multi-client deployment behavior (isolation)
   - Credential rotation behavior (same-client = visibility preserved; replace-client = not)
   - Legacy records without `CreatedBy` remain readable
   - Operator-level cross-client access via S3 OutputBucket rather than API

## ⏳ M2 + M3 — in progress

Your diagnosis is correct: `zip_ref.read(file_info.filename)` loads entries whole into Lambda memory, no uncompressed-size cap, no entry-count cap, any single `put_object` failure aborts the whole batch leaving the job record unset.

**Plan:**

1. **Bounds** — `MAX_UNCOMPRESSED_BYTES` (default 20 GB, env-overridable) and `MAX_ENTRIES` (default 10,000) checked **before any uploads start**. On violation, write a terminal `FAILED` status to the job record with a human-readable reason before raising — callers see the failure via `GET /jobs/{id}` rather than a silent hang.
2. **Streaming** — replace `zip_ref.read()` + `s3.put_object(Body=...)` with `zip_ref.open(file_info)` + `s3.upload_fileobj(stream, ...)`. The stream is chunked; we never hold a full entry in memory.
3. **Per-entry isolation (M3)** — wrap each entry in `try/except`, collect failed filenames, continue processing survivors. Call `update_job_files` with a `Status.FAILED` for each failed filename alongside `Status.IN_PROGRESS` for the uploads that succeeded. JobTracker then advances those to terminal states normally, and the job converges to `PARTIALLY_SUCCEEDED` / `FAILED` / `SUCCEEDED` depending on outcome.
4. **Test** — oversized-zip test in `src/lambda/batch_pre_processor/test_index.py` that constructs a `zipfile.ZipInfo` with an inflated `file_size` and asserts the handler raises before `upload_fileobj` is called. Plus a per-entry-failure test that mocks `upload_fileobj` to raise on one entry and asserts the rest succeed and the job record carries `FAILED` for the bad entry.

**Deliberately not addressing in this PR:**

- Compressed-vs-uncompressed **ratio** check (classic zip-bomb heuristic). The reviewer didn't ask for it, and a hostile zip can set `file_info.file_size` to whatever it wants — the `MAX_UNCOMPRESSED_BYTES` cap enforced **before** streaming handles the same blast radius. If you want a ratio check too, happy to add one in a follow-up.

## ⏳ CHANGELOG — queued

Will add a single entry under `[Unreleased]` that includes:

- New feature flags: `EnableHeadless`, `DeployInVPC`, `DeployBastionHost`, plus the VPC/bastion networking params
- New Jobs REST API (`POST /jobs`, `GET /jobs/{id}`) with OAuth2 client-credentials auth
- Private API Gateway + optional bastion host deployment option
- Per-client job ownership (M1) — including the minor behavior change that `GET /jobs/{id}` on a non-existent job now returns 404 instead of 400 (was a pre-existing response-code bug)
- Zip upload bounds (M2) — `MAX_UNCOMPRESSED_BYTES` / `MAX_ENTRIES` environment knobs
- Partial-failure handling (M3) — job advances even if individual files fail to unpack
- CLI: `--headless` auto-sets `EnableHeadless=true`; `--headless` new-stack no longer requires `--admin-email`
- Publish pipeline: cfn-lint skips UI/AppSync templates in headless mode; cfn-lint E/W regex anchoring
- Jobs API: `SUCCEEDED` is gated on `results.zip` upload completing (`ResultsReady`)
- CFN validation: new `Rules:` block couples `EnableHeadless=true` ⇒ `DeployInVPC=true` + networking params; `DeployBastionHost=true` ⇒ `DeployInVPC=true` + bastion networking params

Doing this last so it's accurate to what actually ships.

## 🟢 M4 — respectfully refuting

`template.yaml:978`:

```yaml
ApiUserPoolDomain:
  Type: AWS::Cognito::UserPoolDomain
  Condition: DeployApiGateway
  Properties:
    Domain: !Sub ${AWS::StackName}-api-${AWS::AccountId}
    UserPoolId: !Ref ApiUserPool
```

The 12-digit `${AWS::AccountId}` is the entropy suffix. Cognito domain prefixes are unique **per region**, not globally across regions — two different AWS accounts necessarily land on different prefixes; the same customer in two different regions lands on different regional Cognito endpoints so the per-region uniqueness is preserved. Two stacks in the *same* account + *same* region with the *same* stack name will collide, but that's the CFN stack-name uniqueness rule itself failing first with a clear error (`Stack with id idp-govcloud-fresh-test already exists`), not a silent Cognito collision.

**Will add** a note in `docs/govcloud-batch-api.md` under the new Authorization model section documenting the `${StackName}-api-${AccountId}` naming so anyone scripting deploys knows to keep stack names distinct per region.

Not planning a code change here unless you feel strongly. If so, the smallest move would be appending `!Select [2, !Split ["/", !Ref AWS::StackId]]` (the CloudFormation stack GUID segment) to the domain for belt-and-suspenders entropy — happy to do that, but I think it's unnecessary.

## Medium & Low — what's landing in this PR vs follow-up

**Landing in this PR alongside the blockers:**

- **M3** (partial-failure in batch_pre_processor) — ships with M2 since the code path is the same loop.
- **L1** (`--headless --parameters EnableHeadless=false` silently accepted) — will emit a warning log; keeping `--parameters` precedence because that's the documented contract for that flag.
- **L2** (cfn-lint headless skip list hardcoded) — will add a comment pointing at `HeadlessTemplateTransformer`'s own skip list.
- **L5** (status enum docs consistency) — will verify all six states (`PENDING_UPLOAD` / `IN_PROGRESS` / `SUCCEEDED` / `PARTIALLY_SUCCEEDED` / `FAILED` / `ABORTED`) are listed consistently in `docs/govcloud-batch-api.md` and in `GetJobResponse` docstring.
- **L6** (conditional put on job create) — trivial `attribute_not_exists(PK)` guard. Will add.

**Deferring to follow-up with issues opened against this repo:**

- **L3** — `scripts/bastion.sh` sudo + `/etc/hosts` newline guard. I hit this personally during the us-gov-west-1 test runs (the corporate "Amazon Endpoint Client" had written to `/etc/hosts` without a trailing newline, and the bastion script's `>> /etc/hosts` concatenated into the comment marker). Separate PR — small, self-contained, and the bastion script is dev-only tooling.
- **L4** — squash-merge on land. Agreed — the commit history on this branch is messy because of an earlier history-loss recovery, not the feature shape. Squash-merge is the right call.

Thanks again for the care on this review — it's made the PR meaningfully better.

alig4lfe:
Addressed all four blockers in **`fd95df6a`** on this branch (just pushed). Single commit covering H1 + M1 + M2 + M3 + CHANGELOG.

**H1 — Headless ↔ VPC coupling.** Added `HeadlessRequiresVPC` and `BastionRequiresVPC` rules to the CFN `Rules:` block with your proposed `AssertDescription` messages, plus defense-in-depth `!If [DeployInVPC, …, AWS::NoValue]` wrap on the two API Lambda `VpcConfig` blocks. Verified live in `us-gov-west-1`: bad-parameter `create-stack` calls hit `ROLLBACK_COMPLETE` with every `AssertDescription` populated in StackEvents. One minor rework during implementation: the `Fn::Join`-based `PrivateSubnetIds` non-empty check isn't valid inside `Rules:` (the Rules-block intrinsic-function set has no portable way to assert a `CommaDelimitedList` isn't `[""]`), so that single assertion is dropped with a comment; the remaining `VpcId` / `VpcEndpoint` / `SG` assertions still catch the common "forgot to pass VPC params" misconfiguration.

**M1 — Per-client auth on GET /jobs/{id}.** `POST /jobs` now persists the caller's Cognito principal (`sub` / `client_id`) as `CreatedBy`; `GET /jobs/{id}` returns **HTTP 404** on mismatch (not 403, per your recommendation). Legacy records without `CreatedBy` remain readable by any caller so in-place upgrades don't orphan in-flight jobs. 6 new unit tests + 2 pre-existing broken tests fixed (28/28 green). Incidentally corrects a pre-existing bug where missing jobs returned 400; they now correctly return 404. Authorization model documented in `docs/govcloud-batch-api.md`.

**M2 + M3 — Zip bounds and partial-failure isolation.** New `MAX_UNCOMPRESSED_BYTES` (default 20 GiB) and `MAX_ENTRIES` (default 10k) env-configurable bounds checked pre-flight. Streaming via `zip_ref.open()` + `s3.upload_fileobj()` replaces the in-memory `read()`. Per-entry `try/except` isolates single-file failures — bad entries marked `FAILED` on the job record, surviving entries still advance through the pipeline. Bound violations write a terminal `FAILED` marker so the API surfaces the failure state. 13 tests, all green.

**CHANGELOG.** Populated the empty `[Unreleased]` section with the full feature surface — Jobs API shape, per-client ownership, zip bounds, CFN parameter additions, CLI behavior changes, and the publish-pipeline fixes that make `idp-cli deploy --from-code . --headless` succeed against GovCloud.

**M4 — refuting.** `template.yaml:978` already uses `Domain: !Sub ${AWS::StackName}-api-${AWS::AccountId}`. 12-digit AccountId is the entropy; Cognito domain prefixes are per-region unique so two accounts / two regions can't collide. Happy to add `!Select [2, !Split ["/", !Ref AWS::StackId]]` for more entropy if you'd prefer, but I don't think it's necessary.

**End-to-end verification** on `us-gov-west-1`: Deploy #3 = `UPDATE_COMPLETE` (~8.5 min). E2E #3 = `c51ca81c-…` → `SUCCEEDED`, 7.48 MB results. H1 fail-fast proven on throwaway stacks.

Ready for re-review.

PR #343, opened 2026-06-04
Feature/dep manifest for artifactory

sirirako:
# feat: add dependency manifest generation for artifact repository mirroring

## Summary

Adds on-demand dependency manifest generation for enterprise teams that need to mirror all Python and Node.js packages into an artifact repository (JFrog Artifactory, AWS CodeArtifact, Sonatype Nexus, etc.) for air-gapped builds and security scanning.

Run `make dep-manifest` locally or let CI generate manifests automatically when dependency files change.

## What's included

- **`scripts/generate-dep-manifest.sh`** — single script that:
  - Parses existing `uv.lock` files for resolved Python packages (217 packages)
  - Collects additional packages from `requirements.txt` not covered by lockfiles (9 packages)
  - Parses committed `package-lock.json` files with `jq` for Node packages (1,233 packages)
  - Warns on stderr when packages lack pinned versions
  - Outputs to gitignored `dist/manifests/`
- **`Makefile`** — `dep-manifest` target
- **`.github/workflows/generate-dep-manifest.yml`** — CI workflow triggered on dependency file changes or manually, uploads manifests as 90-day artifacts; runs as dry-run on PRs to catch parse errors
- **`docs/dependency-mirroring.md`** — setup instructions for artifact repository mirroring

## Output format

```
# dist/manifests/python-packages.txt
boto3==1.42.0
pydantic==2.12.0
...

# dist/manifests/node-packages.txt
@aws-amplify/ui-react@6.7.1
react@18.3.1
...
```

## Design decisions

- **Nothing committed** — manifests are generated on-demand under `dist/` (gitignored), always fresh from the current dependency state
- **No re-resolution** — leverages existing `uv.lock` and `package-lock.json` that are already vetted and committed
- **Vendor-neutral** — works with any artifact repository manager
- **Minimal prerequisites** — `uv`, `jq`, Python 3.12+ (all available in CI runners)
- **Proper semver** — uses `packaging.version.Version` for version comparison, not string ordering
- **Visible warnings** — unresolved packages and skipped lockfiles are reported to stderr

## Test plan

- [x] `make dep-manifest` runs successfully (~5 seconds)
- [x] `python-packages.txt` has 226 entries, all valid package names
- [x] `node-packages.txt` has 1,233 entries, all `name@version` format
- [x] Output is under `dist/` (gitignored, nothing committed)
- [x] Warning emitted for 9 packages not in any lockfile
- [ ] GitHub Actions workflow runs on push to main
- [ ] GitHub Actions workflow runs as dry-run on PRs (no artifact upload)


rstrahan:
Findings
🟡 Should fix — [.github/workflows/generate-dep-manifest.yml:38-46](vscode-webview://14rid8ak7asturkhjli19n9sf8f0jgraqq2l6mtjgkofaqaui5qp/.github/workflows/generate-dep-manifest.yml#L38-L46) — The workflow installs uv but the script never uses it, and Python/packaging are never provisioned. The script runs python3 directly (it does not call uv run), and the inline Python does from packaging.version import Version. The workflow has no setup-python step and no pip install packaging. On ubuntu-latest the system python3 does not reliably have packaging importable, so make dep-manifest can fail at runtime in CI. Either (a) run the python through uv (e.g. uv run --with packaging python3 …), or (b) add a setup-python + pip install packaging step, or (c) drop the packaging dependency and compare versions another way. The setup-uv step is currently dead weight since nothing invokes uv.

🟡 Should fix — [docs/dependency-mirroring.md:1](vscode-webview://14rid8ak7asturkhjli19n9sf8f0jgraqq2l6mtjgkofaqaui5qp/docs/dependency-mirroring.md#L1) — Missing required YAML frontmatter and license header. Every other file in docs/ starts with ---\ntitle: "…"\n--- followed by the Copyright … / SPDX-License-Identifier: MIT-0 header (verified: all docs/*.md comply). This page starts directly with # Dependency Mirroring…, so it will not render correctly on the Starlight docs site and violates the documentation skill convention. Add:


---
title: "Dependency Mirroring for Air-Gapped Builds"
---

Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
SPDX-License-Identifier: MIT-0
🟢 Nice to have — CHANGELOG.md — This is a user-facing developer feature (make dep-manifest, new docs page). Per the pr-review/documentation conventions, add an entry under ## [Unreleased] → ### Added.

🟢 Nice to have — [docs/dependency-mirroring.md:35-39](vscode-webview://14rid8ak7asturkhjli19n9sf8f0jgraqq2l6mtjgkofaqaui5qp/docs/dependency-mirroring.md#L35-L39) — The doc says the workflow "runs automatically when dependency files change on main," but the workflow also triggers on pull_request (dry-run). Minor: the doc undersells the PR dry-run path that the workflow actually implements. Also note the paths filter watches lib/*/pyproject.toml, but the script reads uv.lock too — a uv.lock change without a pyproject.toml change won't retrigger the manifest. Low impact since manifests aren't committed.

🟢 Nice to have — [scripts/generate-dep-manifest.sh:81-83](vscode-webview://14rid8ak7asturkhjli19n9sf8f0jgraqq2l6mtjgkofaqaui5qp/scripts/generate-dep-manifest.sh#L81-L83) — is_path_ref checks startswith((".", "/", "../../")). The "../../" entry is redundant (already covered by "."), and a Windows-style or bare relative path like lib/idp_common_pkg (no leading .) wouldn't be caught. Verified the repo's actual requirements use the ../../lib/... form, so this works today — just brittle. Non-blocking.

🟢 Nice to have — The uv.lock parsing uses a regex split on [[package]] blocks rather than tomllib. It works for the current single lib/idp_common_pkg/uv.lock, but is fragile if lockfile formatting changes. Since tomllib is already imported in the same script, parsing uv.lock with it would be more robust.

Recommendation
Request changes — This is a clean, well-scoped, genuinely useful addition with good security hygiene (SHA-pinned actions, least-privilege permissions, nothing committed). Two small items should be fixed before merge: the CI workflow won't actually run the script successfully as written (missing Python/packaging provisioning, unused uv), and the new docs page is missing the mandatory frontmatter + license header. Both are quick fixes. The CHANGELOG entry is a nice-to-have per repo convention. Once the workflow is verified to run green, this is safe to merge.

rstrahan:
In the interest of speed, I'll merge now, @sirirako  and make these changes,
For future PRs, please do yoiuw own AI review using claude code or cline (repo contains a PR review skill) to catch these things. 
Many tx.

PR #348, opened 2026-06-08
fix: resolve all Python dependencies including transitives

sirirako:
# fix: resolve all Python dependencies including transitives in dep manifest

## Summary

Fixes the dependency manifest script to resolve **all** transitive dependencies, not just direct ones. Previously, 9 packages not found in any `uv.lock` were listed by name only (no version, no transitives) — making the manifest incomplete for JFrog Artifactory pre-warming.

## Problem

The customer reported that pre-warming JFrog with the manifest didn't prevent scan-blocking during builds. Root cause: packages like `opensearch-py`, `sagemaker-mlflow`, and `huggingface-hub` were listed without versions or transitives. Their dependencies (`google-auth`, `grpcio`, `protobuf`, `mlflow-skinny`, etc.) were missing entirely — discovered only at build time, triggering incremental fetch + scan delays.

## Fix

The script now runs `uv pip compile` on the additional packages to produce a fully pinned resolution with all transitives. If resolution fails (e.g., private packages), it falls back to listing names with a clear warning.

**Before:** 217 pinned + 9 unversioned = 226 entries (incomplete)
**After:** 217 from lockfiles + 42 resolved transitives = 259 entries (complete, all pinned)

## Changes

- `scripts/generate-dep-manifest.sh` — After collecting additional packages, writes them to a temp file and runs `uv pip compile` to resolve the full tree. Merges results into the manifest, deduplicating against packages already present from lockfiles.

## Test plan

- [x] `make dep-manifest` completes successfully
- [x] `python-packages.txt` has 259 entries, all `name==version` format
- [x] No unversioned package names in the output
- [x] Known transitives present (e.g., `google-auth`, `protobuf`, `mlflow-skinny` from `sagemaker-mlflow`)
- [x] `node-packages.txt` unchanged (1,515 packages — already complete from `package-lock.json`)


rstrahan:
Looking good @sirirako - merging!  Many thanks!   You are a PR machine these days :)

Issue #372, opened 2026-06-22
publish-pack should mirror publish: private-by-default, require public-read only with --public


rstrahan:
## Summary

`idp-feature-cli publish-pack` (in `lib/idp_feature_sdk`) **unconditionally** treats published pack artifacts as public-read and **fails** publishing if they aren't reachable via anonymous HTTPS. It should instead **mirror the `publish` (feature) command**: private by default, and only expect/require public-read when `--public` is passed.

Observed with `idp_feature_sdk` **v0.5.15** (`@develop`).

## Current behavior

`publish-pack`:
- Has **no `--public` flag** (unlike `publish`, which has `--public` / `make_public`).
- Resolves the bucket with a bare `artifacts_bucket = f"{bucket_basename}-{region}"` (`cli.py` ~L255) instead of `_resolve_bucket(..., make_public=...)` — so it never provisions/relaxes a bucket and never threads a public/private intent through.
- `PackPublisher.publish` (`pack.py`) uploads via `_upload_public(...)` and `put_object(...)` assuming a pre-existing `packs/*` public-read **bucket policy**, then runs an **unconditional** public-reachability HEAD check on the wrapper URL (`pack.py` ~L331-347). On a private bucket this raises:

```
✗ Published wrapper at https://<bucket>.s3.<region>.amazonaws.com/packs/<id>/v<ver>/deploy.yaml
  is not publicly reachable (HTTP 403). The bucket policy did not take effect ...
```

The error message even guesses "account-level S3 Block Public Access", which is misleading: in my case BPA was fully disabled and there was simply **no public-read bucket policy** (correctly, since `publish-pack` never creates one and I didn't pass any opt-in).

Net effect: there is **no way to run `publish-pack` against a private same-account bucket** — it always demands public artifacts, contradicting the secure-by-default model that `publish` and `ensure_artifacts_bucket` already implement.

## Expected behavior (mirror `publish`)

- Add a `--public` / `make_public` flag to `publish-pack` (same help text/semantics as `publish`).
[...]

PR #376, opened 2026-06-23
fix(feature-sdk): make publish-pack/deploy-pack consistent with publish/deploy (#375)

### LINKED ISSUE #375: Make idp-feature-cli publish-pack/deploy-pack consistent with publish/deploy (token baking, layout, args, diagnostics)
## Summary

`idp-feature-cli publish-pack` / `deploy-pack` (in `lib/idp_feature_sdk`) have drifted out of parity with `publish` / `deploy` — different arg names/defaults, and, more importantly, the **pack publisher does not bake the same publish-time tokens the feature publisher does**, nor follow the same artifact layout. The result: a freshly published+deployed pack fails at CREATE because the wrapper carries a stale host contract and the feature template ships with an unsubstituted `<FEATURE_ARTIFACT_PREFIX_TOKEN>`.

Goal of this issue: make `publish-pack`/`deploy-pack` **as consistent as possible** with `publish`/`deploy` (args, semantics, token-baking, layout) and make an end-to-end private same-account pack deploy actually work.

Observed with `idp_feature_sdk` **v0.5.15** (`@develop`). Follows on from #372 (which fixed `--public` parity for `publish-pack`); this issue is the broader parity + correctness pass.

---

## A. Token-baking gap (the functional blocker)

The **feature** publisher (`publisher.py` ~L427-431) bakes ALL publish-time tokens into the template before upload:

```python
baked_text = (
    template_text.replace("<FEATURE_VERSION_TOKEN>", manifest.version)
    .replace("<FEATURE_ARTIFACT_PREFIX_TOKEN>", extension_base)
    .replace("<FEATURE_BUCKET_TOKEN>", bucket)
    .replace("<FEATURE_PRODUCT_CODE_TOKEN>", ...)
    .replace("<FEATURE_LISTING_URL_TOKEN>", ...)
)
```

The **pack** publisher (`pack.py` ~L254-256) bakes **only** the version token:

```python
baked_text = feat_template_local.read_text(...).replace("<FEATURE_VERSION_TOKEN>", version)
```

`grep -c` for the other four tokens in `pack.py` returns **0**. So the pack's published `template.yaml` still contains the literal `<FEATURE_ARTIFACT_PREFIX_TOKEN>` (and PRODUCT_CODE/LISTING/BUCKET if the pack template uses them). The feature stack's ui-deployer fails fast on this:

```python
# ui-deployer/handler.py
for _var, _val in (("FEATURE_VERSION", ...), ("FEATURE_ARTIFACT_PREFIX", ...)):
    if not _val or "TOKEN" in _val:
        raise RuntimeError(f"{_var} env var is unsubstituted/empty ...")
```

**Fix:** the pack publisher must bake the same token set as the feature publisher (VERSION, ARTIFACT_PREFIX, BUCKET, PRODUCT_CODE, LISTING_URL). Ideally factor the token-baking into one shared helper both publishers call, so they cannot drift again.

## B. Artifact layout inconsistency

Three components disagree on artifact location/version format:

| Step | Path | Version format |
|---|---|---|
| `publish-pack` uploads | `packs/<id>/v<version>/...` | `v0.1.0` |
| pack wrapper pre-stager copies to seller bucket | `features/<id>/v<version>/...` | `v0.1.0` |
| feature `template.yaml` ui-deployer reads | `<FEATURE_ARTIFACT_PREFIX>/<version>/...` | `0.1.0` (NO `v`) |

The feature publisher's convention is `extensions/<id>/` (version-free base) + `<base>/<version>/` (no `v` prefix). The pack publisher invents a different `packs/<id>/v<version>/` layout. Whatever `<FEATURE_ARTIFACT_PREFIX_TOKEN>` is baked to MUST match where the artifacts actually land and where the ui-deployer reads them. **Fix:** align the pack layout + baked prefix with the feature publisher's convention (or, if the seller-bucket pre-stage step is retained, ensure the baked prefix points at the pre-staged location with the matching version format).

## C. Stale host/feature contract baked into the example wrapper (`deploy.yaml`)

The accelerator host template was migrated **SellerBucket → FeatureBucket**, but the pack wrapper still references the old contract. Two distinct mismatches:

1. Wrapper passes host params `FeaturePlatformSellerBucket` / `FeaturePlatformSellerBucketRegion` — **these no longer exist** in the host template (it now wires the FeaturePlatform nested stack without any seller-bucket param). CFN fails: `Parameters: [FeaturePlatformSellerBucket, FeaturePlatformSellerBucketRegion] do not exist in the template`.
2. Wrapper passes feature-stack params `SellerBucket` / `SellerBucketRegion`; the feature template now takes a single `FeatureBucket` param (no region; prefix is baked).

This is partly a sample-template (`claims-pack/deploy.yaml`) concern, but it argues for the pack tooling to **own less of the host/feature contract** — see the design question below. (We will fix our local `deploy.yaml` in the marketplace-features repo regardless.)

## D. CLI arg / semantic divergences

Bring the pack commands in line with `publish`/`deploy`:

| Concern | `publish` / `deploy` | `publish-pack` / `deploy-pack` | Desired |
|---|---|---|---|
| `--bucket-basename` | **optional**, auto-bucket `idp-accelerator-artifacts-<acct>-<region>` when omitted | `publish-pack`: **required**, no auto-bucket | make optional w/ same auto-bucket default |
| `--region` default | `publish`: `us-east-1`; `deploy`: AWS session region | both packs: hardcoded `us-west-2` | match: session-region for deploy-pack; align publish-pack with publish |
| `--prefix` default | `publish`: empty → bare `extensions/<id>/`; `deploy`: `idp-cli` | packs: `packs` | reconcile with the layout decision in (B) |
| `--parameters` | `deploy` supports `key=value,...` overrides | only `deploy-pack` has it; `publish`/`deploy` parser shared | reuse the SAME parser (`_parse_parameters`) |
| token baking | all 5 tokens | version only (see A) | shared helper |
| stderr on subprocess failure | n/a | `_sam_build_and_package` runs `sam build`/`package` with `capture_output=True` and re-raises `CalledProcessError` with NO captured stderr — failures are undiagnosable (see E) | surface captured stdout/stderr |

## E. Swallowed subprocess stderr (diagnosability)

`pack.py._sam_build_and_package` calls `subprocess.run([...], check=True, capture_output=True)` for both `sam build` and `sam package`. On failure the `CalledProcessError` propagates with the captured output discarded, so the user sees only a Python traceback ending in `returned non-zero exit status 1` — no AWS error. (We spent real time on an `AccessDenied`/credential-shadowing issue that was completely hidden by this.) **Fix:** catch `CalledProcessError` and re-raise a `RuntimeError` including `exc.stdout`/`exc.stderr` (decoded), as a clear message — matching how the rest of the SDK surfaces errors.

---

## Design question (worth deciding as part of this)

In the post-migration model, a normal feature install has the host pass `FeatureBucket` = the publish bucket, and the feature stack reads artifacts directly from it via IAM (no copy). The pack wrapper's "create a seller bucket + pre-stage artifacts into it via a custom resource" dance may now be **obsolete** — the wrapper could set `FeatureBucket` = the publish bucket and drop the seller bucket + pre-stager entirely. That would:
- delete a lot of wrapper complexity (and the anonymous-fetch / public-read pressure that drove #372),
- make `deploy-pack` behave like `deploy` (read artifacts in place, private by default),
- leave only the host-stack-name-cap workaround (the `HostStackManager` custom resource) as pack-specific.

If that direction is taken, much of (B)/(C) collapses into "packs use the same artifact contract as features." Please advise which way you want it; the marketplace-features repo will update its `claims-pack` sample wrapper to match.

---

## Acceptance criteria

1. `publish-pack` bakes the same publish-time tokens as `publish` (shared helper), so the published feature template has no unsubstituted `<FEATURE_*_TOKEN>`.
2. Pack artifact layout + baked `FEATURE_ARTIFACT_PREFIX` are mutually consistent and match where artifacts are read.
3. `publish-pack`/`deploy-pack` arg names/defaults (`--bucket-basename` optional + auto-bucket, `--region`, `--prefix`, `--parameters`) match `publish`/`deploy`.
4. `sam build`/`package` failures surface captured stderr, not a bare exit code.
5. An end-to-end **private, same-account** `deploy-pack --from-code` of the claims-pack succeeds (host + feature stacks reach CREATE_COMPLETE) with no `--public`.
6. Decision recorded on the seller-bucket-vs-FeatureBucket design question.

## Notes
- Repo context: surfaced from the closed-source marketplace-features repo, but every fix here is in the public accelerator's `lib/idp_feature_sdk` (and the host/feature template contract it targets).
- Related: #372 (publish-pack `--public` parity — already merged to develop).

rstrahan:
## What & why

`idp-feature-cli publish-pack` / `deploy-pack` had drifted out of parity with `publish` / `deploy`, to the point that a freshly published+deployed pack fails at CREATE. Closes the functional gaps and the arg/semantic divergences tracked in #375.

Root problems (all in `lib/idp_feature_sdk`):
- **(A)** The pack publisher baked only `<FEATURE_VERSION_TOKEN>`; the feature publisher bakes all five (`VERSION`, `ARTIFACT_PREFIX`, `BUCKET`, `PRODUCT_CODE`, `LISTING_URL`). So a pack's published `template.yaml` shipped a literal `<FEATURE_ARTIFACT_PREFIX_TOKEN>`, and the feature stack's ui-deployer fails fast on it.
- **(B)** Pack used a bespoke `packs/<id>/v<ver>/` layout vs the feature publisher's `extensions/<id>/` version-free layout — and the baked prefix didn't match where artifacts were read.
- **(C)** Wrapper carried a stale contract (SellerBucket / pre-stage copy fed by anonymous HTTPS).
- **(D)** `--bucket-basename` required (no auto-bucket), `--prefix`/`--region` defaults differed, a `--skip-build` flag `publish` doesn't have.
- **(E)** `sam build`/`package` ran with `capture_output=True` and re-raised with stderr discarded — failures were undiagnosable.

## Approach

**`PackPublisher` now delegates feature-artifact publishing to `FeaturePublisher`.** That single change makes packs inherit, identically: the `extensions/<id>/` version-free layout (B), all five baked tokens (A), the SAM step that surfaces captured stderr on failure (E), and the private-by-default / `--public` ACL semantics. Sharing one publisher is what stops the two from drifting again.

**Seller bucket dropped (design decision in #375).** `publish-pack` bakes the publish bucket + version-free prefix + version into the wrapper's parameter defaults; the feature stack reads artifacts **in place** via IAM, exactly like a normal `deploy`. No pre-stage copy, no anonymous fetch. This also removes most of the pressure behind #372 for same-account deploys.

**Clean break on the wrapper contract:** `wrapperParameters.artifactSourceParam` → `featureBucketParam` + `prefixParam`. Wrappers must migrate (the closed-source claims-pack `deploy.yaml` is updated in lockstep, separately).

**CLI parity (D):** `--bucket-basename` optional with the same auto-bucket default as `publish`; `--prefix` defaults to `""` (bare `extensions/<id>/`); `--region` defaults to `us-east-1`; `--skip-build` removed.

## Changes
- `idp_feature_sdk/pack.py` — delegate to `FeaturePublisher`; bake wrapper `FeatureBucket`/prefix/version; drop the bespoke upload path, `_build_ui`, `_sam_build_and_package`, `_upload_public`, `_content_type_for`, `artifact_source_url`.
- `idp_feature_sdk/manifest.py` + `schemas/feature-manifest.schema.json` — `featureBucketParam`/`prefixParam` replace `artifactSourceParam`.
- `idp_feature_sdk/cli.py` — `publish-pack`/`deploy-pack` arg parity.
- `tests/test_pack.py` — new: layout parity, all-tokens-baked, wrapper defaults point at published artifacts, explicit-prefix propagation, unknown-param error.

## Testing
- `pytest lib/idp_feature_sdk` → **59 passed** (5 new).
- `ruff check` / `ruff format --check` → clean.

## Compatibility note
Breaking for existing pack wrappers: `artifactSourceParam` is gone and the seller-bucket pre-stage is removed. Any `deploy.yaml` must (a) declare params for `featureBucketParam`/`prefixParam`/`versionParam`, and (b) pass `FeatureBucket` + prefix straight to the feature stack instead of creating a seller bucket. The only in-tree pack sample (`subscription-features/feature-platform/claims-pack`) has no wrapper checked in; the closed-source claims-pack wrapper is migrated alongside this.

Closes #375. Related: #372.

### DIFF
--- a/lib/idp_feature_sdk/README.md
+++ b/lib/idp_feature_sdk/README.md
@@ -89,6 +89,27 @@ idp-feature-cli show-schema
 | `deploy`      | Create-or-update **this feature's** stack into an existing host — either `--from-code` (publish first) or `--template-url` (already published). |
 | `deploy-pack` | Deploys a self-contained *pack* that stands up its **own** host stack.       |
 
+A **pack** (`publish-pack` / `deploy-pack`) publishes its feature artifacts
+exactly like `publish` — shared publisher, so the same `extensions/<id>/`
+version-free layout and the same baked tokens — then bakes the publish
+**bucket + version-free prefix + version** into its wrapper template's
+parameter defaults. The pack's feature stack reads those artifacts *in place*
+via IAM; there is no seller bucket and no pre-stage copy. Declare the wrapper
+parameter names under `pack.wrapperParameters` in `feature.yaml`:
+
+```yaml
+pack:
+  wrapperTemplatePath: deploy.yaml
+  wrapperParameters:
+    hostTemplateUrlParam: IdpAcceleratorTemplateUrl
+    featureBucketParam:   FeatureBucket          # bucket holding the artifacts
+    prefixParam:          FeatureArtifactPrefix  # e.g. extensions/<id>
+    versionParam:         FeatureVersion
+```
+
+(The pre-`#375` `artifactSourceParam` / seller-bucket model is gone; wrappers
+that still declare it must migrate to `featureBucketParam` + `prefixParam`.)
+
 `deploy` is the fast inner loop for iterating one extension from source against
 a live IDP deployment. The feature stack is named
 `<host-stack-name>-feature-<feature-id>` by default (override with
--- a/lib/idp_feature_sdk/idp_feature_sdk/cli.py
+++ b/lib/idp_feature_sdk/idp_feature_sdk/cli.py
@@ -204,34 +204,30 @@ def publish(
 )
 @click.option(
     "--bucket-basename",
-    required=True,
+    default=None,
     help="S3 bucket basename for the pack's published artifacts — region is "
-    "appended automatically (matches `idp-cli`). Private by default; pass "
+    "appended automatically (auto-generated as "
+    "idp-accelerator-artifacts-<account>-<region> and created if not "
+    "provided). Matches `idp-cli` / `publish`. Private by default; pass "
     "--public to grant public read on the `packs/*`/`host/*` prefixes for "
     "CROSS-ACCOUNT deploys. Without --public the bucket is left untouched.",
 )
 @click.option(
     "--prefix",
     "artifacts_prefix",
-    default="packs",
-    show_default=True,
-    help="Key prefix under the artifacts bucket. Final layout: "
-    "<bucket>/<prefix>/<feature-id>/v<version>/.",
+    default="",
+    help="Optional S3 key prefix under the artifacts bucket. Default (empty) "
+    "yields the bare `extensions/<id>/...` layout — the SAME layout `publish` "
+    "produces; a non-empty prefix becomes `<prefix>/extensions/<id>/...`.",
 )
 @click.option(
     "--host-template-url",
     required=True,
     help="Public HTTPS URL of the IDP accelerator main template (idp-main.yaml). "
-    "Produced by `python3 publish.py`. Baked into the wrapper as a parameter "
+    "Produced by `idp-cli publish`. Baked into the wrapper as a parameter "
     "default so deploy-pack doesn't need to specify it.",
 )
-@click.option("--region", default="us-west-2", show_default=True)
-@click.option(
-    "--skip-build",
-    is_flag=True,
-    default=False,
-    help="Skip `npm run build` for the UI bundle (assume dist/ is already built).",
-)
+@click.option("--region", default="us-east-1", show_default=True)
 @click.option(
     "--public",
     "make_public",
@@ -240,18 +236,17 @@ def publish(
     help="Make the published artifacts world-readable on the `packs/*` and "
     "`host/*` prefixes (relaxes BlockPublicPolicy/RestrictPublicBuckets and "
     "applies a public-read bucket policy). ONLY needed to share artifacts for "
-    "CROSS-ACCOUNT / Quick-Create pack deploys, where the deploying account's "
-    "pre-stager Lambda fetches the wrapper via anonymous HTTPS. Default: "
-    "private (same-account). A pre-existing bucket's Block Public Access "
-    "settings are never weakened unless this flag is set.",
+    "CROSS-ACCOUNT / Quick-Create pack deploys, where the deploying account "
+    "fetches the wrapper template via anonymous HTTPS. Default: private "
+    "(same-account). A pre-existing bucket's Block Public Access settings are "
+    "never weakened unless this flag is set.",
 )
 def publish_pack_cmd(
     project_dir: Path,
-    bucket_basename: str,
+    bucket_basename: Optional[str],
     artifacts_prefix: str,
     host_template_url: str,
     region: str,
-    skip_build: bool,
     make_public: bool,
 ) -> None:
     """Publish a vertical-product pack as a single-template wrapper.
@@ -276,7 +271,6 @@ def publish_pack_cmd(
             artifacts_prefix=artifacts_prefix,
             host_template_url=host_template_url,
             region=region,
-            skip_feature_build=skip_build,
             make_public=make_public,
         )
     except (ManifestError, RuntimeError, ValueError, FileNotFoundError) as exc:
@@ -288,7 +282,6 @@ def publish_pack_cmd(
     console.print(
         f"  artifacts:        s3://{result.artifact_bucket}/{result.artifact_prefix}/"
     )
-    console.print(f"  artifact source:  {result.artifact_source_url}")
     console.print(f"  feature template: {result.feature_template_url}")
     console.print(f"  host template:    {result.host_template_url}")
     console.print(f"  wrapper template: {result.wrapper_template_url}")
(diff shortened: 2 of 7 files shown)


rstrahan:
Thanks — the blocking finding was spot on. Pushed 1b65d9f.

**🔴 --public broken under the new layout — fixed.** Confirmed all three failure points (stale `packs/*` policy prefix, wrapper uploaded with no ACL, 4b HEAD 403). Fixed to mirror `publish` exactly rather than invent a pack-specific path:
- `_PACK_PUBLIC_PREFIXES`: `packs/` → `extensions/` (the real layout; also covers the host accelerator under `host/`).
- Wrapper now uploaded with `ACL=public-read` when `--public`, identical to every feature artifact — so an explicit-basename public bucket works the same as `publish` without relying on the bucket policy.
- Dropped the divergent in-`publish` `apply_public_artifacts_policy()` call (`publish` has none; the CLI's `_resolve_bucket` already applies the auto-bucket policy). Extracted the HEAD check into `_assert_publicly_readable()`.

On the ACL-vs-policy contradiction you noted: `publish`'s *primary* mechanism is in fact per-object `ACL=public-read` (`FeaturePublisher._upload` + `_set_public_acls`), with the bucket policy only on the auto-created bucket. So this path now matches `publish` byte-for-byte — same ACL behavior, same `--bucket-basename must permit ACLs` caveat. I kept the publisher's existing approach rather than re-architecting it here.

**🟡 No --public test — added.** `test_public_makes_wrapper_and_feature_artifacts_readable` asserts the wrapper *and* feature artifacts are public-read (the HEAD is stubbed since the URL isn't reachable under moto). It fails against the pre-fix code, so it's a real regression guard.

**🟢 Stale docs — partially addressed.** Documented the new `wrapperParameters` (`featureBucketParam`/`prefixParam`) contract in the SDK README. I left `docs/feature-platform*.md` alone: as you noted, those describe the host's `SellerBucketObjectArns` presign mechanism, which is still legitimate and unrelated to pack distribution.

**🟢 CHANGELOG — not touched** (feature-sdk has no prior CHANGELOG presence; out of scope as you suspected).

Full suite: 60 passed; ruff clean.

PR #387, opened 2026-06-26
Feature/per job config

sirirako:
### Summary

  Allows Jobs API callers to specify which configuration version to use when processing a document submission. Without this, all jobs use the stack's
  active/default configuration — there's no way to pin a specific version per job.

  ### Problem

  External systems integrating via the headless Jobs API have no way to select a processing configuration at submission time. If you have multiple config
  versions (e.g. `lending-v1`, `lending-v2`), every job uses whatever is currently active — you can't A/B test or gradually migrate callers.

  ### Solution

  Add an optional `configurationVersion` field to `POST /jobs`. The pipeline already supports per-document config selection via S3 object metadata
  (`config-version`) — this change wires the API to set that metadata at upload time.

  **Flow:**
  1. Caller passes `configurationVersion` in the POST request
  2. API includes `x-amz-meta-config-version` in the presigned POST fields
  If `configurationVersion` is omitted, behavior is unchanged (stack default).

  ### Changes

  | File | Change |
  |------|--------|
  | `src/lambda/api_handler/models.py` | Add `configurationVersion` field with validation (max 128 chars, `^[a-zA-Z0-9._-]+$`) |
  | `src/lambda/api_handler/index.py` | Set S3 metadata on presigned POST, store on job record, surface in GET response |
  | `lib/idp_common_pkg/idp_common/dynamodb/job_service.py` | Accept + persist `configuration_version` on job record |
  | `src/lambda/batch_pre_processor/index.py` | Read version from job record, propagate as S3 metadata on extracted files |
  | `src/lambda/api_handler/test_index.py` | 5 new tests |
  | `src/lambda/batch_pre_processor/test_index.py` | 4 new tests |

  ### API

  **Request (new optional field):**
  ```json
  POST /jobs
  {
    "fileName": "documents.zip",
    "configurationVersion": "lending-v2"
  }

  Response (GET /jobs/{id} now includes it):
  {
    "jobId": "...",
    "status": "IN_PROGRESS",
    "configurationVersion": "lending-v2",
    ...
  }

  Validation

  - configurationVersion is validated: max 128 characters, alphanumeric + ., _, - only
  - Invalid values return 422

  Testing

  - 9 unit tests covering: with/without version, invalid input, S3 metadata propagation, GET response, graceful DDB fallback
  - E2E tested: deployed to a stack with EnableHeadless=true, invoked Lambda directly, confirmed x-amz-meta-config-version appears in presigned POST fields

  Backward compatibility

  - Field is optional, defaults to null
  - Omitting it produces identical behavior to before (no metadata, pipeline uses active config)
  - No template/infrastructure changes required


rstrahan:
## Review

Thanks for this — focused, well-tested (43 passing), backward-compatible, and I traced the full data path end-to-end (POST → DDB `ConfigurationVersion` → batch pre-processor re-applies `config-version` S3 metadata on extracted files → `Document.from_s3_event` → `queue_processor`). The working path checks out.

### Verdict

| Question | Verdict | Notes |
|---|---|---|
| Good PR? | ✅ | Focused, well-described, targets `develop`. |
| Safe / no regressions? | ✅ | Optional field, null default; omitting it is identical to prior behavior. |
| Good UX? | ⚠️ | Good validation + graceful DDB fallback; no feedback when version doesn't exist. |
| No security issues? | ✅ | Regex+length-bounded input, no secrets/ARNs/wildcards. |
| Well documented? | ⚠️ | Missing CHANGELOG + `docs/govcloud-batch-api.md` update. |
| **Safe to merge?** | ⚠️ after small fixes | |

### Should fix
- 🟡 **Redundant presigned-POST metadata path** — the `x-amz-meta-config-version` field/condition set on the staging zip (`api_handler/index.py`) is never read: `batch_pre_processor` gets the version from the DDB job record (not from the zip's S3 metadata) and re-applies it to the extracted files. The presigned-POST condition also *narrows* behavior — it forces the uploader to echo that field or get a hard S3 rejection. Recommend dropping the `fields`/`conditions` changes and keeping the DDB-driven path as the single source of truth.
- 🟡 **CHANGELOG** — add an `## [Unreleased] → ### Added` entry for this user-facing API field.
- 🟡 **Docs** — update `docs/govcloud-batch-api.md` (POST request field, GET response field, validation rules / 422).

### Nice to have
- 🟢 No existence check on the supplied version — an unknown version submits 200 then falls back to pipeline mode / errors downstream with no clear signal.
- 🟢 `POST` response doesn't echo `configurationVersion` back (only GET does).

I'll push the three "should fix" items to this branch.

PR #439, opened 2026-07-05
fix(confidence): recover truncated assessment batches instead of losing scores

rstrahan:
## Problem

On an agentic run of a lending-package bank statement, **50 of 68 transaction rows had no confidence scores or bounding boxes**, and every scalar field came back at a flat `confidence: 0.5`.

Root cause (traced through the deployed extraction Lambda logs): the confidence model (**Amazon Nova Lite**, 10,000 max output tokens) was **truncating every assessment call**. This config uses `geometry.mode: llm`, so the assessment prompt asks for a bounding box on every cell — that roughly doubles per-row output (measured ~55% of each row's JSON is geometry), pushing even a 25-row batch past the token ceiling. A truncated response is unparseable JSON, and the service silently:

- assigned the parse-failure fallback `confidence: 0.5` ("Unable to parse assessment response") to every **scalar**, and
- left most **list rows** as null-confidence placeholders ("Not individually assessed…").

Every assessment call — including the batched missing-row retries — hit exactly `outputTokens: 10000` and failed to parse, with no error surfaced to the operator.

## Fix (three layers)

1. **Detect truncation** — `AssessmentCoreResult.truncated`, set from the Converse `stopReason` (`max_tokens`) and the OpenAI-Responses adapter's `status` (`incomplete` / `max_output_tokens`). The parse-error log now distinguishes a truncated response from genuinely malformed JSON.
2. **Adaptive batch-split recovery** — any slice the model truncates is **recursively halved and re-assessed** until it parses or bottoms out at a single row, instead of accepting the placeholder. Runs in both the initial batch loop and the missing-row retry, so it protects the **standalone, single-agent, and sharded** assessment paths (all share `assess_results_batched`).
3. **Visibility** — surfaced only when a run actually had to shrink:
   - `metadata.assessment_batch_split_stats` on the section result (`truncated_calls`, `splits`, `min_batch_size_used`, `rows_recovered_by_retry`, `unrecoverable_rows`)
   - an `⚠ Assessment Batch Splitting` block in the agentic **processing report**
   - aggregated across shards for the sharded path

If rows remain unscored even at a single-row batch, `unrecoverable_rows` is non-zero; the durable fix is then to reduce per-row output (e.g. `geometry.mode: ocr_only`, the default).

## Verification

Reproduced the exact 68-row scenario (batch 25, Nova truncating): **before → 18/68 rows scored; after → 68/68**, with the split activity fully recorded in `split_stats`.

- **7 new unit tests** (batching split / bounded-recursion / cross-shard merge + service-level truncation detection), all pass.
- Full assessment + extraction suite: **386 passed** (1 unrelated pre-existing failure that fails identically on `develop`).
- `ruff` clean; `basedpyright` 0 errors on all changed files.

## Docs

- `lib/idp_common_pkg/idp_common/assessment/README.md` — new *Truncation-aware adaptive batch splitting* section
- `docs/extraction-and-confidence.md` — *Automatic recovery when the model truncates* note
- `CHANGELOG.md` — Fixed entry

## Notes for reviewers

- This is a **safety net**: the underlying trigger is `geometry.mode: llm` asking a small-cap model for a bbox per cell. The change makes that degrade gracefully and visibly rather than silently; `ocr_only` remains the cheaper/cleaner config for that workload.
- Public method signatures are unchanged; all new behavior lives inside the shared `assess_results_batched` path and is additive.


rstrahan:
### Follow-up: extended truncation recovery to the `integrated` path

Reviewer asked whether **simple mode** hits the same truncation and whether the recursive batch-reduction should apply to both simple and agentic. Answer: yes — and there was a gap.

There are effectively four confidence code paths. The first commit covered three (all funnel through the shared `assess_results_batched`):
- standalone Assessment step (`separate`) — **this is the simple-mode default and the granular-assessment replacement, so it was already covered**
- agentic single-agent in-shard
- agentic sharded in-shard

The one gap was **`integrated` mode** (simple *or* agentic): its inline-confidence missing-row retry (`_retry_missing_integrated_rows`) had its own fixed-batch-size loop and did not split on truncation. Commit `d8200de` refactors that loop to delegate to the same shared `_assess_slice_adaptive` helper, so **all paths now share one truncation-recovery implementation and one `split_stats` shape**, and integrated split activity is surfaced in `metadata.assessment_batch_split_stats` + the processing report (I reordered `_attach_explainability` to run before the report for parity).

+2 unit tests; docs/CHANGELOG updated to state coverage is uniform across paths.

rstrahan:
### Addressed review findings

**🔴 Blocking (rebase/conflicts) — resolved.** Rebased onto latest `develop` (now on top of #438). The only conflict was `CHANGELOG.md` (both PRs added a `### Fixed` entry); resolved by keeping both. Branch is now **MERGEABLE** and **CI is green** (Lint, Type Check, and Test → SUCCESS).

**🟡 Metering under-reporting — fixed** (commit `count truncated calls in metering`). Good catch. When a truncated slice was split, `_assess_slice_adaptive` merged only `left + right` sub-slice metering and **discarded the truncated parent call's `core.metering`/`duration`** — but a truncated call still burns real output tokens (up to the model max), so cost was under-reported by exactly the wasted work this PR exists to surface. Now the parent call's metering + duration are folded in before recursing. New test asserts it: the 5-row scenario reports **30,150** output tokens (3 truncated ×10k + 3 clean ×50) vs the old buggy **150**.

Also (from an earlier review round) added a **§0 "Choosing a configuration" decision guide** to `docs/extraction-and-confidence.md` — pros/cons tables for Simple vs Advanced extraction and separate vs integrated vs off confidence, and clarified that large-list batching + recursive truncation recovery apply to **all** confidence paths.

rstrahan:
### ✅ Verified live end-to-end (not just unit tests)

Deployed this branch to a real stack and reprocessed the original failing document (`lending_package-long.pdf`, agentic + `geometry.mode: llm` + Nova Lite confidence — the exact config that produced the bug). Real Nova Lite truncation, real Lambda:

**Before:** 18/68 transaction rows scored, 50 null placeholders, scalars at flat 0.5, no signal.

**After (this run):**
- `transaction_details`: **68/68 rows scored, 0 null placeholders**
- `metadata.assessment_batch_split_stats`:
  ```json
  { "truncated_calls": 1, "splits": 1, "min_batch_size_used": 7,
    "rows_recovered_by_retry": 7, "unrecoverable_rows": 0 }
  ```

Real truncation detected → batch split (25 → 7 rows) → all rows recovered → 0 unrecoverable, and the activity is visible in metadata as intended. Confirms detection, recovery, and visibility all work on the deployed stack.

PR #457, opened 2026-07-09
perf(agentic): stop re-sending parsed table rows through the agent loop

rstrahan:
## Problem
Advanced (agentic) extraction is 3–10× more expensive than simple mode **even when the
deterministic table tool is used**. Measured root cause (Strands v1.14.0 source + live
metering): `parse_table` / `map_table_to_schema` returned a dict with a `status` key but
**no `content` key**, so Strands' `_wrap_tool_result` fell back to inlining `str(result)`
— the *entire parsed row set* — into `agent.messages`. Strands re-sends the whole message
list every turn of the ~8–11-turn agent loop, so the dense rows were re-transmitted N times
as **input** tokens (adv-400-row: **1.46M input tokens** vs simple's 28.7K).

The rows already flow through `agent.state` (`last_parse_table_result` → used by
`finalize_table_extraction`), and `AgentState` is explicitly *not* part of model context —
so the return-value echo was pure waste.

## Fix
Return a dict with **both** `status` and `content`, where `content` is a compact summary
(columns + counts + head/tail sample + a "rows are in state, call the next tool" note). This
flips Strands to the preformatted-ToolResult path; Bedrock's request formatter transmits only
`content` + `toolUseId` + `status`, dropping the full `rows` keys (kept for state/pipeline/tests).
Tables ≤ `INLINE_ROW_CAP` (50) still inline all rows so the manual/fallback workflow is unchanged.

## Evidence
- Verified three ways (Strands source, real `DecoratedFunctionTool`, real `BedrockModel`
  request formatter): a 400-row result now emits ~805-char content instead of ~107K; the full
  `rows` key is absent from the transmitted request. Local experiment: **~98–100% reduction**
  in per-turn re-transmitted payload.
- 6 new unit tests (`test_tool_writeback.py`) + 134 existing table tests pass; 26/26
  `test_map_table_to_schema` pass in isolation (a full-run monkeypatch-leak batch is
  pre-existing on develop, unrelated).

## ⚠️ Needs live A/B before merge
This changes a core extraction path. A live A/B on a scratch stack is running to confirm
(1) the cost drop materializes end-to-end, (2) completeness stays 1.0, and (3) extraction
accuracy has parity vs the current develop behavior. `INLINE_ROW_CAP` is the tuning knob.
Follow-ups: same treatment for `view_buffer_data*`, make the cap a config field.


rstrahan:
## ⚠️ Live A/B result — HOLD: cost goal NOT validated (no functional regression)

Ran this on a deployed stack (advanced mode, Textract TABLES, escalation off, table tool confirmed used), multiple repeats:

| doc | AFTER (writeback) input tokens | BEFORE (develop, 1 sample) |
|-----|-------------------------------|----------------------------|
| n100 (100 rows) | 115K / 115K / 152K / 207K (mean 147K) | 78.9K |
| n400 (400 rows) | 480K / 771K / 1,986K (mean 1,079K) | 1.46M |

- **No functional regression:** completeness recall 1.0 and scalar accuracy 1.0 on every sample; 0 failures.
- **But the cost goal is not demonstrated.** Agentic extraction cost is dominated by run-to-run
  **turn-count variance** — n400 spans ~4× (480K→1.99M tokens) run to run. The writeback effect, if real, is
  smaller than that noise. The ~98% reduction the prototype measured was on the *tool return payload in
  isolation*; it does not translate to total-input savings in the live loop, where OCR text re-sent per turn
  + variable turn count dominate.
- Verified the deployed build **does** contain the writeback code, so this is an effect-size question, not a
  deploy miss.

**Recommendation:** do not merge on a cost basis yet. Needs a controlled retest — writeback-only build vs
clean develop, ≥5 repeats each, comparing distributions (ideally with turn count held constant), and likely
a broader fix that also keeps re-sent OCR text / caps turns. Keeping the PR open as a correctness-preserving
prototype pending that experiment.

rstrahan:
## Update: added lazy-images + cache-aware cost analysis

### New commit: lazy page images on the table path
When the pre-flight table parser already parses the doc's table(s), the agentic loop is
text/markdown-driven (the parser never reads images) and `view_image` covers the rare case a
page must be seen — so we no longer pre-load page images. New knob
`extraction.agentic.table_parsing.lazy_images` (default true; set false for image-dependent corpora).
Local+live A/B: **identical completeness/accuracy (recall 1.0)** with images off on this path.

### Cache-aware cost efficiency (dollars, not tokens) — this corrects the earlier token-only read
The cachepoint sits AFTER the OCR text + images, so those are cache-**written once + read at 10%/turn**,
while the growing conversation (tool results/reasoning) AFTER the cachepoint is **full price every turn**.
Real live metering, n100 advanced extraction ($0.765 total):

| term | $ | share |
|------|-----|-------|
| **inputTokens (full price, after-cachepoint growth)** | **$0.45** | **59%** |
| output | $0.17 | 22% |
| cacheWrite | $0.10 | 13% |
| cacheRead | $0.05 | 6% |

Implications:
- The **59% full-price input term is exactly what the writeback fix targets** → in dollars the writeback
  change is the **bigger lever on typical docs** (its ~16% token cut lands on the most expensive term).
- **Images are largely cached**, so lazy-images saves only ~4% on a 3-page doc — but the saving scales with
  pages×turns: **~$0.11 @9pg, ~$0.23 @17pg, ~$1.1 @66pg, ~$2.4 @131pg**, and it relieves context-window
  pressure on the large docs that otherwise risk overflow.

**Net:** both changes are justified and complementary — writeback wins on typical docs, lazy-images wins on
large multi-page tables. Completeness/accuracy preserved (recall 1.0) in every arm. A clean 4-arm live $
comparison (≥5 repeats, turn-count controlled) on a deployed lazy-images build is the remaining validation
before claiming a single headline % number.

Full analysis: scratch/ab/COST_RESOLUTION.md.

rstrahan:
## ✅ Cost goal VALIDATED — controlled 4-arm A/B (n=5/arm, cache-aware $)

Ran a clean 2×2 factorial (writeback {off,on} × images {on,off}) on identical 100-row docs, advanced
mode, Sonnet 5, escalation off. Cost is cache-aware (input + output + cacheRead@10% + cacheWrite@125%),
from real per-doc metering.

| arm | writeback | images | n | mean $ | full-price input $ | vs baseline | recall/acc |
|-----|-----------|--------|---|--------|--------------------|-------------|-----------|
| A baseline | OFF | ON | 5 | $0.916 | $0.624 | — | 1.0 / 1.0 |
| C images-off | OFF | OFF | 3 | $0.694 | $0.463 | −24% | 1.0 / 1.0 |
| **B writeback** | ON | ON | 5 | **$0.396** | **$0.150** | **−57%** | 1.0 / 1.0 |
| **D both (default)** | ON | OFF | 5 | **$0.389** | $0.162 | **−58%** | 1.0 / 1.0 |

### Findings
- **The writeback fix is the dominant lever: −57% extraction cost.** It cuts the full-price input term
  $0.624 → $0.150 (−76%) — confirming the re-sent parsed rows were the bulk of the after-cachepoint,
  full-price tokens. (My earlier "not validated" was a measurement artifact: agentic turn-count variance
  + a single baseline sample. With n=5 controlled arms the win is unambiguous.)
- **Lazy-images alone: −24%.** On top of writeback it's marginal on this 3-page doc (writeback already
  removed most re-sent content), but it scales with page count (~$1–2.4/doc at 60–131 pages) and relieves
  context-window pressure on large docs.
- **No regression: completeness recall 1.0 and scalar accuracy 1.0 across all 18 runs.**
- Verified the deployed build genuinely suppressed images on the lazy arm (suppression log fired).

**Recommendation: this PR (writeback + lazy-images, arm D) is the default — ~57–58% cheaper on typical
docs, more on large multi-page tables, zero quality loss. Cost goal achieved; ready to merge pending review.**

Full data: scratch/ab/RESULT_4arm.md.

rstrahan:
## Review findings addressed (commit 2f5ef6951)

- 🟡 **#1 status fold mutating shared state** — Fixed. `parse_table` no longer overwrites the `status`
  of the object stored in `agent.state["last_parse_table_result"]`. The Bedrock-facing return status is
  folded to `success`/`error`, while the parser's semantic status (`success`/`no_tables_found`/`error`) is
  preserved under a new `parser_status` key; docstring updated to the new contract. Added a regression test.
- 🟡 **#2 global monkeypatch leak** — Fixed. Replaced the module-level, never-restored `strands.tool = …`
  with a passthrough on the *bound* `table_parser.tool` name plus an **autouse fixture that restores the
  prior value per-test**. The file is now import-order independent: verified `test_tool_writeback` +
  `test_map_table_to_schema` pass in **both** orderings (previously writeback-first broke map 25/26).
- 🟡 **#3 missing CHANGELOG + docs** — Added. `[Unreleased]` CHANGELOG entry for the agentic cost work;
  `lazy_images` documented in `docs/extraction-and-confidence.md` (table_parsing block) and
  `config_library/README.md`.
- 🟢 **#4 DRY the two summary blocks** — Done. Shared `_build_row_summary(rows, count, rows_key, next_tool)`
  helper now backs both `parse_table` and `map_table_to_schema`.
- 🟢 **#5 configurable INLINE_ROW_CAP / SAMPLE_ROWS** — Deferred as a follow-up (noted in PR body); the
  constants are unchanged for now.

Lint clean; 348 extraction unit tests pass (all orderings). Ready for re-review.

rstrahan:
## Advanced-extraction config UI cleanup (commit 3ae14297f)

Addressing review of the advanced-extraction settings:

1. **Removed `Shard Token Budget` from the UI.** It's auto-sized from the model's context window (config default `0` = auto). Kept in the config model as an optional YAML-only override so existing configs that set it still work — no migration break.
2. **`Max Concurrent Batches` default fixed 1 → 10** (matched `base-extraction.yaml`); corrected the misleading "1 = single pass (default)" text.
3. **`Max Pages Per Shard` kept** (it's the wall-clock/timeout guard that is *not* auto-derived — a roomy token budget must not collapse a huge doc into one long-running agent) with a compact description.
4. **Schema Validation & Escalation:** `Check Formats`, `Fail Action`, `Min Population Ratio` now `dependsOn enabled=true`; `Escalation Model` also `hideWhen validation.enabled=false` — all hidden when validation is off.
5. **Table Parsing Tool:** `Min Confidence Threshold`, `Min Parse Success Rate` (+ the other tuning knobs, incl. the new `lazy_images`) now `dependsOn enabled=true` — hidden when table parsing is off.

Also trimmed the now-redundant "Only used when … enabled" prefixes from the gated fields. Template parses (CFN-aware); config round-trips verified (auto default + override both preserved).

PR #459, opened 2026-07-09
feat(evaluation): show per-field method + weight for nested comparisons

rstrahan:
## Summary

The evaluation report's top-level attributes table shows each field's comparison **Method** and **Weight**, but for **arrays of objects** only the list-level `HUNGARIAN` method was shown — each *item field's* own comparator/weight (e.g. `LineItems[].Amount` → `NumericExact`) was hidden in both:

1. the **"Nested Field Comparison"** markdown table (Document detail → *View Evaluation Report*, and version comparison), and
2. the **JSON Visual Editor "Show Evaluation"** overlay (Sections panel).

This PR surfaces per-nested-field **Method** and **Weight** in both places, consistent with the top-level table.

## Why re-derive instead of read from Stickler?

Stickler computes each nested field's comparator/weight internally but **drops them** from its `field_comparisons` output (only expected/actual keys+values, match, score, reason survive). Rather than fork the dependency, we re-derive them on read from the already-translated schema (`x-aws-stickler-comparator` / `-threshold` / `-weight`). No re-run of evaluation is required for the report.

## Changes

- **`service.py`** — extracted `_format_evaluation_method()` as the single source of truth for the Method string (now used by both the top-level table and nested annotation, so they can't drift). Added `_annotate_nested_comparison_methods()` + `_resolve_leaf_schema()` (walks array `items` / object `properties`, ignoring `[idx]`) to annotate each `field_comparison_details` entry with `evaluation_method` + `weight`.
- **`models.py`** — added **Weight** and **Method** columns to the Nested Field Comparison table (order matches the main table: `Match | Score | Weight | Method | Reason`).
- **`VisualEditorModal.tsx`** — added `buildComparisonKey()` + `findNestedComparisonDetail()` for an exact match on the canonical `expected_key`. This also fixes a latent gap: the old leaf matcher only fired when a detail's `actual_value` was an *object*, so **per-list-item scalar leaf fields got no eval decoration at all** — they now show status plus a compact `Method: X · Weight: Y` sub-line.
- **Tests** — added `test_resolve_leaf_schema` and `test_annotate_nested_comparison_methods`.
- **CHANGELOG** — entry under `### Changed`.

## Backward compatibility

Evaluation results serialized **before** this change lack the annotated keys, so they render with an empty method and a default weight of `1.00` (rather than breaking). Newly generated evaluations get full method/weight. No config migration needed.

## Testing

- Backend: `71 passed` in `tests/unit/evaluation/` (incl. 2 new tests); verified end-to-end against real Stickler `compare_with` output for a nested `LineItems` list.
- UI: `eslint` clean, `tsc --noEmit` clean, `vite build` succeeds.


rstrahan:
## Self-review (high-effort, multi-angle) — 1 blocker found & fixed

Ran an adversarial code review over the diff. Summary:

**🔴 Blocker (fixed in 90e0420):** `buildComparisonKey` in `VisualEditorModal.tsx` did `[...path, fieldKey]`, but the render `path` already ends with the current field key (children render with `path={[...path, key]}`; documented invariant at line ~650). This duplicated the last segment (`LineItems[0].Description.Description`, `Agency.Agency`), so `findNestedComparisonDetail` never matched a backend `expected_key` — the new Method/Weight sub-line would have **silently never rendered** (dead feature). Now builds the key from `path` alone; verified keys exactly match backend format for list-item leaves, nested-object leaves, scalar-array items, and top-level scalars.

**🟡 Latent regression (guarded, same commit):** with the key fixed, the keyed match runs before the legacy match and could capture an object/AggregateObject group node sharing a path. Restricted the keyed match to **scalar (non-object/array) leaves** so existing group/array eval handling is untouched.

**🟡 Display consistency (fixed, same commit):** a nested field that is itself a list now passes its own `x-aws-stickler-match-threshold` so it shows the field-level Hungarian threshold, matching the top-level table.

**Verified no-ops:** `_format_evaluation_method` is behavior-identical to the prior inline logic (diffed against `develop`); the `evalWeight`/`.toFixed` guards and JSX render conditions are safe (no falsy-`0` pitfalls).

Two known pre-existing limitations are intentionally left as-is (shared with the existing `field_comparison_map` grouping, harmless, no crash): empty `expected_key` edge case, and field names containing literal `.`/`[`.

**Post-fix checks:** backend `71 passed` (incl. 2 new tests); UI eslint + `tsc --noEmit` clean; `vite build` succeeds.

PR #463, opened 2026-07-10
fix: replace urlparse with parse_s3_uri to handle '#' in S3 keys

justinmcgowanlsi:
## Summary

`urlparse` treats `#` as a URL fragment delimiter (RFC 3986), silently truncating S3 keys at that character. Documents with `#` in their filenames cause `NoSuchKey` errors in every Lambda pipeline stage that calls `Document.decompress()` or `S3Util.s3_url_to_bucket_key()`.

The fix replaces all three `urlparse`-on-S3-URI call sites with the existing `parse_s3_uri()` utility (`idp_common.utils`), which uses `str.split("/", 3)` and is not subject to fragment parsing.

## Affected call sites

| File | Symbol | Impact |
|---|---|---|
| `idp_common/models.py` | `Document.decompress()` | **Primary** — all 12 Lambda handlers call `Document.load_document()` → `decompress()`. Any document with `#` in its filename fails with `NoSuchKey` at every processing stage. |
| `idp_common/utils/s3util.py` | `S3Util.s3_url_to_bucket_key()` | All callers receive a truncated key for `#`-bearing URIs. |
| `idp_common/reporting/save_reporting_data.py` | `SaveReportingData._parse_s3_uri()` | Reporting data for documents with `#` in filenames written to wrong S3 path. |

## Root cause

```python
# BEFORE — urlparse strips everything from '#' onward
parsed_uri = urlparse("s3://bucket/path/invoice #123.pdf")
s3_key = parsed_uri.path.lstrip("/")
# s3_key == "path/invoice "  ← truncated, causes NoSuchKey
```

```python
# AFTER — parse_s3_uri uses str.split, safe for any key character
_, s3_key = parse_s3_uri("s3://bucket/path/invoice #123.pdf")
# s3_key == "path/invoice #123.pdf"  ✓
```

`parse_s3_uri` already exists in `idp_common.utils` (`utils/__init__.py` lines 52–71) and validates the `s3://` prefix with a `ValueError` on malformed input.

## Tests added

- `test_document_compression.py::TestDocumentCompression::test_decompress_key_with_hash` — full `compress → decompress` round-trip with `#` in `input_key`; asserts recovered key is untruncated.
- `test_s3util.py::TestS3UtilParsing::test_s3_url_to_bucket_key_with_hash` — asserts `S3Util.s3_url_to_bucket_key` returns the full key including `#`.
- Regression tests for plain URIs (no `#`) confirm no behavior change on normal inputs.

## Test plan

- [ ] `cd lib/idp_common_pkg && python -m pytest tests/unit/test_document_compression.py tests/unit/test_s3util.py -v`
- [ ] `make lint` (ruff) — no new violations
- [ ] `make test` — full suite green

## Notes

- `%23`-percent-encoded filenames are a separate concern and out of scope for this PR.
- Truncated keys stored in DynamoDB job-state rows before this fix are not remediated here — those would require a separate data-cleanup pass.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/lib/idp_common_pkg/idp_common/models.py
+++ b/lib/idp_common_pkg/idp_common/models.py
@@ -899,9 +899,10 @@ class Document:
             Full Document object with all content restored
         """
         import logging
-        from urllib.parse import urlparse
 
         import boto3
+
+        from idp_common.utils import parse_s3_uri
 
         logger = logging.getLogger(__name__)
         s3_client = boto3.client("s3")
@@ -912,8 +913,7 @@ class Document:
             if not s3_uri:
                 raise ValueError("No s3_uri found in compressed data")
 
-            parsed_uri = urlparse(s3_uri)
-            s3_key = parsed_uri.path.lstrip("/")
+            _, s3_key = parse_s3_uri(s3_uri)
 
             # Retrieve full document from S3
             response = s3_client.get_object(Bucket=bucket, Key=s3_key)
--- a/lib/idp_common_pkg/idp_common/reporting/save_reporting_data.py
+++ b/lib/idp_common_pkg/idp_common/reporting/save_reporting_data.py
@@ -11,7 +11,7 @@ import json
 import logging
 import re
 from typing import Any, Dict, List, Optional
-from urllib.parse import urlparse
+from idp_common.utils import parse_s3_uri
 
 import boto3
 import pyarrow as pa
@@ -126,14 +126,7 @@ class SaveReportingData:
         Returns:
             Tuple of (bucket, key)
         """
-        parsed = urlparse(uri)
-        if parsed.scheme != "s3":
-            raise ValueError(f"Not an S3 URI: {uri}")
-
-        bucket = parsed.netloc
-        # Remove leading slash from key
-        key = parsed.path.lstrip("/")
-
+        bucket, key = parse_s3_uri(uri)
         return bucket, key
 
     def _infer_pyarrow_type(self, value: Any) -> pa.DataType:
--- a/lib/idp_common_pkg/idp_common/utils/s3util.py
+++ b/lib/idp_common_pkg/idp_common/utils/s3util.py
@@ -8,7 +8,7 @@ Utility functions for working with Amazon S3.
 import json
 import boto3
 from typing import Dict, Any, Tuple, Optional, Union
-from urllib.parse import urlparse
+from idp_common.utils import parse_s3_uri
 
 
 class S3Util:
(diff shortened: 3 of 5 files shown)


rstrahan:
Thanks @justinmcgowanlsi !  

Claude's review:

### Findings

- 🔴 __Blocking__ — `lib/idp_common_pkg/idp_common/reporting/save_reporting_data.py:14` — The new import `from idp_common.utils import parse_s3_uri` is inserted into the stdlib import group (right after `from typing import ...`), leaving the block unsorted. I reproduced a __ruff `I001` (Import block is un-sorted or un-formatted)__ on the PR-head version of this file. This file is __not__ in ruff's `extend-exclude` list, so `make lint` will fail. Fix: move it into the first-party group with the other `from idp_common...` imports (ruff `--fix` sorts it correctly).

  - Note: the identical mis-ordering in `utils/s3util.py:11` does __not__ fail lint only because `lib/idp_common_pkg/idp_common/utils` is in ruff's `extend-exclude`. It's still worth fixing for consistency (🟢).

- 🟡 __Should fix__ — `CHANGELOG.md` — This is a user-facing bug fix (documents with `#` in filenames failing at every stage), and the repo convention is to log user-facing changes under `## [Unreleased]`. Add a short `### Fixed` entry.

- 🟡 __Should fix (behavior change to verify)__ — `s3util.py` `s3_url_to_bucket_key` and `save_reporting_data.py` `_parse_s3_uri` — the old code accepted an S3 URI and returned an __empty key__ for inputs like `s3://bucket` / `s3://bucket/`. `parse_s3_uri` instead __raises `ValueError`__ when there are fewer than 4 `/`-split parts (no key). This is arguably more correct, but it's a contract change for any caller that previously tolerated a bucket-only URI. Worth a quick confirmation that no caller relies on the empty-key behavior (and ideally a test asserting the new `ValueError` for a keyless URI).

- 🟢 __Nice to have__ — `test_document_compression.py` new test — it creates `self.bucket` but the assertion only checks `restored.input_key`. Consider also asserting the round-trip doesn't raise and that a `#`-key object is actually retrievable, to lock in the `NoSuchKey` regression specifically.

- 🟢 __Nice to have__ — The PR body notes `%23`-encoded names and pre-existing truncated DynamoDB rows are out of scope. Good scoping — no action needed.

### Recommendation

__Request changes.__ The fix itself is correct, well-explained, and appropriately tested. The only true blocker is the ruff `I001` import-ordering failure in `save_reporting_data.py`, which will break `make lint`. Add a CHANGELOG entry and confirm the keyless-URI `ValueError` change is safe, and this is good to merge.



Are you in a position to have your Claude address these in an update?  
If not, LMK, and I'll have my Claude do it :) 

Thanks!!

rstrahan:
Thanks again @justinmcgowanlsi for finding and fixing this — great catch, and a well-documented PR!

Rather than wait on the review follow-ups, I've carried your fix forward in #477, which keeps your commit (with your authorship) intact and adds the items from the review on top:

- import ordering fix in `save_reporting_data.py` (ruff `I001` would have failed `make lint`) and `s3util.py`
- `CHANGELOG.md` entry
- verified the keyless-URI `ValueError` contract is unchanged for all existing callers, with tests locking it in
- strengthened `test_decompress_key_with_hash` so the compressed-state S3 URI itself contains `#` (the original version passed even against the pre-fix code, since `id` didn't contain `#`) — confirmed it now fails against the old code and passes with the fix
- rebased onto latest `develop`

Closing this one in favor of #477. Your contribution is very much appreciated — the fix itself and the root-cause analysis are all yours. 🙏

justinmcgowanlsi:
Thank you!

PR #549, opened 2026-07-23
fix: handle '#' in S3 object keys in getFileContents resolver and processresults

fhuthmacher:
## Problem

`urllib.parse.urlparse` treats `#` in an S3 URI as a URL-fragment delimiter and silently truncates the key. Object keys legitimately contain `#` (e.g. `Report_#2.pdf/pages/1/result.json`), so `urlparse`-based parsing yields a wrong key and `NoSuchKey` downstream:

- The UI's **View Data / View Page Text** actions fail with *"No response from getFileContents"* for any document whose name contains `#` — the URI `s3://…/Report_#2.pdf/pages/1/result.json` is parsed to key `Report_` (truncated at `#`).
- The processresults functions write `.metadata.json` files under truncated keys.

The library-level fixes for this bug class already landed via `idp_common.utils.parse_s3_uri` (used by `s3util`, `reporting`, and `models.decompress`); this PR ports the remaining Lambda-level cases.

## Changes

- **`nested/api-resolvers/src/lambda/get_file_contents_resolver/index.py`** — parse the `s3://` URI with a plain string split (strip scheme, split once on `/`) so `#` is preserved. Inline rather than `parse_s3_uri` because this Lambda has no `idp_common` dependency. Validation behavior (scheme check, empty-key check, bucket allow-list) is unchanged.
- **`patterns/unified/src/processresults_function/index.py`** / **`patterns/unified/src/bda_processresults_function/index.py`** — replace `urlparse`-based S3 URI parsing (`create_metadata_file`, `parse_s3_path`) with `idp_common.utils.parse_s3_uri`. Non-S3 `urlparse` uses: none remained; the `urlparse` imports are dropped.

## Testing

- New resolver unit tests pin `#`-in-key round-trips for both the inline-contents and presigned-URL branches, plus a missing-key validation case (`nested/api-resolvers/src/lambda/get_file_contents_resolver/test_index.py` — 8/8 pass).
- `lib/idp_common_pkg` unit tests for `s3util`/`lambdas` pass; `ruff check` and `ruff format --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ SPDX-License-Identifier: MIT-0
 
 ### Fixed
 
+- **Documents with `#` in their name no longer break View Data / View Page Text and results metadata.** `urllib.parse.urlparse` treats `#` in an S3 URI as a URL-fragment delimiter, so keys like `Report_#2.pdf/pages/1/result.json` were silently truncated at the `#`, causing `NoSuchKey` errors ("No response from getFileContents" in the UI) and wrongly-named `.metadata.json` files. The `getFileContents`/`getFilePresignedUrl` resolver and the two processresults functions now parse S3 URIs with a plain string split (matching `idp_common.utils.parse_s3_uri`), preserving `#` in keys.
+
 - **Claude Sonnet 5 now appears in the UI model picker.** Sonnet 5 (`claude-sonnet-5` and the `:1m` extended-context variant) was present in pricing, model limits, and the Bedrock client routing — and is the system default extraction model — but was missing from the CloudFormation model enums in `patterns/unified/template.yaml` that feed the config schema, so it never showed in the config UI's model dropdown. Both variants are now registered across all three inference-profile prefixes (`us.`/`eu.`/`global.`). (#544)
 
 - **Addressed package vulnerabilities flagged by recent dependency scans.** Upgraded affected direct and transitive dependencies across the web UI (`src/ui`), the docs site, and the multi-doc-discovery and OCR-benchmark Lambdas to their patched versions.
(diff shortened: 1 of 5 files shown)


rstrahan (approved):
Reviewed — approving.

**Summary:** Replaces `urllib.parse.urlparse`-based S3 URI parsing with a plain string split (inline in the `getFileContents`/`getFilePresignedUrl` resolver, via `idp_common.utils.parse_s3_uri` in the two processresults functions), so object keys containing `#` (e.g. `Report_#2.pdf/pages/1/result.json`) are no longer truncated at the fragment marker. Fixes the *View Data / View Page Text* `NoSuchKey` failures in the UI and wrongly-named `.metadata.json` files.

**Verdict:**
- Good PR — tightly scoped, clear problem/why, `fix:` convention, targets `develop`, CI green.
- Safe / no regressions — validation semantics preserved (scheme check, empty-key check, bucket allow-list all unchanged); new unit tests pin `#`-in-key round-trips for both resolver branches plus a missing-key case; extends the already-established library-level `parse_s3_uri` pattern.
- Security — bucket allow-list guard untouched; no new creds/IAM/deps.
- Documented — CHANGELOG `[Unreleased] › Fixed` entry and clear inline comments explaining *why*.

Two non-blocking observations: the resolver inlines its own `split('/', 1)` (justified — no `idp_common` dependency in that Lambda) which changes the bucket-only-URI error message to "expected s3://<bucket>/<key>" (new test matches); and `parse_s3_path` now fails fast with `ValueError` on malformed URIs instead of returning garbage, which is a strict improvement for its trusted BDA-output callers.

Approve — safe to merge.

PR #566, opened 2026-07-24
fix(feature-platform): content-hash feature UI bundle paths so immutable caching survives same-version republish

rstrahan:
## Problem

Feature ui-deployers copy `ui-bundle.js` into the Web UI bucket at a **version-only** path (`features/<id>/v<ver>/ui-bundle.js`) with `Cache-Control: public,max-age=31536000,immutable`. Republishing the same feature version (dev iterations, or a reinstall after resetting a feature version for release) reuses the identical URL — so browsers holding the old immutable copy serve the stale bundle for up to a year without ever revalidating. Symptom: the old extension UI page renders in a normal browser profile while an incognito tab shows the newly deployed one. There is also no CloudFront invalidation on feature install (the `/*` invalidation only runs for main-UI CodeBuild deploys), and reinstalling the extension doesn't help because nothing about the URL changes.

## Fix

The `immutable` header is only safe if the URL changes whenever the content changes, so the destination path now embeds a content hash of the bundle bytes:

```
features/<id>/v<ver>-<sha8>/ui-bundle.js     (sha8 = first 8 hex chars of sha256)
```

- The hashed dir is what gets registered as `uiBundlePath`, so `FeatureLoader` picks up the new URL from the registry with **no UI changes**.
- On **Delete**, cleanup lists the feature's whole `features/<id>/` prefix, removing every published copy: current, superseded hashed dirs, and keys from the old non-hashed layout. Superseded copies are deliberately kept on **Update** so SPA sessions opened before the update keep loading until refreshed.
- The ui-deployer roles gain `s3:ListBucket` on the Web UI bucket, scoped with an `s3:prefix` condition to `features/<id>/*` (writes remain restricted to the feature's own prefix as before).

Applies to all five bundled feature ui-deployers: `pii-anonymizer`, `idp-data-generator`, `sample-feature`, `sample-health-insurance-review`, and `feature-template` (so third-party features scaffolded from the template inherit the fix).

## Testing

- New `test_bundle_cache_busting.py` in all four registered ui-deployer test roots: hashed-key copy + immutable header preserved, same-version-different-bytes yields a different path, Delete removes all copies under the feature prefix (including old-layout keys), and cleanup failures never block stack delete.
- All 11 feature-platform test suites pass (191 tests).
- `cfn-lint` clean on the five modified templates; `ruff check` / `ruff format` clean.

## Notes for reviewers

- `registerFeature` runs on every Create/Update and overwrites `uiBundlePath`, so the registry row always points at the latest hashed dir.
- The `uiBundlePath` API field is a plain string (no format validation), so the `v<ver>-<sha8>` shape needs no schema change.
- Existing installs migrate on their next feature stack update: the deployer writes the new hashed key and registers it; the old `v<ver>/ui-bundle.js` key remains until uninstall (harmless, and still cleaned up by the prefix-wide delete).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ SPDX-License-Identifier: MIT-0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Extension UI pages no longer serve a stale bundle from the browser cache after a same-version republish.** Feature ui-deployers copied `ui-bundle.js` into the Web UI bucket at a version-only path (`features/<id>/v<ver>/`) with `Cache-Control: public,max-age=31536000,immutable` — so republishing the same feature version (dev iterations, reinstall after a rollback) left browsers serving the old bundle for up to a year without revalidating, since the URL never changed. The destination path now embeds a content hash of the bundle bytes (`features/<id>/v<ver>-<sha8>/`), so any byte change yields a new URL and immutable caching stays safe; uninstall cleans up every published copy (current, superseded, and old-layout keys) via a new prefix-scoped `s3:ListBucket` grant. Applies to all five bundled feature ui-deployers (pii-anonymizer, idp-data-generator, sample-feature, sample-health-insurance-review, feature-template).
+
 ## [0.6.2]
 
 ### Added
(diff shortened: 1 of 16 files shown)


rstrahan:
closed in v0.6.2

PR #567, opened 2026-07-26
fix(mcp): get_results by document_id, import-robust tool handler, schema re-sync on stack update

fhuthmacher:
## Problem

An external MCP consumer (a downstream AgentCore gateway fed by IDP post-processing hook events) hit two defects in the IDPTools MCP target:

1. **`get_results` is batch-only.** The gateway tool schema hard-requires `batch_id` (`ValidationException: Missing required field(s): 'batch_id'`), but post-processing hook events carry only a document reference (e.g. `s3://<output-bucket>/Borrowing_Notice_#3.pdf/`) — never a batch id. Downstream integrations have no way to retrieve a processed document's results. Results actually live at `<doc_id>/sections/<n>/result.json` in the output bucket; the batch id is only used to enumerate documents, so a document-level lookup requires no batch context at all.
2. **A missing agents dependency disables every tool.** The MCP handler imports the analytics agent stack (→ `strands`) at module load through two chains (`index.py`'s `configure_logging` import and `tools/__init__` → `search.py` → `idp_common.agents.analytics.agent`). If the deployed agents Lambda layer is missing or incomplete, the whole handler fails with `Runtime.ImportModuleError: No module named 'strands'` — all five tools die, not just `search`.

While fixing (1), a third gap surfaced: **the tool schema is only sent at gateway-target CREATE time.** On stack updates, `update_gateway_target_if_needed` refreshed the Lambda ARN but deliberately preserved the existing `toolSchema` — so schema changes in the code never reach already-deployed gateways.

## Changes

**`get_results` by document id**
- `idp_sdk`: new `client.batch.get_document_results(document_id=...)`; the per-document retrieval logic is extracted from `get_results` into a shared `_get_document_metadata` helper (no behavior change for the batch path). Accepts a bare object key or an `s3://` output-prefix URI, parsed `#`-safely (string split, not `urlparse`).
- MCP `GetResultsTool`: accepts `document_id` as an alternative to `batch_id`; response keeps the same shape (`documents` array). Passing neither returns a structured `missing_identifier` error.
- Gateway target schema: adds `document_id`, drops `required: ["batch_id"]`, updated descriptions.

**Import robustness**
- `tools/__init__` imports tool modules lazily inside `get_tool`, so a dependency problem in one tool surfaces as a per-call error for that tool only.
- `search.py` imports its agent stack inside `execute()` and returns a structured error naming the missing module on `ImportError`.
- `index.py` guards the `configure_logging` import with a stdlib-logging fallback (that import transitively requires `strands` via the `agents.common` package `__init__`).
- Batch tools (`process`/`status`/`reprocess`/`get_results`) now work even when the agents layer is broken; `search` fails loudly and diagnosably.

**Schema re-sync on stack update**
- The IDPTools schema now lives in one module constant (`IDP_TOOLS_SCHEMA`), used at target create and re-applied by `update_gateway_target_if_needed` when the deployed schema differs (only for the IDPTools target; other targets untouched).
- New `ToolSchemaVersion` property on `Custom::AgentCoreGateway` in `template.yaml` — custom resources only re-run when a property changes, so this must be bumped with schema changes to push them to existing stacks (documented inline).

## Docs

- `docs/mcp-server.md`: documents the two `get_results` addressing modes and the batch-mode-only pagination params.
- `CHANGELOG.md` Fixed entry.

## Testing

- New `lib/idp_sdk/tests/unit/test_get_document_results.py` (7 tests): status/fields/confidence extraction, `#`-in-key preserved in the S3 results key, `s3://` reference normalization, no S3 read for non-completed documents.
- New `lib/idp_common_pkg/tests/unit/lambdas/test_agentcore_mcp_handler_tools.py` (6 tests): handler module imports and batch tools construct with `strands` blocked; `search` returns a structured dependency error; `get_results` routes `document_id` / `batch_id` correctly and rejects neither.
- Full `tests/unit/lambdas/` suite (31) and `idp_sdk` batch tests pass; `ruff check` / `ruff format --check` clean. (8 pre-existing `test_config_operations.py` failures on clean develop in my local env are unrelated.)


webarch-ai:
LGTM, Reviewed the changes, approach is backward compatible. The document id path, import updates, and the schema re-sync on update are good improvements. Thanks

rstrahan:
## Review — merged ✅

Thanks Felix, this is a well-scoped fix for three genuine defects. Merged to `develop` as 6df995fb0.

### What I verified

- Ran both new test files plus the full `lib/idp_common_pkg/tests/unit/lambdas/` (47) and `lib/idp_sdk/tests/unit/` (222) suites on the rebased tree — all pass. The `strands`-blocking tests genuinely exercise the failure mode (they evict `idp_common.agents` and poison `sys.modules["strands"]`), so they'd catch a regression back to a module-level import.
- Repo-wide `ruff check` and `ruff format --check` clean (what CI's `lint-cicd` runs).
- `_get_document_metadata` is a faithful extraction of the old per-document loop body in `get_results` — status precedence, confidence extraction, and the ERROR fallback are identical, so no behavior change on the batch path.
- The `#`-safety claim holds: `_document_id_from_ref` uses a string split, and the test pins `Report_#2.pdf` surviving intact into the S3 results key. Consistent with the `parse_s3_uri` fix elsewhere in the codebase.
- The schema re-sync only touches the target named `IDPTools`, and `credentialProviderConfigurations` is still passed through as `update_gateway_target` requires.
- No new IAM, no hardcoded partitions/endpoints, no secrets — the `document_id` read path reuses the same role and output-bucket read the batch path already had.

### Merge notes

Conflicts with `develop` (`CHANGELOG.md`, `template.yaml`) were resolved on merge — both trivial, an entry insertion and the new `ToolSchemaVersion` property. The `git-defender` hook on my side blocked pushing a rebase to your fork, so this went in as a merge commit of your original head; the resulting tree is byte-identical to the rebase I validated above.

### Non-blocking follow-ups (no action needed for this PR)

1. **`ToolSchemaVersion` is redundant.** `SourceCodeHash` on the same resource is already a checksum of `./src/lambda/agentcore_gateway_manager` (see `lib/idp_sdk/idp_sdk/_core/publish.py:2448`), so editing `IDP_TOOLS_SCHEMA` in `index.py` already changes a property and re-runs the custom resource. Harmless belt-and-braces, but the inline comment ("this must be bumped with schema changes") overstates it — future schema edits propagate even if nobody bumps it. Worth softening the comment so a later contributor doesn't think they've broken something by forgetting.
2. **Both args → batch path.** Docs say "exactly one of `batch_id` or `document_id`", but passing both silently takes the batch path (`if document_id and not batch_id`). Reasonable precedence; either document it or return a structured error for consistency with the `missing_identifier` case.
3. **Unknown `document_id` returns `success: true`.** A typo'd id yields `status: "UNKNOWN"` with null fields rather than a not-found signal. Matches the batch path's existing semantics, so this is consistency-preserving — but for a consumer following up on one specific document, an explicit not-found would be more diagnosable.

PR #576, opened 2026-07-28
fix(ci): container build resilience, unwedge rollbacks, and root-cause pinpointing in deploy summaries

rstrahan:
## Summary

The integration pipeline failed after 26 minutes with `ROLLBACK_FAILED`. The deployment summary reported *"CodeBuild DockerBuildRun build failed"* and recommended `aws cloudformation continue-update-rollback`.

That summary was accurate as far as it went, but it stopped one layer short of the root cause and its recovery command was invalid for the stack's actual state. Investigating turned up **four distinct defects**, one of which had been silently discarding merged security patches.

### Root cause of the failure: a transient PyPI connection drop

From the CodeBuild log the summary pointed at but never read:

```
#8 6.437 Downloading numpy-1.26.4-...whl (18.0 MB)
#8 6.458    ━━━━╸  2.2/18.0 MB 140.1 MB/s eta 0:00:01
#8 6.653 ERROR: Exception:
   BrokenPipeError: [Errno 32] Broken pipe
ERROR: process "/bin/sh -c pip install ..." did not complete successfully: exit code: 2
```

The connection dropped 2.2 MB into an 18 MB wheel, 6 seconds in. Nothing to do with the code under test. It then cascaded unrecoverably: `pip` → `docker build` → `docker push` → CodeBuild FAILED → custom resource FAILED → nested stack CREATE_FAILED → parent rollback.

---

### 1. Dependabot security bumps never reached the container image 🔴

The buildspec heredoc'd its **own inline copies** of `requirements.txt` and the `Dockerfile`, so the checked-in `nested/multi-doc-discovery/{requirements.txt,Dockerfile}` were dead files nothing consumed. The two forked silently:

| | Pillow pin |
|---|---|
| on-disk `requirements.txt` (Dependabot #530) | `12.3.0` |
| inline buildspec copy | `12.1.1` |
| **what CI actually installed** | **`12.1.1`** |

A merged CVE patch that never shipped — and wouldn't have, indefinitely. Compounding it, neither file was in the component's smart-rebuild dependency list, so a bump didn't even invalidate the checksum and `publish` would skip the component entirely.

The buildspec now builds `-f nested/multi-doc-discovery/Dockerfile` from the source zip (which `publish.py` now includes, failing the publish loudly if either input is missing), both files are rebuild triggers, and tests assert the buildspec contains no inline dependency pins so the fork cannot quietly return.

### 2. No resilience to transient build faults

`pip install` now uses `--retries 10 --timeout 60` inside a bounded 3-attempt loop (pip's own retries don't cover stream-level faults like a mid-wheel `BrokenPipeError`), `docker push` retries up to 3×, and the `DockerBuildRun` custom resource retries a FAILED/FAULT build **once** — within its Lambda budget, always reserving time to answer CloudFormation, and never retrying a deliberately STOPPED or TIMED_OUT build. Its failure message now names the build and project so a reader knows which log stream holds the real cause.

### 3. The rollback itself failed, wedging the stack 🔴

Independent of the build failure. S3 permits only **one** conditional bucket-config write per bucket at a time, but CloudFormation deleted `TestSetBucketPolicy`, `TestSetBucketAutoDelete` and `TestSetBucketNotificationConfiguration` **in the same second** (all at `21:33:32`), so `PutBucketNotificationConfiguration` returned `OperationAborted`. The handler had no retry and failed hard, turning a recoverable rollback into `ROLLBACK_FAILED` — a state needing manual intervention — even though the identical call **succeeded 3 minutes later** at `21:36:49`.

This is a pure race that would bite any rollback or delete, including on a perfectly healthy branch. Three fixes:
- Bounded backoff on transient S3 error codes (ladder sized to fit inside the Lambda timeout, which is raised accordingly — asserted by a test).
- A **Delete** now always reports success, so notification cleanup can never block a teardown. The bucket is `DeletionPolicy: Retain` anyway; this mirrors what `TestSetBucketAutoDeleteFunction` already did correctly.
- `DependsOn: TestSetBucketPolicy` serializes the two conditional writes, removing the race at its source.

### 4. The summary couldn't see the root cause, and gave invalid advice

`generate_deployment_summary` only ever received CloudFormation events. For a CodeBuild custom-resource failure the CFN reason is just `"CodeBuild failed with status: FAILED"` — a dead end that **no amount of prompt tuning can fix**, because the evidence chain terminates at the CloudFormation boundary. (The grounding rules correctly stopped the model from guessing.)

The harness now resolves the CodeBuild project from the nested stack that reported the failure, finds its most recent non-successful build, and attaches that build's failing phase, phase error, log tail, and console URL — snapshotted before teardown, alongside the existing CF events. Two related corrections:

- **The recovery command is now derived in Python from the stack's real status.** `continue-update-rollback` applies *only* to `UPDATE_ROLLBACK_FAILED`; a CREATE rollback lands in `ROLLBACK_FAILED`, where that command just errors out. The old summary guessed, and guessed wrong.
- **Primary vs. secondary failures are now distinguished.** The failure that *caused* the rollback is reported separately from one that merely *blocked* it, rather than listing both as peers and sending readers after the wrong resource.

Also adds the scoped `codebuild:ListBuildsForProject` / `BatchGetBuilds` grant this needs — the `app-sdlc` role had no `codebuild:*` permissions, so the new calls would have silently `AccessDenied`. (Reading the build's log stream was already covered by the existing `logs:*` on `/aws/codebuild/*`.)

### Before / after, same input

Replaying the real incident's events + log tail through the new path:

> 📋 Status: idp-0727-211933 FAILED - pip install broke on network error during docker build
>
> 🔍 **Root Cause (primary):**
> • `"BrokenPipeError: [Errno 32] Broken pipe"` downloading numpy
> • Failed pip command: `pip install --no-cache-dir --target /opt/python -r /tmp/requirements.txt` (exit code 2)
> • Resource: DockerBuildRun (Custom::DockerBuild)
>
> ⚠️ **Also blocked rollback:**
> • TestSetBucketNotificationConfiguration DELETE_FAILED — `OperationAborted`
> • Secondary only; did not cause the deploy failure
>
> 💡 **Fix Commands:**
> • `aws cloudformation delete-stack --stack-name idp-0727-211933`
> • Failing build log: `https://console.aws.amazon.com/cloudwatch/...`

---

## Testing

**65 new unit tests** across four files (197 passing in the affected suites):
- `scripts/sdlc/tests/test_codebuild_failure_trail.py` — trail-following, recovery-command routing per stack state, the `OperationAborted` race, pre-teardown capture. Includes a regression guard that `continue-update-rollback` is never suggested for a CREATE rollback.
- `nested/multi-doc-discovery/docker_build_lambda/tests/` — retry-once, no-retry-on-STOPPED/TIMED_OUT, and budget safety (never die without answering CloudFormation).
- `lib/idp_sdk/tests/unit/test_testset_bucket_notification_handler.py` — extracts the inline handler from `template.yaml` and exercises it; asserts the retry ladder fits inside the configured Lambda `Timeout`.
- `lib/idp_sdk/tests/unit/test_multi_doc_discovery_source_zip.py` — zip contents, no-inline-pins, and rebuild triggers. Mutation-tested: reverting the fix fails these.

**Checks:** `make fastlint` clean (exit 0), `cfn-lint` 0 errors on all three touched templates, `validate_service_role_permissions.py` passes. `make typecheck` is unchanged from baseline — the 2 errors are pre-existing in `calculate_capacity/index.py` (verified by stashing).

**Live E2E** (fresh stack, `us-west-2`) — full create + delete lifecycle with **zero failed CloudFormation events**:
- ✅ `MULTIDOCDISCOVERYSTACK` CREATE_COMPLETE; CodeBuild build #1 SUCCEEDED (no retry needed)
- ✅ Real CI log confirms `Using checked-in Dockerfile and requirements.txt` → `Collecting Pillow==12.3.0`, and **zero** inline-heredoc lines
- ✅ All 4 container Lambdas `Active`/`Successful`; invoking one cold-starts through every heavy import and reaches the handler body
- ✅ `DependsOn` ordering observed: policy CREATE_COMPLETE at `00:25:49`, notification starts `00:26:22` — no longer same-instant
- ✅ **Stack delete took `TestSetBucketNotificationConfiguration` to DELETE_COMPLETE in 5s with no `OperationAborted`** — the exact resource that previously wedged the stack
- ✅ Stack reached DELETE_COMPLETE and is gone; no wedge, no manual intervention

Also verified the pip retry loop directly in Docker: recovers when a command fails twice then succeeds, and still exits non-zero when all attempts are exhausted.


rstrahan:
## Review findings 1 & 2 addressed (7ce4210b9)

Both were cases where the original fix stopped short **and** the accompanying test asserted the adjacent thing rather than the load-bearing one — which is exactly why CI stayed green.

### 1. 🔴 Dependabot fix didn't work on in-place stack updates

The reviewer was right, and this invalidated a headline claim. Adding `requirements.txt`/`Dockerfile` to `get_component_dependencies()` invalidates the *component* checksum, which re-SAM-builds the **template** — but the container rebuild is gated separately on `BuildHash`, the only meaningful property of the `DockerBuildRun` custom resource, and that hashed only `src/lambda/multi_doc_discovery`.

So a bump left `BuildHash` byte-identical → CloudFormation saw no delta → the custom resource was never re-invoked → CodeBuild never ran → ECR `:latest` kept the vulnerable dependency. **A fresh create always builds, which is why the original E2E validation could not have caught this.**

`BuildHash` now hashes the handler directory together with both build inputs. Verified by mutating `requirements.txt` on disk with a real Pillow bump:

```
BuildHash before: c229228cb183bd3c
BuildHash after:  150b8bb1a7de8a94   → CloudFormation sees a delta, rebuild triggers
```

### 2. 🔴 In-progress build reported as the failure

`get_codebuild_failure_details` selected the newest build that was `!= "SUCCEEDED"`, which also matches `IN_PROGRESS`. Since this same PR added a one-shot build retry, the newest build is frequently still running at capture time — and an in-progress build has no `FAILED`/`FAULT`/`TIMED_OUT` phase, so `phase_error` and `failed_phase` came out empty while the genuinely-failed build was skipped. Compounding it, the prompt labels that section authoritative, so it actively steered the model toward empty evidence. Now filters on a terminal-failure allowlist.

### Tests

Added `TestBuildHashCoversTheImageInputs` (parametrized per build input, plus a drift guard so the test's local reimplementation can't diverge from `publish.py`) and three in-progress-build cases including all four terminal statuses.

**Both fixes mutation-tested** — reverting either one fails the new tests. 430 passing across the affected suites; `make fastlint` clean.

CHANGELOG corrected: it had credited the rebuild-trigger fix to the dependency-list change alone.

### Deferred as follow-ups (findings 3 & 4)

- **#3** trail resolves by stack *name*, which CloudFormation rejects for deleted stacks. The pre-teardown snapshot covers the primary path; only the fallback degrades, and it degrades to the old behavior rather than breaking. Fix is to preserve the ARN in `get_cloudformation_logs`.
- **#4** 352s retry ladder vs 420s timeout leaves a thin margin once 11 contended S3 calls and VPC cold start are counted. Worth trimming the ladder (the incident's own evidence showed success at ~3 min) or raising the timeout.

Thanks for the review — finding 1 in particular was a claim I'd over-asserted on evidence that structurally couldn't test it.

PR #578, opened 2026-07-28
feat(hooks): let pipeline hooks return a modified Document

rstrahan:
## Summary

Post-step pipeline hooks were structurally **read-only**. A hook could rewrite the S3 objects a document pointed at, but could not change the `Document` itself — no relabelling a section, adding/dropping sections, correcting an extracted attribute, adjusting confidence alerts, or appending metering.

Two things enforced that:

1. **The dispatcher discarded the hook's response.** `pipeline_hooks_function/index.py` extracted only the `halt` flag; the rest was kept purely as an observability artifact. Chained hooks at the same point all received the *same* input document, so hook #2 could not see hook #1's work.
2. **The ASL wrote hook output to a side branch.** Every hook state used `ResultPath: $.HookResults.<point>`, then handed off to a step that read the document from the path it occupied *before* the hook ran. `postSummarization` used `ResultPath: null` — discarded outright.

This PR makes hooks able to inject business logic into the pipeline itself, while preserving the existing contract exactly.

## Approach

The dispatcher becomes a **pass-through transform**: it now *always* returns a top-level `document` (the inbound one verbatim unless a hook modified it), and one small `Apply<Point>HookDocument` Pass state copies that into the canonical path the next step already reads. No step Lambda changes.

A hook opts in by returning `{"updatedDocument": <doc>}` — deliberately not `"document"`, so a read-only hook echoing its input cannot accidentally start mutating the pipeline. Two accepted shapes: an **inline dict** (the dispatcher spills it to the working bucket) or a **compressed reference** the hook wrote itself (no size ceiling).

Chained hooks now **compose**: each sees the previous hook's output, in `order`.

### Guardrails

A violation is **refused** — the pre-hook document continues and the reason lands in `$.HookResults.<point>.Payload.results[].documentUpdateRejected`. It never fails the workflow.

| Rule | Why |
|---|---|
| `id`/`input_key`/`input_bucket`/`output_bucket` immutable | The tracking-table row and output S3 prefixes are keyed off them |
| `sections` must be a list of id strings | `ProcessSections`' `ItemsPath` iterates it directly |
| `config_version` preserved | It resolves hooks for the rest of the pipeline |
| Inline documents capped at 5 MB | Bounded by Lambda's 6 MB sync response limit |

`allowDocumentUpdate: false` on a hook entry pins it to observe-only.

### New `idp_common.hooks` helpers

`load_hook_document(event)` / `updated_document_result(doc)` make the round-trip two calls and avoid the trap of hand-building a `Document` (which silently drops `metering`, `errors`, `hitl_metadata`, `processing_issues`).

## A real defect this surfaced

`use_bda` and `bda_project_arn` ride on the document *payload* (injected by the queue processor from the resolved config) purely so the state machine can read them via JSONPath — but they are **not** `Document` model fields. A hook doing the natural load → mutate → return therefore dropped them, and losing `use_bda` failed the execution outright:

```
Invalid path '$.document.use_bda': The choice state's condition path references an invalid value.
```

The dispatcher now carries both forward, back-filling only keys the hook did not set (so a hook can still deliberately reroute a document). Found by E2E, not by unit tests — covered by four regression tests now.

## Backward compatibility

| Concern | Outcome |
|---|---|
| Hook returns arbitrary JSON, no `updatedDocument` | Document echoes through **byte-identical** |
| No hooks registered | Dispatcher echoes the inbound document; Apply state is a no-op copy |
| `$.HookResults.<point>.Payload.halt` consumers | Untouched; `document` is additive |
| Dispatcher/hook error | Existing `Catch` blocks **already bypass** the Apply state → pre-hook document, no restore logic |
| Stored config shape | No new required fields; `PipelineHook` is `extra="allow"` |
| Step Lambdas, final output shape, workflow tracker | **No changes** |

Cost: 6 new `Pass` states (~free) plus one `PutObject` per *mutating* hook only.

## Testing

**Unit** — 2564 pass. 20 new dispatcher tests (echo-through, inline compression, compressed-ref pass-through, identity rejection, malformed `sections`, oversized inline, S3-failure degradation, multi-hook threading, `allowDocumentUpdate` gate, control-field preservation) + 10 for the helpers.

**E2E on a live stack** (IDP1/us-west-2, v0.6.3.dev4) — a probe hook registered at **all six** points via a dedicated config version:

- **mutate** — all 6 markers reached the final document; **6/6 sections carried the `postClassification` relabel** (proving `ProcessSections` iterated the hook's document, not the pre-hook one); 6/6 carried the `postExtraction` attribute injection. `COMPLETED`.
- **reject_identity** — every hijack attempt refused at every point, identity intact, document still `COMPLETED` (refusal degrades to passive).
- **baseline, no hooks** — document unchanged, `COMPLETED`.

Also verified: `make lint-cicd` clean, `make check-arn-partitions` clean, basedpyright clean on changed files, and all 6 Apply states deployed with correct paths.

## Docs

Both tiers: `docs/feature-platform.md` (full contract + per-point propagation table), `docs/feature-platform-developer-guide.md` (a "writing a mutating hook" walkthrough), new `lib/idp_common_pkg/idp_common/hooks/README.md`, and `CHANGELOG.md`.

### Note for reviewers

`postExtraction` runs **inside** the `ProcessSections` Map, so its document is a *section* document — `Pipeline_ProcessResultsStep` rebuilds the final document from the `ClassificationResult` base and takes only `sections[0]` + `metering` from each Map result. Section-scoped mutations propagate; top-level/page-level ones made there do not. Documented in both tiers.


rstrahan:
## Self-review findings addressed (commit `eed5f77eb`)

A review pass over this branch turned up two guardrail gaps, both the **same class** as the `use_bda` defect the second commit already fixed: an unvalidated field on the document payload that the state machine consumes directly, where a malformed hook response fails the execution rather than degrading to the passive behavior.

### 1. A compressed reference's `s3_uri` was accepted for any bucket and key

This is a **key-injection vector**, not merely a cross-bucket read. `Document.decompress()` (`lib/idp_common_pkg/idp_common/models.py:920-923`) parses the URI but **discards its bucket**, reading the key against the consumer's own working bucket. So a hook returning `s3://attacker/x/evil.json` didn't cause a foreign read — instead the key `x/evil.json` was read from the *real* working bucket, escaping the `compressed_documents/` prefix the design assumes.

The identity check couldn't catch it: it compares the wrapper's `document_id` field, not the content at the URI. A hook could point the next step at any object in the working bucket — e.g. another document's `sections/N/result.json` — and have it deserialized as this document.

The URI must now name an object under `compressed_documents/` in the stack's own working bucket.

### 2. A compressed reference omitting `num_pages`/`status`/`sections` was accepted, then failed downstream

`ProcessSections`' `ItemsPath` (`$.ClassificationResult.document.sections`) and `BDA_CheckExistingData`'s Choice read those paths directly, so an absent one yields `States.Runtime` and kills the execution. The dispatcher's own inline path and `Document.compress()` both emit them, so this only bit a hand-rolled reference — but that's exactly the path the docs advertise for large documents. They're now back-filled from the inbound document, alongside the existing control fields.

### Also

Documented that hooks should make mutations **idempotent**, since the workflow's `Retry` on a hook dispatch re-invokes the hook — an append-style mutation (`classification += "-X"`) can apply twice where a set is safe.

### Verification

- 7 new tests, including one pinning `_compress_inline_document`'s output as satisfying `_validate_compressed_ref`, so the write and validate paths can't drift. Older tests used bare `s3://wb/x.json` placeholders and were rewritten to realistic prefixed URIs.
- 2571 unit tests pass; `ruff` + `basedpyright` clean.
- **E2E re-run on IDP1/us-west-2 after the change** — the mutation scenario still passes end to end (6/6 sections relabelled at `postClassification`, 6/6 attributes injected at `postExtraction`, all 6 markers in the final document), confirming the tightened checks don't affect the documented helper-based path.

Three findings were assessed and deliberately **not** changed:
- Path traversal via a document id in the S3 key — contained, since S3 keys are opaque strings with no path normalization, so `..` stays literal and the `compressed_documents/*` IAM grant still holds.
- `featureId` in the S3 key is unsanitized — same reasoning, and it's admin-controlled config.
- `docs/aws-services-and-roles.md` not updated — this adds a statement to an existing role rather than a new role or service.

PR #593, opened 2026-08-05
feat(extensions): repackage ConfBench test set as an on-demand Extension (supersedes #583)

rstrahan:
Supersedes #583. Repackages the ConfBench dataset deployer contributed there by @sujimart as an optional **Extension Feature** listed in the extensions catalog, deployable on demand.

*Issue #, if available:* n/a

## Why

#583 added ConfBench as a fifth always-on main-stack dataset deployer. Review of that PR raised three problems, all of which follow from the dataset's size rather than from the code:

| Concern (#583 review) | How this PR resolves it |
|---|---|
| 🔴 The ~14-chunk ingest had to finish inside CloudFormation's 1-hour custom-resource deadline; measured happy path was 60–90 min, and overrun meant a **stack rollback** | Ingest is an on-demand Step Functions job. Nothing waits on a CloudFormation response, so an overrun is a retryable job, not a failed stack. |
| 🔴 `FailedIdsSoFar` accumulated inside the async self-invoke payload and could exceed Lambda's 256 KB `Event` limit — silently ending the chain during exactly the systemic failure you most want reported | Failures append to per-variant S3 objects; state-machine payloads carry counts only. |
| 🟡 **32.71 GB** added to every deployment with no opt-out — ~42× the combined size of the other benchmark sets | Opt-in twice: install the extension, then pick a subset. |

Measured sizes for context:

| Dataset | Size |
|---|---|
| RealKIE-FCC-Verified | 0.08 GB |
| OmniAI-OCR-Benchmark | 0.39 GB |
| Fake-W2-Tax-Forms | 0.31 GB |
| **ConfBench** | **32.71 GB** |

## Description of changes

**New extension** — `feature-platform/confbench-testset/`, registered in `config_library/extensions-oss.yaml`. Uses only existing host exports, so **`template.yaml`, `patterns/unified/template.yaml` and `publish.py` are untouched**; the entire main-stack footprint of #583 is removed.

Installing creates only machinery (under a minute, downloads nothing). An admin then picks a tier or individual noise variants and starts a background job:

| Tier | Documents | Size |
|---|---|---|
| Clean baseline | 75 | 0.02 GB |
| Light noise | 253 | 0.31 GB |
| Representative spread | 525 | 4.22 GB |
| Full dataset | 1,346 | 32.71 GB |

Per-variant sizes are shown in the picker with a live total and approximate monthly storage cost before confirming. Subsetting is a real lever, not a token one — `original` is 0.02 GB while `custom15` alone is 7.12 GB.

**Ingest** — `Plan` (download the parquet row index once) → `Map` over variants at concurrency 4, streaming CDN→S3 via bounded 16 MB multipart → `Finalize` (count what actually landed). Sharding is **by bytes, not file count**: variant sizes span 0.02 GB/75 files to 7.12 GB/74 files, so file count was never the right unit. Each pass returns a resume offset. Partial failures still register the test set, with a report at `_confbench_jobs/<jobId>/report.json`.

**Host change — `TestSet.configVersion`** (separate commit). Test Studio inferred a test set's configuration purely by name, looking for a config version whose name *equals* the test set id. That convention can't serve extension test sets, since the Feature Platform names presets `<featureId>-v<version>`. Test sets can now declare their configuration; resolution order is declared → id-name convention → active version. Nothing changes for the managed benchmark sets, which don't set the field.

**Carried over from the #583 review** — findings 4, 6, 8, 9, 10: `document_id` bound before the `try` (it was the first statement *inside* it, so a malformed row raised `NameError` from the `except` clause itself), guarded `abort_multipart_upload`, `ruff format`, timezone-aware datetimes, no ground truth in the resume state. Also 404/403 are no longer retried like a transient 5xx. Finding 7 resolves by subtraction — no fifth auto-deployed dataset, so "four benchmark datasets" stays correct. Finding 5 (CHANGELOG) is addressed.

## Two things worth reviewer attention

**The dataset has 21 noise variants, not 18.** The upstream dataset card and #583 both say 18. Five variants are *partial* (`archetype2`, `archetype9`, `custom22`, `custom23`, `default` cover 20–46 of the 75 documents), which is how a per-document reading undercounts them. Corrected throughout; the picker labels partial variants, because comparing accuracy across variants with different document populations isn't like-for-like.

**A pre-existing bug this surfaced** (separate commit, independent of the extension). `deleteTestSets` listed objects with a single unpaginated `list_objects_v2` and passed the result to one `delete_objects` — both cap at 1,000 keys. For any larger test set the DynamoDB record was deleted while every object past the first page stayed in the bucket: invisible to the UI, still billed. **Already reachable today** — Fake-W2-Tax-Forms is ~4,000 objects. The regression test fails against the previous code.

## Testing

- **93 new unit tests** for the extension, targeting the #583 failure modes specifically (retry classification, malformed rows, the abort guard, byte-budget sharding, the resume loop, Admin gating, paginated delete).
- **2,611** existing `idp_common_pkg` unit tests pass.
- A network-gated test (`CONFBENCH_NETWORK_TESTS=1`) re-derives the variant size table from the live dataset, so an upstream re-publish fails a test rather than quietly showing a wrong cost estimate.
- Clean: `ruff check`, `ruff format --check`, `basedpyright`, `cfn-lint`, `bandit`, `make check-arn-partitions`, `make codegen` (committed), `sam build`, `sam validate`, `idp-feature-cli build`, UI `tsc --noEmit` + `eslint`.

**Not yet done:** the ingest path has not run against a live stack, so first-run validation against real HuggingFace CDN throughput is outstanding. Everything above is static/unit verification.

## Attribution

Dataset deployer, streaming multipart transfer, and ground-truth baseline format originate from #583 by @sujimart.


rstrahan:
## Live validation on IDP1 (us-west-2)

The "not yet validated against a live stack" caveat in the description is now resolved. Deployed and ran end to end.

**Setup:** rebased onto `origin/develop` (GitLab, a strict superset of GitHub's `develop`), published as `v0.6.3.dev6`, updated IDP1 via the immutable versioned template `idp-main_0.6.3.dev6.yaml`, then installed the extension with `idp-feature-cli deploy --from-code`. Stack reached `UPDATE_COMPLETE`; all 82 parameters carried over with none dropped or added.

**Ingest — light-noise tier, 253 documents across 4 variants:**

| Metric | Result |
|---|---|
| Job status | `COMPLETED` |
| Uploaded / skipped / failed | 253 / 0 / 0 |
| Wall clock | **77 s** (22:23:26 → 22:24:43) |
| Bytes transferred | 307,936,465 — **exactly** the catalog's predicted `expectedBytes` |
| Objects in bucket | 253 PDFs + 253 baselines |
| Per-variant counts | 75 `original`, 75 `archetype4`, 75 `archetype10`, 28 `archetype9` — exact match to the committed catalog |

The byte-for-byte match between predicted and actual transfer confirms the measured size table is accurate, which matters because it's what the cost estimate shown to the admin is built from.

**Verified live:**
- `GET /variants` → 200, 21 variants, correct tier sizes, `deployed` annotation
- Admin gate → `POST /ingest` as `Viewer` returns **403**; as `Admin` returns 202
- Config preset applied as **non-active** `Config#confbench-testset-v0.1.1`
- Baseline shape → `document_class: {type: Invoice}`, `page_indices: [0,1]`, all 7 ground-truth fields, 10 line items
- Row index cleaned up post-job; only `report.json` remains
- **`configVersion` end-to-end**: `getTestSets` returns `confbench-testset-v0.1.1` for `confbench-light` and `None` for `realkie-fcc-verified` — so the extension's config is preselected while managed sets still use the id-name convention
- **No main-stack footprint**: `grep -c confbench` on the published `idp-main_0.6.3.dev6.yaml` → **0**

Not exercised: the byte-budget multi-pass resume path, which only triggers on variants above 1.2 GB (the light tier's largest is 155 MB). The heavier tiers would cover it.

## One change from review feedback

Renamed the display to **"Test Set - ConfBench"** (`featureId` unchanged) so it groups with "Test Set Generator" in the Extensions nav and catalog, which sort alphabetically on `displayName`. Bumped to v0.1.1 and re-verified on IDP1.

rstrahan:
## Two issues found during live testing on IDP1, both fixed

### 1. "Update available: v0.1.0" shown to a stack running v0.1.1

A real pre-existing bug in the Feature Platform, not specific to this extension. `listInstalledFeatures` computed:

```python
"updateAvailable": bool(latest_version and latest_version != installed_version)
```

Inequality, so **any** difference lit the badge — including a catalog *behind* the installed version, which is a downgrade prompt.

This state is routine, not exceptional. `idp-feature-cli deploy --from-code` (the documented per-extension dev loop) installs a newer extension immediately, while `catalog.json` only refreshes on a host stack create/update — so the catalog is legitimately older for the entire window between the two, which on a shared dev stack can be days. Any extension author following the documented workflow hits this.

Now compares parsed SemVer and requires strictly newer. Two behaviours the old check also got wrong:

- **Numeric, not lexicographic** — `0.1.10` > `0.1.9`, which string comparison reverses.
- **SemVer §11 prerelease precedence** — `1.0.0` is an update over `1.0.0-rc1`, not the reverse.

Non-SemVer strings fall back to the old inequality and log a warning: better to over-report than to hide a real update behind a parser.

Verified against the **deployed** Lambda on IDP1:

| installed | latest | updateAvailable | |
|---|---|---|---|
| 0.1.1 | 0.1.0 | `False` | the reported bug |
| 0.1.0 | 0.1.1 | `True` | genuine update |
| 0.1.1 | 0.1.1 | `False` | same version |
| 0.1.9 | 0.1.10 | `True` | numeric, not lexicographic |
| 1.0.0 | 1.0.0-rc1 | `False` | prerelease is not an upgrade |

Four unit tests added; the two negative ones fail against the previous code.

### 2. Catalog still showed the old name

The rename reached the nav but not the **Browse catalog** page, because `catalog.json` is generated at publish time and IDP1's copy predated the rename. Not a separate code bug — `publish.py:1966` already sorts catalog entries by `displayName`, so republishing was the fix.

Live catalog on IDP1 now:

```
PII Anonymization                        v0.1.0
Sample: Document Status (feature add-on) v0.1.0
Sample: Health Insurance Review          v0.1.3
Test Set - ConfBench                     v0.1.1   <-- renamed, and adjacent
Test Set Generator                       v0.1.2
```

Both the nav and the catalog now group the two test-set extensions together, and every installed extension correctly reports `updateAvailable=False`.

PR #604, opened 2026-08-07
fix(hooks): paginate active-config-version scans (#599)

### LINKED ISSUE #599: Pipeline hooks silently stop firing as the ConfigurationTable grows (filtered Scan with Limit bounds items examined, not matched)
## Summary

`_resolve_active_version` in the pipeline-hooks dispatcher resolves the active configuration version with a **filtered `Scan` bounded by `Limit`**. DynamoDB applies `Limit` to items **examined**, not items **matched**, so once the active `Config#` row falls outside the first *N* items the scan happens to examine, the dispatcher silently falls back to `Config#default` — which has no `postHook` entries — and **every pipeline hook stops firing**.

The workflow still succeeds, the document still reaches `COMPLETED`, and the hook state in the state machine is entered and exited normally. The hook Lambda is simply never invoked and writes no log entries at all.

This is a **growth-triggered, silent failure**: it becomes more likely with every configuration version an admin saves and every feature installed, so it is most likely to strike a long-lived production stack.

## Affected code

**1. `patterns/unified/src/pipeline_hooks_function/index.py`** — [`_resolve_active_version`, line ~241 on `develop`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/patterns/unified/src/pipeline_hooks_function/index.py#L224-L249):

```python
resp = table.scan(
    FilterExpression="begins_with(Configuration, :p) AND IsActive = :t",
    ExpressionAttributeValues={":p": "Config#", ":t": True},
    ProjectionExpression="Configuration",
    Limit=10,                     # <-- bounds items EXAMINED, not matched
)
items = resp.get("Items") or []
if items:
    ...
return "default"                  # <-- silent fallback
```

**2. `feature-platform/main-stack-extensions/lambdas/register_feature_hooks/index.py`** — same function name, same pattern, [`Limit=1`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/feature-platform/main-stack-extensions/lambdas/register_feature_hooks/index.py#L98-L113). Strictly more fragile: it fails as soon as the *single* item examined is not the active row, i.e. almost always on a table with more than a handful of versions.

Neither call paginates, and neither distinguishes "found an active version that has no hooks" from "failed to find the active version at all" — which is why the fallback is invisible.

## Reproduction / evidence

Observed on a live stack (host v0.6.3) while end-to-end testing a feature that registers a `postRuleValidation` hook:

| | |
|---|---|
| `Config#` rows in `ConfigurationTable` | 35 |
| Position of the active row in scan order | **33** |
| Items the dispatcher examines | 10 |
| `ScannedCount` required to reach it | 39 |

Directly measurable:

```console
$ aws dynamodb scan --table-name <ConfigurationTable> \
    --filter-expression 'begins_with(Configuration, :p) AND IsActive = :t' \
    --expression-attribute-values '{":p":{"S":"Config#"},":t":{"BOOL":true}}' \
    --projection-expression Configuration --max-items 10
# Count (matched): 1   ScannedCount (examined): 39
```

**Nothing changed in the dispatcher or in the feature between the run that worked and the run that did not.** Hooks fired at 18:59 UTC and stopped at 20:20 UTC. The only difference was that installing a new version of a feature added a 35th config row, which reshuffled the scan's internal ordering and pushed the active row past position 10.

### How it presents

Dispatcher log (INFO — easy to miss, reads like a normal no-op):

```
No hooks registered for postRuleValidation in Config#default
```

State machine output for the hook state:

```json
{"hookPoint": "postRuleValidation", "invoked": 0, "halt": false, ...}
```

…while the active version genuinely does contain the hook:

```console
$ aws dynamodb scan --table-name <ConfigurationTable> \
    --filter-expression 'IsActive = :t' ... 
Config#claims-pack-v0.4.0   ACTIVE      # rule_validation.postHook -> [{featureId: claims-pack, arn: ...}]
```

## Impact

- **Any feature that injects business logic via a pipeline hook silently stops doing so.** For a claims/vertical pack this means documents are ingested and never decided, with every dashboard green.
- **Unpredictable across environments.** It depends on hash distribution, so the same software behaves differently on two stacks, and the failure can appear and disappear as unrelated config versions are created or deleted.
- **Hard to attribute.** From the feature's own data store, "the hook was never invoked" and "the hook decided nothing" are indistinguishable. The natural first assumption is that the feature is broken.

## Suggested fix

Page until a match is found (or drop `Limit` and take the first match) in both locations:

```python
def _resolve_active_version(table, pinned=None):
    if pinned:
        return pinned
    start_key = None
    while True:
        kwargs = {
            "FilterExpression": "begins_with(Configuration, :p) AND IsActive = :t",
            "ExpressionAttributeValues": {":p": "Config#", ":t": True},
            "ProjectionExpression": "Configuration",
        }
        if start_key:
            kwargs["ExclusiveStartKey"] = start_key
        resp = table.scan(**kwargs)
        for item in resp.get("Items") or []:
            key = item["Configuration"]
            return key.split("#", 1)[1] if "#" in key else None
        start_key = resp.get("LastEvaluatedKey")
        if not start_key:
            logger.warning(
                "No active Config# version found after a full scan; falling back "
                "to 'default'. Hooks registered in a feature's config version will "
                "NOT run."
            )
            return "default"
```

Two smaller improvements worth making at the same time, both aimed at the fact that this failed *silently* for over an hour:

1. **Log the fallback at WARNING, not INFO**, and distinguish the two cases. "No hooks registered for X in `Config#default`" is a normal no-op on a host with no features installed; *failing to find an active version and then falling back* is not. Only the second is suspicious, and today they log identically.
2. **Report the resolved version in the hook result.** `{"invoked": 0}` alone cannot distinguish "no hooks configured" from "resolved the wrong version". Adding `"configVersion": <resolved>` to the dispatcher's returned payload would have made this diagnosable from the state machine output at a glance, without needing to read dispatcher logs.

A regression test is straightforward: seed >10 `Config#` rows with the active one written last, and assert the dispatcher resolves it rather than `default`.

## Suggested detection for consumers

For anyone testing a feature against a deployed stack, asserting on the feature's own output is not sufficient — it cannot see this. Reading the dispatcher's own count out of the execution history does:

```python
# From the workflow execution history, at the hook state's exit:
payload = output["HookResults"]["postRuleValidation"]["Payload"]
assert payload.get("invoked", 0) > 0, "the dispatcher never invoked the hook"
```

## Note on the pattern

A filtered `Scan` with a `Limit` is the wrong shape whenever you need *matches* rather than a sample. It may be worth a lint rule or a shared helper: the same defect has appeared four times across the platform and one downstream consumer, with symptoms as varied as "configuration appears empty", "a list view shows fewer rows than exist", and "hooks stop firing". In each case the fix was to page.

rstrahan:
Fixes #599.

## The defect

Lambdas across the platform resolve rows out of DynamoDB with a **filtered `Scan`**. DynamoDB applies both `Limit` and the implicit 1MB page size to the items it **examines**, not the items that pass `FilterExpression` — so a single scan call finds nothing whenever the matching row happens to sort beyond that window.

The failure is **silent and growth-triggered**: it gets likelier with every row added, so it strikes long-lived production stacks hardest. Because it depends on hash distribution, two stacks running identical software behave differently, and the failure can appear and disappear as unrelated rows are created or deleted.

Observed live on a v0.6.3 stack (per the issue): 35 `Config#` rows, the active row at scan position **33**, behind a 10-item examine window. Every registered pipeline hook stopped being invoked — the workflow still succeeded, the document still reached `COMPLETED`, the hook state was entered and exited normally, and the hook Lambda wrote no log entries at all.

## 1. Active-config-version resolution (the reported bug, ×4)

Four Lambdas resolved the active configuration version this way; none paginated. All four now page to the first match and project only the key, so the examine window isn't spent reading whole config bodies.

| Site | Consequence of resolving the wrong version |
|---|---|
| `patterns/unified/src/pipeline_hooks_function` | **Every pipeline hook silently stops firing.** For a vertical pack, documents are ingested and never decided with every dashboard green. |
| `feature-platform/.../register_feature_hooks` | Was `Limit=1` — strictly more fragile. Hooks are written into a **non-active** version, so registration reports success and they never run. |
| `src/lambda/queue_sender` | The document is processed under the **default** configuration rather than the active one — different classes, prompts and models, no error anywhere. |
| `nested/api-resolvers/.../test_runner` | A test run's comparisons are scored against a configuration the documents were not processed under. |

**On why the dispatcher scans at all** (raised in review): the pin usually *is* present — `queue_processor` compresses the document before starting the execution and `Document.compress()` carries `config_version` in the wrapper, so the dispatcher's `pinned` fast path normally short-circuits the scan. But the pin is only ever *populated* when the uploader supplied one: `upload_resolver` sets `x-amz-meta-config-version` only when the UI passes an explicit version, and `Document.from_s3_event` leaves it `None` otherwise. For the ordinary drop-a-file-in-the-input-bucket path there is no pin, and resolution falls to `queue_sender` — itself one of the four broken sites. So the scan is a real, regularly-exercised path, and the two defects compounded: `queue_sender` failed to stamp a version, which pushed the dispatcher onto its own broken scan.

## 2. Pin the version once, at the entry point

Raised in review: if `queue_sender` now always resolves a version, does the dispatcher need its own scan? Tracing it, the pin is more robust than it first appears — every step Lambda round-trips `config_version`, and the HITL reprocess path (`complete_section_review.trigger_reprocessing`, which bypasses `queue_sender` entirely) survives because the tracking table persists and restores `ConfigVersion`.

But the pin was still *conditional*, so the answer was "usually, not always". `queue_processor` — the single chokepoint every execution passes through — now stamps `document.config_version` before compressing, via a new `ConfigurationManager.resolve_active_version()` (built on the already-paginating `list_config_versions()`). That collapses N independent resolutions into one recorded value, and guarantees the state machine's `use_bda` / `bda_project_arn` flags come from the *same* version as the rest of the pipeline.

**Deliberately not fail-fast**, which is what I'd originally floated. I checked what a fresh stack actually looks like first: `update_configuration` writes `Config#default` with **no `IsActive` attribute at all**, and the Create path's `activate_version()` call sits inside a log-only `except`. So "no active version" is a reachable, normal state — failing there would reject documents that process fine today. Instead: an existing pin is never overwritten (upload-time choice and HITL reprocess both still win), no active version resolves to `default`, and a DynamoDB failure degrades to the historical behavior rather than failing the document.

The dispatcher's scan is **kept** as a defensive fallback — documents queued by an older release, or pinned during a DynamoDB outage — and its module docstring now says that explicitly instead of implying it's a primary path. Reaching it is worth noticing, which is what the WARNING and the `configVersion` in the payload are for.

## 3. A lint check, so this stops recurring

The issue notes this defect has now appeared repeatedly with symptoms as unalike as "configuration appears empty", "a list view shows fewer rows than exist" and "hooks stop firing". New `make check-filtered-scans` (`scripts/check_filtered_scans.py`) makes it enforced rather than re-reviewed: an AST pass flags every `.scan(FilterExpression=...)` whose enclosing function never references `LastEvaluatedKey`, and reports whether the call is bounded by an explicit `Limit` or just the implicit page.

Deliberate bounded samples opt out with `# filtered-scan-ok: <reason>`. The marker requires prose after the colon, so silencing the check leaves a reviewable justification rather than a bare pragma.

It deliberately does **not** accept a boto3 paginator as evidence of paging — a paginator pages the call made *through* it, and says nothing about a sibling `table.scan(...)`. That exact trap was live in `user_management`, which pages Cognito while its DynamoDB scan read a single page; an earlier draft of the checker missed it, which is why the rule is now explicit and pinned by a test.

Wired into `make lint`, `make fastlint`, and `make lint-cicd`, and added to the DynamoDB section of `.claude/skills/code-review.md`.

## 4. Three more real bugs, found by that check

| Site | Consequence |
|---|---|
| `user_management.list_users` | The **admin user list was truncated** to the first page — fewer users shown than exist, no error, no indication it was partial. |
| `user_management.sync_cognito_users_to_dynamodb` | Worse: the "already in DynamoDB" email set was built from one page, so **every user beyond it looked absent and was re-created under a fresh `uuid4`** — duplicating the record on every sync. |
| `bda_blueprint_service.delete_project` | `BdaProject#` cleanup could **miss the very row it was deleting**, orphaning a tracking entry pointing at a deleted project. |

`backfill_gsi_attributes` is the one genuine bounded sample in the codebase — any `doc#` row answers "has the backfill run?" — so it is annotated with that reason rather than changed.

## 5. The test gate's own guard was failing closed

`scripts/run_all_tests.py` errors on any test directory in neither `RUN_ROOTS` nor `QUARANTINE`, so that new tests are never silently skipped. `feature-platform/confbench-testset/tests` (93 tests, added in #593) was in neither — so the guard meant to prevent silent skipping was itself **hard-failing the entire gate on a clean checkout**, and had been since #593 merged.

Registered and running. `test_planner.py` self-skips via `importorskip` unless `huggingface_hub` and `pyarrow` are installed (`ingest/planner.py` imports both at module scope, and they're heavy runtime pins rather than general test deps); the other three modules run unconditionally. The gate now covers **46 roots**, up from 45.

## Diagnosability

This hid for over an hour, so two changes target that directly:

1. **Failing to resolve an active version now logs at `WARNING`**, worded distinctly from the ordinary no-op of a host with no features installed. Previously both logged identically at INFO (`No hooks registered for X in Config#default`), so the suspicious case was indistinguishable from the benign one.
2. **The dispatcher reports `configVersion` in its no-hooks result too** (it already did on the success path). `{"invoked": 0}` alone cannot distinguish "the active version has no hooks here" from "we resolved the wrong version". The state machine reads only specific JSONPaths, so an added field is backward-compatible.

`docs/feature-platform-developer-guide.md` gains a **"Verifying a hook actually fired"** note: asserting on a feature's own output cannot detect this, since "the hook ran and decided nothing" and "the hook was never invoked" are indistinguishable from the feature's data store.

## Tests

**56 new cases across seven modules. Every behavioral one was verified to fail against the pre-fix code** — reverted each fix, confirmed red, restored.

- `test_pipeline_hooks_dispatcher.py` — 9 cases. A fake table reproduces examine-then-filter paging. Covers the reported scenario (active row at 33 of 35), the last-page boundary, early-exit efficiency, the WARNING on genuine no-active-row, scan failure, the pinned fast path still skipping the scan, and an end-to-end handler assertion that the hook in the late active version is invoked.
- `test_register_feature_hooks.py` *(new)* — 8 cases against **moto** with 40 seeded versions, so real service semantics are exercised rather than a fake. Asserts hooks land in the active row and *not* in `Config#default`.
- `test_active_config_version_resolution.py` *(new)* — 13 cases for `queue_sender` and `test_runner`, including that each projects only the key.
- `test_check_filtered_scans.py` *(new)* — 14 cases pinning the checker in **both** directions, including the cross-service-paginator trap, marker-without-reason, and a repo-wide clean assertion.
- `test_scan_pagination.py` *(new)* — 6 cases for `user_management`, including that sync writes **nothing** when every Cognito user already exists.
- `test_delete_project_scan_pagination.py` *(new)* — 4 cases for the BDA cleanup.
- `test_config_version_pin.py` *(new)* — 6 cases: unpinned gets stamped, an existing pin is never overwritten, no-active-version pins `default` rather than failing, resolution failure doesn't fail the document, routing flags read the pinned version.
- `test_configuration_manager.py` — 5 cases for `resolve_active_version`, including an active row beyond the first scan page.

## Validation

- `lib/idp_common_pkg` unit suite: **2694 passed**, 55 skipped
- `python scripts/run_all_tests.py`: **all 46 roots passed** (was failing outright before this PR)
- `make test-packages-cicd`: all suites pass
- `make fastlint`, `ruff check`, `ruff format --check`: clean
- `scripts/sdlc/typecheck_pr_changes.py`: 0 errors (2 pre-existing warnings in untouched lines)

No template, IAM, or config-schema changes. The fallback to `Config#default` is preserved for the genuine "nothing is active" case, so behavior is otherwise unchanged.


rstrahan:
## End-to-end validation on a live stack

Deployed this branch to the **IDP1** stack (us-west-2) and ran real documents through. The stack happened to be in *exactly* the state the issue describes, so the bug reproduced before the fix and is gone after it.

### The bug, reproduced live (pre-fix code paths, against the real table)

36 `Config#` rows; active = `Config#claims-pack-v0.4.0`:

```
=== PRE-FIX dispatcher (Limit=10):
   matched : 0  examined: 10
   result  : NONE -> falls back to Config#default (BUG)
=== PRE-FIX register_feature_hooks (Limit=1):
   result  : NONE -> writes hooks into Config#default (BUG)
=== POST-FIX (paginating):
   result  : Config#claims-pack-v0.4.0   (in 2 scan call(s))
```

And the consequence was real on this stack — the active version carries three hooks, the fallback carries none:

```
Config#claims-pack-v0.4.0:  postHooks = {extraction: [claims-pack], rule_validation: [claims-pack], summarization: [claims-pack]}
Config#default:             postHooks = NONE
```

### Run 1 — normal upload (no `config-version` metadata)

`queue_sender`'s now-paginating resolver stamped the right version, and **every hook fired**:

```
  preprocessing          invoked=0  configVersion=claims-pack-v0.4.0
  postOcr                invoked=0  configVersion=claims-pack-v0.4.0
  postClassification     invoked=0  configVersion=claims-pack-v0.4.0
  postExtraction         invoked=1  configVersion=claims-pack-v0.4.0   (x7, one per section)
  postRuleValidation     invoked=1  configVersion=claims-pack-v0.4.0
  postSummarization      invoked=1  configVersion=claims-pack-v0.4.0
```

Pre-fix these would all have been `invoked=0` against `Config#default`. Note `configVersion` is present on the `invoked=0` points too — that's the new diagnostic, and it's what makes "no hooks here" distinguishable from "wrong version resolved" at a glance.

All three hook Lambdas wrote logs during the run (20:18–20:21 UTC) against a baseline of 23 hours earlier, so the invocations are unambiguously from this test.

### Run 2 — the pin's own path

Run 1 didn't actually exercise the new `queue_processor` pin, because `queue_sender` had already set `config_version` (correct no-overwrite behavior — but it means the pin block was skipped). To test the pin itself I enqueued a message with `config_version: null`, which is what an older release or a failed `queue_sender` resolution produces:

```
[INFO] Pinned config version 'claims-pack-v0.4.0' for document e2e-599-unpinned-...pdf
```

That document reached `COMPLETED` with `ConfigVersion=claims-pack-v0.4.0` in the tracking table and identical hook behavior (`postExtraction` 7, `postRuleValidation` 1, `postSummarization` 1).

### Clean

- Both documents reached `COMPLETED`
- **No** ERROR / Traceback / WARNING in `queue_processor` or `queue_sender` for the whole window
- `Config#default` untouched; active version unchanged
- Test objects deleted

### One unrelated issue found

`python3 publish.py` fails building the `confbench-testset` extension: `PythonPipBuilder:ResolveDependencies - {pyarrow==23.0.1(wheel)}` (that function builds arm64 and the pin has no installable arm64 wheel). I confirmed this is **pre-existing** by reproducing `sam build` on base commit `eeb8aff94` in a clean worktree, and worked around it for this deploy by temporarily dropping the extension from `extensions-oss.yaml` (restored afterwards — no change in this PR). It will break CI publish too, so it likely wants its own ticket.

Issue #570, opened 2026-07-27
Sample: Health Insurance Review — prior-auth docs produce no rule-validation output and never appear in Claims dashboard


rstrahan:
## Bug

Processing `samples/rule-validation/Prior-Auth-*.pdf` with the `sample-health-insurance-review-v0.1.0` config (rule validation enabled) produces **no rule-validation output** in the document details UI, and the document **never appears in the Claims dashboard**.

## Root cause (two compounding defects)

**1. Filename regex mismatch (config preset).** Each of the 7 NCCI policy classes matches only on the document filename via `x-aws-idp-document-name-regex: "(?i).*(medicare|medicaid|pa_packet|prior_auth).*"` — note `prior_auth`/`pa_packet` with underscores. The shipped samples use hyphens (`Prior-Auth-123789456.pdf`), so no class matches. With multiple policy classes a regex match is required, so classification returns `NO_POLICY_MATCH` and rule validation is skipped (0 rules). Confirmed in the policy-classification Lambda log and the consolidated summary (`overall_status: NO_POLICY_MATCH`).

**2. Workflow bypasses the postRuleValidation hook on the skip path.** In `patterns/unified/statemachine/workflow.asl.json`, the rule-validation skip states — `SetSkippedRuleValidationResult` (no policy match) and `SetEmptyRuleValidationResult` (rule validation disabled) — both transition directly to `SummarizationStep`, bypassing `PostRuleValidationHook`. So even though the feature's `postRuleValidation` hook is registered, it never fires for a skipped document, and no claim row is written (not even to flag it as needing documentation). Only the full path (`ProcessRuleValidationSections → RuleValidationOrchestration → PostRuleValidationHook`) invokes the hook. Verified: the claim table was empty and the hook Lambda had zero invocations.

## Fix

1. Broaden the preset regex to accept `-`, `_`, or space: `prior[_ -]?auth`, `pa[_ -]?packet` (all 7 classes).
2. Route both skip states through `PostRuleValidationHook`. The dispatcher no-ops when no hook is registered; the feature's hook records a `NO_POLICY_MATCH` run as `INSUFFICIENT_DOCUMENTATION`.
3. Bump the feature to v0.1.2 so subscribed stacks re-apply the corrected preset.

Verified live: reprocessing `Prior-Auth-123789456.pdf` now matches all 7 policy classes, evaluates 14 rules, fires the hook, and records the claim.

[...]

PR #607, opened 2026-08-10
fix(policy-discovery): stop four silent-underdelivery paths (#600-#603)

### LINKED ISSUE #600: Policy Discovery: documented output shape (rule_classes / x-aws-idp-rule-type) is not what persists or what the runtime reads (policy_classes / x-aws-idp-policy-type) — rules silently never fire
## Summary

`RulesDiscovery`'s **documented output contract and its actual persistence contract disagree**, and the tutorial notebook documents the one that does not work. A user who follows `notebooks/examples/step_rules_discovery.ipynb` step 6 saves rules under a config key that nothing reads, with a discriminator no consumer looks at — and gets no error.

## The disagreement

| | Says / uses | Where |
|---|---|---|
| Module docstring | *"The extracted rules are stored in the `rule_classes` configuration field"* | [`rules_discovery.py#L11`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/discovery/rules_discovery.py#L11) |
| Class docstring | *"RulesDiscovery: policy document → rule definitions → `config.rule_classes`"* | [`rules_discovery.py#L52`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/discovery/rules_discovery.py#L52) |
| LLM output / default prompt | `x-aws-idp-rule-type` | `base-rule-discovery.yaml` |
| Notebook step 5 "verify compatibility" | asserts `x-aws-idp-rule-type` is present | `step_rules_discovery.ipynb` |
| Notebook step 6 "save" | writes `{"rule_classes": rules}`, and instructs: *"Load rule_classes.json into your IDP config as the `rule_classes` field"* | `step_rules_discovery.ipynb` |
| **What actually persists** | **`policy_classes`**, reshaped to **`x-aws-idp-policy-type`** | [`_save_rules_to_config`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/discovery/rules_discovery.py#L588-L688) |
| **What the runtime reads** | **`x-aws-idp-policy-type` only** | [`rule_validation/service.py#L117-L141`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/rule_validation/service.py#L117) |

`_save_rules_to_config` does the reshape correctly — it maps `x-aws-idp-rule-type` → `x-aws-idp-policy-type`, adds `$schema`/`$id`/`type: object`, and forces `rule_properties[*].type = "string"`. So **the S3/DynamoDB path works.** The problem is that this reshape is invisible to anyone using the documented local/manual path.

## Why it fails silently

Two independent silent failures, and they compound:

1. **The config key.** `rule_classes` was renamed to `policy_classes` in v0.5.9. `IDPConfig` has a migration shim ([`models.py#L2631-L2636`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/config/models.py#L2631)) — but note the `elif`:

   ```python
   if "rule_classes" in data and "policy_classes" not in data:
       data["policy_classes"] = data.pop("rule_classes")
   elif "rule_classes" in data:
       del data["rule_classes"]        # <-- silently DISCARDED
   ```

   If `policy_classes` already exists (the common case — any config with policy classes already configured), the user's `rule_classes` is **deleted without a warning**. `rule_classes` is also in the deprecated-keys set, so no "unknown field" error fires either.

2. **The discriminator.** Even when the key survives migration, the raw discovery output carries `x-aws-idp-rule-type` and no `$id`/`type`. `PolicyClassificationService` matches on `x-aws-idp-policy-type` exclusively, so every class is unmatched and rule validation is skipped. The Policy Schema UI also keys off `x-aws-idp-policy-type`, so the class does not render.

Net effect for a notebook user: discovery reports "✅ 47 rules extracted, all with descriptions", the config saves without error, and **no rule ever evaluates.** There is no log line anywhere saying why.

## Reproduction

```python
from idp_common.discovery import RulesDiscovery

d = RulesDiscovery(input_bucket="unused", input_prefix="unused", config=cfg, region=region)
result = d.discovery_rules_from_document_local("policy.pdf")

# Exactly what the notebook's step 6 instructs:
print(result["rules"][0].keys())
# -> dict_keys(['x-aws-idp-rule-type', 'description', 'rule_properties'])
#    No $id, no type, no x-aws-idp-policy-type.

# Saved as the notebook says, into a config that already has policy_classes:
#   -> `rule_classes` is silently dropped by the IDPConfig migration `elif`
#   -> rule validation never fires, with no error and no warning
```

Compare `discovery_rules_from_document()` (the S3 path), which routes through `_save_rules_to_config` and therefore *does* work. The two entry points documented side by side in the same notebook produce different, silently-incompatible results.

## Suggested fix

**Extract the reshape so both paths share it.** The logic in `_save_rules_to_config` lines 618-668 is currently inlined in the persistence method; lifting it out fixes the local path and removes the duplication:

```python
def to_policy_class(rule_class: dict, unique_type: str) -> dict:
    """Reshape a discovered rule class into the persisted policy_classes shape."""
    ...  # existing lines 641-668

def discovery_rules_from_document_local(self, file_path: str) -> Dict[str, Any]:
    rules = self._extract_rules(...)
    # Return the SHAPE THAT ACTUALLY WORKS, so a caller can persist it directly.
    return {"status": ..., "rules": [to_policy_class(rc, rc.get("x-aws-idp-rule-type")) for rc in rules]}
```

Then:

- **Correct the docstrings** (`rules_discovery.py` lines 11 and 52) to say `policy_classes`.
- **Correct the notebook**: step 6 should emit `{"policy_classes": [...]}` and its guidance should name `policy_classes`. Step 5's compatibility check should assert `x-aws-idp-policy-type` — i.e. assert the thing the runtime actually requires, which would have caught this.
- **Make the migration `elif` loud.** Discarding user-supplied config silently is the root of failure mode 1:

  ```python
  elif "rule_classes" in data:
      logger.warning(
          "Both 'rule_classes' (deprecated) and 'policy_classes' are present; "
          "DISCARDING 'rule_classes' (%d entries). Merge them into 'policy_classes'.",
          len(data.get("rule_classes") or []),
      )
      del data["rule_classes"]
  ```

- **Optionally, accept both discriminators at read time** in `PolicyClassificationService` (`x-aws-idp-policy-type or x-aws-idp-rule-type`), which would make hand-written and older configs work rather than silently not matching.

A regression test would be: run `discovery_rules_from_document_local`, save the result verbatim into an `IDPConfig` that already has a `policy_classes` entry, and assert `PolicyClassificationService` matches the new class. That fails today.

## Context

Found while assessing reuse of Policy Discovery from a downstream vertical pack that also writes `policy_classes`. Related: #599 (same family — a silent fallback in the feature-platform hook dispatcher). The common thread is worth noting: **the failure mode in both is a path that quietly does nothing rather than reporting that it could not do the thing.**

### LINKED ISSUE #601: Policy Discovery: a second discovered policy class silently disables rule validation for all of them (append-only save + no auto-generated regex + regex required for multiple classes)
## Summary

Policy Discovery saves every extracted policy class **without a document-matching regex**, and `PolicyClassificationService` requires one as soon as there is more than one policy class. So a second Policy Discovery run silently switches off rule validation for the first policy — and for everything else already configured.

`docs/policy-discovery.md` documents "No automatic regex generation" as a limitation and names the missing regex as *the most common cause* of rules not firing. This issue is about the **interaction with append-only saving**, which turns a documented manual step into a silent regression of previously-working configuration.

## The interaction

Two behaviours are individually reasonable and together are a trap:

1. **Discovery is append-only and never sets a regex.** [`_save_rules_to_config`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/discovery/rules_discovery.py#L588-L688) appends each run as a new class with a disambiguating suffix. The reshaped class contains `$schema`, `$id`, `type`, `x-aws-idp-policy-type`, `rule_properties` — and **no** `x-aws-idp-document-name-regex` or `x-aws-idp-page-content-regex`.

2. **Classification requires a regex once there are ≥2 classes.** [`classify_document`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/rule_validation/policy_classification.py#L138-L181):

   ```python
   if len(self.policy_classes) == 1:
       ...  # regex optional — single-class mode always matches
   # Multiple policy classes - regex is required
   if not self._has_any_regex_configured():
       logger.warning("Multiple policy classes but no regex configured. ...")
       return PolicyClassificationResult(matched_policy_types=[])   # <-- nothing matches
   ```

### The failure sequence

| Step | State | Rule validation |
|---|---|---|
| Run Policy Discovery on policy A | 1 class, no regex | ✅ **Works** — single-class mode makes the regex optional |
| Run Policy Discovery on policy B | 2 classes, neither has a regex | ❌ **Silently stops for BOTH** |

The user did nothing wrong, changed no configuration by hand, and got a warning in a Lambda log. The first policy's rules had been evaluating; now nothing does. Because step 1 works, the natural mental model is "discovery output is ready to use" — which is true exactly until it isn't.

`_has_any_regex_configured()` is checked **globally**, not per class, so this also means: adding one regex-less discovered class to an existing hand-configured setup that *did* have regexes is fine, but a config consisting only of discovered classes can never match once there are two of them.

## Impact

- **A working configuration regresses with no user error and no visible signal.** Uploading a second policy manual — the explicitly supported "Cross-Policy Coverage" use case in the module's own docs — is what breaks it.
- **The only signal is a WARNING in the `RuleValidationPolicyClassificationFunction` log.** Job status says success, the Policy Schema UI shows both classes, and documents complete normally with empty rule-validation output.
- **It affects the documented happy path.** The tutorial and UI flow both end at "extracted N rules, view in Configuration" with no prompt to add a regex.

## Suggested fixes

Any one of these closes the silent-regression hole; the first two are complementary.

**1. Propose a regex at discovery time (removes the manual step).** The LLM has already read the policy document and knows what it governs. Ask it for a document-matching pattern in the same call, and save it on the class:

```yaml
# base-rule-discovery.yaml, appended to the output format
"x-aws-idp-document-name-regex": "(?i).*(medicare|prior_auth|pa_packet).*"
```

Even a conservative suggestion the user then edits is better than an empty field, because it makes the field's existence discoverable. A cheap non-LLM fallback: derive it from the source filename stem, which is already used to build `$id`.

**2. Warn at save time, not only at classification time.** `_save_rules_to_config` knows exactly when it is creating the dangerous state — it can see the resulting class count:

```python
if len(existing_policy_classes) > 1 and not any(
    rc.get("x-aws-idp-document-name-regex") or rc.get("x-aws-idp-page-content-regex")
    for rc in existing_policy_classes
):
    logger.warning(
        "Config#%s now has %d policy classes and NONE has a document-matching regex. "
        "PolicyClassificationService requires a regex when multiple classes exist, so "
        "rule validation will NOT fire for ANY of them until one is added.",
        version, len(existing_policy_classes),
    )
```

…and surface the same text in the discovery job's `statusMessage`, so it reaches the UI rather than only CloudWatch. The job that *creates* the problem is the right place to report it.

**3. Surface it in the UI.** Flag a policy class with no regex in the Policy Schema tab (a badge or inline warning), and ideally on the discovery job completion page next to "View in Configuration". The condition is trivially checkable client-side.

**4. Reconsider the fail-closed default.** Returning "no matches" when multiple classes lack a regex is defensible (better than running every rule against every document). But it is inconsistent with single-class mode, which fails *open*. Whatever the choice, the transition from one class to two should not be silent.

## Suggested regression test

```python
def test_second_discovered_class_does_not_disable_the_first():
    # Simulates two Policy Discovery runs, neither adding a regex.
    cfg = IDPConfig(policy_classes=[discovered("policy_a"), discovered("policy_b")])
    result = PolicyClassificationService(cfg).classify_document(some_document)
    assert result.matched_policy_types, (
        "two regex-less discovered classes matched nothing — the second upload "
        "silently disabled rule validation for the first"
    )
```

That fails today.

## Context

Found while assessing reuse of Policy Discovery from a downstream vertical pack. Related: #600 (same module, output-shape mismatch) and #599 (feature-platform hook dispatcher). All three share a shape: **a path that quietly does nothing instead of reporting that it cannot do the thing.**

### LINKED ISSUE #602: Policy Discovery: no cross-policy rule deduplication — duplicated rules multiply per-document validation cost and can return contradictory answers
## Summary

When two policy documents state the same underlying rule, Policy Discovery produces it twice — once per policy class — and the Rule Validation stage then evaluates it twice, at full LLM cost, potentially with contradictory answers. `docs/policy-discovery.md` lists this as a known limitation ("No cross-policy deduplication"); this issue is to track it with the cost and correctness consequences made explicit.

## Why it is more than cosmetic

Duplication here is not just untidy output — it multiplies through the validation stage:

1. **Cost scales with the duplication, per document.** Rule Validation answers each rule's question per section via an LLM call (`ProcessRuleValidationSections` Map → `RuleValidationFunction`). A rule duplicated across 3 overlapping payer manuals is answered 3× for every incoming document, forever. For a 100-rule manual set with meaningful overlap this is a standing multiplier on per-document cost, not a one-time extraction cost.

2. **Duplicates can disagree.** The same underlying requirement extracted from two manuals will be *worded* differently (each is extracted verbatim from its own source, per the default prompt's "Extract rules VERBATIM"). Two differently-worded questions about the same requirement can get different `Pass`/`Fail` answers from the model. The consolidated summary then contains a self-contradiction with no indication that the two rules were ever the same rule — which is materially worse than a single answer, because a reviewer cannot tell which to trust.

3. **It compounds the append-only design.** Re-uploading the *same* document (iterating on prompts, per the docs' own "Best Practices") also appends. So duplication arrives both across policies and across runs, and the docs' remedy — "delete the duplicate in the Policy Schema UI" — requires the user to first notice it among possibly hundreds of rules.

## Suggested approaches

Ordered by cost to implement; the first is cheap and probably sufficient.

**1. Report overlap at save time.** No semantic matching needed to catch the common cases. In `_save_rules_to_config`, compare incoming rule *names* (the LLM is instructed to produce descriptive `snake_case` identifiers, which converge for the same concept) against existing classes and log/surface the collisions:

```python
existing_names = {
    name
    for rc in existing_policy_classes
    for name in (rc.get("rule_properties") or {})
}
incoming = {n for rc in rules for n in (rc.get("rule_properties") or {})}
if overlap := existing_names & incoming:
    logger.warning(
        "%d rule name(s) already present in other policy classes and will be "
        "evaluated more than once per document: %s",
        len(overlap), sorted(overlap)[:10],
    )
```

Put the same text in the job's `statusMessage` so it reaches the UI. This turns an invisible cost multiplier into a visible one, which is most of the value.

**2. Flag likely duplicates in the Policy Schema UI.** Exact name matches across classes are already detectable client-side; near-duplicates via embedding similarity would reuse `idp_common.discovery.embedding_service`, which already exists for classes clustering.

**3. Opt-in dedup at extraction.** A `discovery.rules.deduplicate_against_existing: true` setting that passes existing rule names/descriptions into the extraction prompt and asks the model to omit rules already covered. Keeps append-only semantics (nothing is deleted) while preventing the duplicate from being created. Should be opt-in: silently omitting a rule that genuinely differs in a subtle way would be worse than a duplicate.

**4. Shared-rule references.** A more invasive option: let a rule be defined once and referenced by several policy classes, so it is evaluated once per document and the answer is shared. Correct, but a schema change affecting the validator and the UI.

## Suggested test

```python
def test_overlap_across_policy_classes_is_reported():
    # policy A and policy B both state prior_auth_required
    save(rules_for("policy_a", ["prior_auth_required", "x"]))
    with caplog.at_level(logging.WARNING):
        save(rules_for("policy_b", ["prior_auth_required", "y"]))
    assert "prior_auth_required" in caplog.text, (
        "a rule duplicated across policies will be evaluated twice per document "
        "with no indication to the user"
    )
```

## Context

Found while assessing reuse of Policy Discovery from a downstream vertical pack. Related: #600, #601 (same module).

### LINKED ISSUE #603: Policy Discovery: no multi-document clustering, and output truncation can silently persist a PARTIAL ruleset as success
## Summary

Two related scale limits, both documented in `docs/policy-discovery.md` ("Single-document rule extraction only", "Token limits on very large manuals"). Filing to track them, and to report one consequence the docs understate: **on output truncation, extraction can silently return a partial ruleset rather than failing.** For a compliance-validation feature, silently dropping rules is worse than erroring.

## 1. Truncation can silently under-extract (the part worth fixing first)

The docs suggest raising `max_tokens` if the LLM "is truncating mid-JSON", implying truncation presents as a parse failure. It often will — but not always, and the path where it doesn't is silent.

[`_extract_rules`](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/lib/idp_common_pkg/idp_common/discovery/rules_discovery.py#L295-L313):

```python
content_text = bedrock.extract_text_from_response(response)
json_text = extract_json_from_text(content_text)   # tolerant: strips fencing/preamble
parsed = json.loads(json_text)
rules = self._normalize_rules_response(parsed)
is_valid, error_msg = self._validate_rules_response(rules)
if is_valid:
    logger.info(f"Successfully extracted {len(rules)} rule classes ...")
    return rules
```

`_validate_rules_response` / `_validate_rule_class` check **structure, not completeness**: a class needs `x-aws-idp-rule-type` and a non-empty `rule_properties` whose entries have descriptions. A response truncated after rule 60 of 140 that happens to close its braces — or that `extract_json_from_text` recovers to a valid prefix — satisfies every one of those checks.

Result: the job logs `Successfully extracted 1 rule classes`, the UI reports success, 60 rules are saved, and **80 rules from the policy manual are silently absent from the validation set**. Nothing in the config, the job status, or the logs indicates the ruleset is incomplete. For a compliance feature, a validation set that is quietly 43% of the policy is a worse outcome than a failed job.

Note also that the default `max_tokens` is `""` (use the model maximum), so the ceiling is model-dependent and not visible to the user, and the `stop_reason` from the Bedrock response is not inspected anywhere in this path.

### Suggested fix

Check the stop reason — it is the direct signal and it is already in the response:

```python
if response.get("stopReason") == "max_tokens":
    validation_feedback = (
        "Response was truncated at the output token limit; the rule list is "
        "incomplete."
    )
    logger.warning(
        "Rules extraction hit the output token limit on attempt %d — extracted "
        "%d rule(s) so far, but the ruleset is INCOMPLETE. Split the document or "
        "raise discovery.rules.max_tokens.",
        attempt + 1, sum(len(rc.get("rule_properties") or {}) for rc in rules),
    )
    # Do not treat as success; retry or fail rather than persisting a partial set.
```

…and surface that in the job's `statusMessage` so it reaches the UI rather than only CloudWatch. Persisting a knowingly-partial ruleset should require an explicit opt-in, not be the default.

## 2. No multi-document clustering

Each policy manual is extracted independently; there is no equivalent of [Multi-Document Collection Discovery](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/blob/develop/docs/discovery.md#multi-document-collection-discovery) for rules. Consequences:

- **Chapter-split manuals cannot be recombined.** The docs' own workaround for limit (1) is "split the document by chapter/section into multiple uploads" — but each upload then becomes a *separate policy class*, which (a) collides with the regex requirement once there are ≥2 classes (#601) and (b) fragments one logical policy across several classes that must each be regex-matched identically. The recommended workaround for the token limit therefore triggers two other limitations.
- **Amendments/riders cannot supersede a base policy.** A base manual plus an amendment that changes one threshold produce two independent classes containing contradictory rules, with no notion that the later one governs.

### Suggested direction

A `policy_group` (or reuse of the collection concept from classes discovery) so several uploads can accumulate into one policy class, with later uploads able to supersede same-named rules rather than only appending. That would make the chapter-split workaround safe and give amendments a natural representation.

At minimum, until then: it would help for the docs to note that the chapter-split workaround interacts with the multiple-class regex requirement, since following one piece of documented advice currently breaks a different documented behaviour.

## Suggested tests

```python
def test_truncated_response_is_not_reported_as_success(monkeypatch):
    monkeypatch.setattr(bedrock, "invoke_model", lambda **kw: truncated_response())
    assert RulesDiscovery(...)._extract_rules(...) is None, (
        "a truncated response yielded a structurally-valid partial ruleset and was "
        "persisted as if complete"
    )
```

## Context

Found while assessing reuse of Policy Discovery from a downstream vertical pack. Related: #600 (output-shape mismatch), #601 (regex silent regression), #602 (cross-policy duplication). Common thread across all four: **a path that quietly does less than asked instead of reporting that it could not do it.**

rstrahan:
Fixes #600, fixes #601, fixes #602, fixes #603.

Five defects in one module, all the same shape: **a path that quietly does less than asked instead of reporting that it could not do the thing.** Four were filed; the fifth was found while verifying #570's fix.

## The unfiled one: the shipped preset evaluated 2 of its 14 rules

`config_library/unified/rule-validation/config.yaml` is a CloudFormation-selectable preset. It defines 7 NCCI policy classes but carried a `Document Name Regex` on only the **first**, and rule validation evaluates only regex-matched classes — so the other 12 rules could never fire. The regex was also underscore-only, missing the hyphenated filenames real intake systems produce. Against its own bundled sample:

| | before | after |
|---|---|---|
| `medicare_respiratory_pa_packet.pdf` | 1/7 classes, **2/14 rules** | 7/7 classes, **14/14 rules** |
| `Prior-Auth-123789456.pdf` | 0/7 classes, 0/14 rules | 7/7 classes, 14/14 rules |
| `invoice.pdf` | 0 (correct) | 0 (correct) |

This is the `config_library` twin of the preset fixed in #570 — that fix was applied to the feature's copy and this one was left behind. New gate in `config_library/test_config_library.py` fails **any** preset with >1 policy class where a class has neither regex, so a class can't ship unreachable again.

## #603 — truncation persisted as success

Extraction validated *structure*, not *completeness*. A response cut off after rule 60 of 140 that happened to close its braces passed every check: job logged "Successfully extracted 1 rule classes", UI reported success, 80 rules silently absent from a **compliance** validation set.

Both the traditional and agentic paths now check the stop reason (`max_tokens`, plus the OpenAI Responses adapter's `max_output_tokens`/`incomplete` — reusing the pattern already in `assessment/service.py`), retry, then **fail** rather than persist a knowingly-partial ruleset. Verified the remediation text survives the S3 path's exception wrapper into the job's `FAILED` status.

## #600 — documented shape wasn't the working shape

`discovery_rules_from_document_local` returned raw LLM output keyed on `x-aws-idp-rule-type`; only `x-aws-idp-policy-type` is read. So the two entry points documented side by side in one notebook produced silently incompatible results. The reshape is now shared (`to_policy_class`) so they can't drift.

Compounding it, `rule_classes` was **deleted without a warning** when both keys were present, and being a known-deprecated key it skipped the unknown-field warning too. That now logs what it dropped.

Readers accept the legacy discriminator in **both** the classifier and the service's rule-question lookup — those had to agree, or a class would classify and then yield zero questions (the same silent no-op, one stage later).

## #601 — a second class disabled validation for all of them

Discovery sets no regex; validation requires one at ≥2 classes. So a second upload switched off the *first* policy's rules, with only a downstream Lambda WARNING as signal. The save that creates that state now warns, and it reaches the **job status message** — the job that creates the problem is the right place to report it.

Runtime stays **fail-closed** (failing open would run every rule against every document) and is now pinned by a test, so the warning and the behavior it describes can't drift apart.

## #602 — duplicate rules were invisible

Each copy is answered by its own LLM call for every document, forever — a standing per-document multiplier, not a one-time cost — and differently-worded copies can return contradictory verdicts. Saving now reports rule names already present in other classes. Name matching only; the prompt asks for descriptive `snake_case` identifiers that converge.

## Out of scope

Still documented as limitations: multi-document clustering / `policy_group`, LLM-proposed regexes, automatic dedup. Docs now spell out that the documented chapter-split workaround for the token limit **triggers** the multiple-class regex requirement — following one piece of documented advice broke another.

## One correction to #601 as filed

Its suggested cheap fallback — derive the regex from the source filename stem — would be wrong. That regex matches the *incoming transaction documents*, not the policy manual, so a stem derived from `NCCI Medicare Policy Manual.pdf` would match nothing. Warn-loudly and LLM-proposed regexes are the viable options.

## Validation

- **40 new/updated tests**, each verified to fail with its fix reverted (not just to pass with it).
- 2721 `idp_common` unit tests, 114 `config_library`, ruff, format, basedpyright all green. The 1 `models.py` basedpyright warning is pre-existing on `develop`.
- Not yet exercised against a live stack — the truncation path in particular is unit-tested only.


rstrahan:
Thanks — the agentic finding was the valuable one, and it was right about the mechanism. All four addressed in c5f47315a. One correction to the conclusion, though.

## The agentic guard: mechanism right, conclusion wrong

Confirmed `_response_was_truncated` cannot fire there. `structured_output()` builds its envelope with only `output` (`agentic_idp.py:1695`, `:2083`); I verified by constructing the real envelope and calling the guard — `False`, exactly as you found.

But **#603 was not unfixed for agentic**, because that path never had the silent-persist failure mode in the first place. Strands fails **hard** on `stopReason == "max_tokens"`:

```python
if stop_reason == "max_tokens":
    raise MaxTokensReachedException(...)
...
except (ContextWindowOverflowException, MaxTokensReachedException) as e:
    # Special cased exceptions which we want to bubble up rather than get wrapped
```

`invoke_agent_with_retry` is a bare pass-through and `agentic_idp.py:1421` is a bare `raise`, so it reaches the caller. Proven end-to-end — the job raises rather than persisting a partial ruleset:

```
$ patch structured_output -> MaxTokensReachedException
RAISES: Exception
MSG   : Failed to extract rules from p.pdf: max_tokens limit
```

So the real defect was **diagnosability, not silence**: the user got an opaque `"Agent has reached an unrecoverable state due to max_tokens limit"` with no indication of what to do. That's now translated into the same `RulesTruncatedError` the traditional path raises, so both paths name the same remedies. The detector matches on type name + message rather than importing the Strands type, mirroring the existing `_is_context_overflow_error` — this module has to stay importable without the agentic extras.

I kept the envelope check rather than dropping it, but took your point that unfired code reading as protection is worse than a documented gap — so it now says so explicitly:

> NOTE: on this path the primary protection is Strands raising MaxTokensReachedException (translated in the handler below) [...] This check is a belt-and-braces fallback: `structured_output()` currently synthesizes its envelope WITHOUT a stopReason, so it does not fire today, but it will start working if that envelope ever carries the real stop reason.

Docs and CHANGELOG now state the per-path mechanism instead of implying one shared guard.

**On "no tests over the agentic guard at all" — worth recording why.** `tests/conftest.py:28` stubs `strands` with a `MagicMock`, so `agentic_idp`'s import fails and `AGENTIC_AVAILABLE` is `False` for the entire suite. The agentic branch was *unreachable* from tests, not merely untested — which is also why `patch(...)` on `structured_output` fails with `does not have the attribute`. New tests force the flag and inject the symbol. 8 added, including one asserting the envelope guard works **if** `stopReason` is ever populated (otherwise it's decoration, not defence) and one asserting an unrelated error isn't relabelled as truncation. The translation test was verified to fail when reverted.

## Finding 3 — fixed, and you identified the more likely sequence

Confirmed exactly as described: policy A with a regex + new class B without one → no warning, and `classify_document` returns `['policy_a']` only. Warning solely on the all-or-nothing case left the follow-up to my *own new advice* uncaught. Now warns per regex-less class and names them, which matches the stricter rule the new `config_library` gate already enforces — so the two agree rather than the library gate being stricter than the runtime warning.

## Finding 2 — fixed, plus something else in there

Same left-behind-twin pattern, agreed. Fixed the display cells, output filenames and intro. While in there I also cleared a committed **failed-run traceback** that leaked an absolute developer path (`~/efs/EFSSpace/...`) — unrelated to this PR but not worth shipping.

## Finding 4 — fixed
Plain `"entry"`/`"entries"` branch.

## CI failure — agreed, and filed

Your diagnosis is right and worth more than "flake": it's deterministic given a small `time.monotonic()`, so it will recur on fresh containers rather than being random. Filed as #609. One addition — this is a latent bug in the **production** path, not just the test: back-dating `_cache_time` by subtracting from a clock whose origin may be more recent than the offset can't express "retry in 30s", and it's only harmless in Lambda because a warm container's uptime exceeds any sane TTL. Suggested fix there is an explicit `_retry_after` deadline rather than a `max(0.0, ...)` clamp, which would stop the negative value but silently retry immediately.

Validation: 2729 `idp_common` unit tests, 114 `config_library`, ruff, format, basedpyright all green.

PR #620, opened 2026-08-13
fix(test-studio): stop a completed test run reporting EVALUATING forever (#619)

### LINKED ISSUE #619: Test run stuck at EVALUATING with all files processed and 0 evaluating when the test set name contains spaces or parentheses
## Symptom

A completed test run shows a blue **EVALUATING** badge in Test Studio, but the progress detail underneath reads `10/10 processed` and `0 evaluating` — a state that reads as self-contradictory. It never clears; the badge stays EVALUATING indefinitely and the results page never renders metrics.

Observed on stack `IDP1` (us-west-2) for test run `ConfBench (light noise)-20260813-132501`.

## Trigger

The test set name contains characters outside the SQL *identifier* allow-list — here the spaces and parentheses in `ConfBench (light noise)`. Any test set name containing a space, `(`, `)`, `'`, etc. reproduces this.

## Root cause

The run genuinely finished. In the tracking table: `Status=COMPLETE`, `CompletedAt` set, all 10 documents at `ObjectStatus=COMPLETED` / `EvaluationStatus=COMPLETED`. The aggregation even computed good numbers — `overall_accuracy=0.7232`, `avg_weighted_score=0.6784`, `document_count=10`.

But `testRunResult` was never written to the metadata item, and `get_test_run_status` deliberately reports `EVALUATING` whenever the status is terminal but `testRunResult` is missing (`test_results_resolver/index.py`). That override is the badge; `evaluatingFiles` is correctly `0` because every document really is evaluated. Hence the apparent inconsistency.

The metrics were lost like this (log timestamps `13:28:20`):

1. Stickler aggregation **succeeded** — `Using Stickler aggregation for test run ConfBench (light noise)-...`
2. Immediately after, `_get_evaluation_metrics_from_athena()` called `_validate_sql_input()`, which enforces `^[a-zA-Z0-9_\-./]+$`. `test_run_id` is derived from the user-chosen test set name and sits in a **string-literal** context (`WHERE document_id LIKE '{test_run_id}%'`), not an identifier context — so applying an identifier allow-list to it rejected a perfectly valid run:
   ```
   ValueError: test_run_id contains invalid characters: ConfBench (light noise)-20260813-132501
   ```
3. That `ValueError` was caught by the broad `except` around the whole Stickler block, which logged the misleading `Test execution aggregation Lambda failed ... falling back to Athena` — **and discarded the good Stickler metrics**. The aggregation Lambda had not failed at all.
4. The Athena fallback called the same validator → raised again → propagated into `handle_cache_update_request`, whose `except` swallows it without re-raising. SQS deleted the message (the queue has no DLQ), so nothing retried.

## Why it never recovers

Nothing re-enqueues the aggregation, so the run is permanently latched:

- `get_test_run_status` only calls `_queue_cache_update` inside the `overall_status != stored_status` block. The stored status is already `COMPLETE`, so there is no transition and no re-queue.
- `get_test_results` only re-queues in the `cached_metrics is not None` stale-cache branch. Here `cached_metrics` is `None`, so it takes the else branch and queues nothing.

Confirmed live: at `13:41:30`, 13 minutes after the failure, a UI poll logged `aggregate metrics not yet available` and queued nothing.

## Contributing factors

- **Misleading log message.** `Test execution aggregation Lambda failed ... falling back to Athena` fires for a failure in the *optional Athena supplement*, after the aggregation Lambda succeeded. This actively misdirects troubleshooting.
- **Redundant duplicate work.** Three concurrent resolver invocations each detected the same status transition and each enqueued a cache update, so the aggregation Lambda ran 3× for one run.
- **No DLQ** on `*-test-result-cache-update`, so swallowed aggregation failures leave no trace outside the Lambda log.

## Blast radius

Of 49 test runs on `IDP1`, exactly one is stuck in COMPLETE-without-metrics — the one whose ID is not SQL-safe. The other 3 without metrics are legitimately `RUNNING`/`QUEUED`. Every run with a SQL-safe ID cached fine. So the bug is fully determined by the test set name.

## Expected behavior

1. A failure in the optional Athena supplement (split-classification metrics, cost) must not discard accuracy/confidence metrics that Stickler already computed successfully.
2. A test set name containing spaces/parens/quotes must aggregate normally — the correct defense for a string literal is escaping, not an identifier allow-list.
3. A terminal run with no cached metrics must retry, rather than reporting EVALUATING forever.

rstrahan:
Fixes #619.

## Symptom

A completed test run showed a blue **EVALUATING** badge while the progress detail underneath read `10/10 processed` / `0 evaluating`, and never cleared. Observed on stack `IDP1` for test run `ConfBench (light noise)-20260813-132501`.

The run had genuinely finished — every document `ObjectStatus=COMPLETED` / `EvaluationStatus=COMPLETED`, `Status=COMPLETE`, `CompletedAt` set — and the aggregation Lambda had already computed good numbers for it (`overall_accuracy=0.7232`, `document_count=10`). But `testRunResult` was never written to the metadata item, and `get_test_run_status` deliberately reports `EVALUATING` while a terminal run has no cached metrics. That override is the badge; `evaluatingFiles` was correctly `0`. Hence two halves of one display that looked contradictory.

## Root cause: identifier guard applied to a string literal

`test_run_id` derives from the user-chosen test set name and is interpolated into a string **literal**:

```sql
WHERE document_id LIKE '{test_run_id}%'
```

…but it was checked against the **identifier** allow-list `^[a-zA-Z0-9_\-./]+$`, so a perfectly legal name raised:

```
ValueError: test_run_id contains invalid characters: ConfBench (light noise)-20260813-132501
```

Any test set name containing a space, `(`, `)`, or `'` reproduces this.

The correct defense for a literal is escaping, not an allow-list. `test_run_id` now goes through `_sql_like_prefix()`, which doubles `'` — the complete escape for a Trino/Presto single-quoted literal, where backslash is *not* an escape character, so it cannot be bypassed — and neutralises the LIKE wildcards `%` and `_` via a paired `ESCAPE '\'` clause. That wildcard escaping fixes a second latent bug: a name containing `_` previously matched any character there, silently widening the prefix match to other runs. The database name **is** a genuine identifier and keeps the strict allow-list.

## Two failure-containment bugs fixed alongside

Either one alone would have prevented the permanent stall, so both are addressed:

**1. An optional supplement took down the whole aggregation.** The Athena split-classification and cost queries only *supplement* accuracy/confidence metrics that Stickler has already computed. But the `ValueError` escaped through the broad `except` wrapping the entire Stickler block — which discarded the good metrics and logged the actively misleading `Test execution aggregation Lambda failed ... falling back to Athena`, when that Lambda had succeeded — then propagated into `handle_cache_update_request`'s bare `except`, which swallowed it and let SQS delete the message. They are now fetched via `_athena_supplements()` and degrade to their documented "no data" shapes.

**2. Nothing ever re-enqueued the aggregation.** The enqueue in `get_test_run_status` fires only on a status *transition* (already `COMPLETE`, so no transition), and the stale-cache re-enqueue in `get_test_results` fires only when `testRunResult` already *exists* (it was `None`). Confirmed live: 13 minutes after the failure a UI poll logged `aggregate metrics not yet available` and queued nothing. Both entry points now re-enqueue when a terminal run has no metrics.

That re-enqueue is throttled by an atomic conditional write on a new `CacheUpdateQueuedAt` attribute, which serves two purposes: it caps recovery attempts at one per 5 minutes even though the UI polls `getTestRunStatus` every 5 seconds, and it collapses the redundant duplicate work seen in the incident, where three concurrent readers each fired their own aggregation for the same run. It fails **open** — a broken throttle must never block a legitimate recompute.

## Also

Corrects an overstated claim in the guard's own comment: it asserted "no comment markers (`--`)", but hyphens are allowed because real database names need them (`idp1-reporting-db`), so `--` *is* admitted. That is harmless here — the value only ever appears inside double quotes, where it is identifier text rather than the start of a comment — and the comment now says so. A test pins the guard's actual behavior.

## Deliberately not changed

`handle_cache_update_request` still swallows exceptions rather than re-raising. The queue has **no DLQ**, so re-raising would redeliver a poison message until the 14-day retention expired. Adding a DLQ with bounded retries is the better fix and is an infrastructure change worth doing separately; the view-triggered re-enqueue covers recovery in the meantime.

## Testing

13 tests added to `lib/idp_common_pkg/tests/unit/test_results_resolver.py`:

- the escaping helpers (quote doubling, injection payloads rendered inert, wildcard neutralisation, escape-char-first ordering, backslash left alone)
- the identifier guard still rejects injection, and still accepts hyphenated database names
- both Athena query builders accept the offending name and emit a correctly escaped query with the `ESCAPE` clause
- Stickler metrics survive both Athena supplements raising — the core regression
- the re-enqueue on both `get_test_results` and `get_test_run_status`, including the exact reported state (`EVALUATING` + 10/10 + 0 evaluating)
- convergence: no re-enqueue once metrics are cached, and none for `ABORTED`
- the throttle's claim / deny-on-condition-failure / fail-open paths

Results:
- `tests/unit/test_results_resolver.py` — 24 passed
- all five test files touching the changed functions — 105 passed
- full `idp_common_pkg` unit suite — **2763 passed, 55 skipped, 0 failed**
- `ruff check` clean on both changed files

Note: `ruff format --check` already fails repo-wide on `develop` (51 files, `index.py` among them), so I did not reformat it — that would bury this fix in unrelated churn. Verified `ruff format` would not touch any of the new code. `bandit` is not installed in this environment so the SRT scan was not run locally; the `# nosec B608` annotations are retained with updated justifications naming the new helpers.

## Follow-ups worth considering

- Add a DLQ to `*-test-result-cache-update` so swallowed aggregation failures are visible outside the Lambda log.
- Consider normalising test-run IDs at creation — the ID also lands in S3 prefixes and Athena `document_id` values, so a name with spaces/parens is awkward well beyond this query path.
- The existing run on `IDP1` needs a one-off backfill after deploy (send `{"testRunId": "ConfBench (light noise)-20260813-132501"}` to the cache-update queue); the fix does not retroactively populate it.


rstrahan:
## Self-review (author) — approve with non-blocking follow-ups

Posting my own review for the record. This **cannot substitute for an independent reviewer** — I wrote the patch.

**CI:** ✅ passing · **Target:** `develop` ✅ · **Size:** +598/−31 across 3 files (388 additions are tests)

### Live verification against the IDP1 reporting DB

The riskiest thing here was SQL that unit tests only assert as a *string*: the new `ESCAPE '\'` clause was never exercised against real Athena. That risk is compounded by this same PR making Athena failures silent — had the clause failed to parse, **every** run would have quietly lost split metrics and cost, with only a `logger.warning` as evidence. So I ran it live:

| Check | Result |
|---|---|
| `LIKE 'ConfBench (light noise)-...%' ESCAPE '\'` | ✅ SUCCEEDED, **10 rows** — the previously-broken run |
| Safe name (`OmniAI-OCR-Benchmark-...`) with ESCAPE | ✅ **293 rows** — no regression |
| `\_` (escaped underscore) | ✅ **0 rows** — matches literally, correct |
| `_` (unescaped) | ⚠️ **293 rows** — confirms the wildcard-widening bug was real, not theoretical |

That closes the main risk and independently confirms the secondary bug this PR fixes.

### Findings

- 🟡 **`index.py:729`** — The "terminal but no metrics → EVALUATING" rule now exists in **three** places (`_build_test_run_list:729`, `get_test_run_status:977`, `get_test_results`), and this PR adds self-heal to only two. `_build_test_run_list` backs `getTestRuns` (the Executions list), so recovery there depends incidentally on `TestRunnerStatus` polling `getTestRunStatus` per row rather than on the list query itself. It works, but a triplicated rule with divergent behavior is how the next bug of this class gets written. Should be one shared helper.

- 🟡 **`index.py:229-247`** — `_claim_cache_update_slot` issues a conditional `update_item` on a path the UI polls every 5 seconds per stuck run. `get_test_run_status` already holds `item`, which carries `CacheUpdateQueuedAt` after the first write, so an in-memory pre-check would remove a DynamoDB write from a hot path (failed conditional checks still consume WCU).

- 🟢 **`index.py:236`** — `update_item` upserts, so calling `_queue_cache_update` for a nonexistent run would create a phantom `testrun#<id>/metadata` item holding only PK/SK/`CacheUpdateQueuedAt`. Unreachable from all three current call sites (each already loaded the item), and the list-view fallback scan filters on `CreatedAt` so a phantom wouldn't surface — but `attribute_exists(PK)` closes it for free.

- 🟢 **`index.py:977-985`** — A permanently-failing aggregation now re-enqueues every 5 minutes indefinitely, with no backoff and no attempt cap. Bounded in practice because it is viewer-driven (the poll only runs while Test Studio is open), but an attempt counter would be more disciplined.

- 🟢 **`index.py:1019`** — The Athena *fallback* path still calls the raw helpers and can propagate, deliberately (caching all-zero metrics for a run that has real data is worse than caching nothing). That asymmetry with `_athena_supplements` deserves a comment at the fallback site, not only in the helper.

- 🟢 New `except` blocks use `logger.warning(...)` without `exc_info=True`, which this project's error-handling convention calls for.

- 🟢 Tests: `side_effect=lambda q, d: captured.append(q) or []` leans on `append` returning `None` — spell it out. And there's no test for the seam itself, that `handle_cache_update_request` writes `testRunResult` end to end for an unsafe-named run; the pieces are each covered, the join isn't.

### Pushing back on two things in my own PR

**The suggested follow-up about normalising test-run IDs at creation is questionable.** `docs/creating-custom-test-sets.md` and `docs/test-studio.md` say only "Give the test set a name" and impose no character constraints — so spaces and parens are legitimate per the documented contract, and restricting them at creation would break a documented affordance to work around a bug this PR already fixes properly. Escape-don't-restrict is the right call; that follow-up should be narrowed to sanitising only where the ID becomes an S3 prefix, if anywhere.

**Bandit/SRT was not run.** It isn't installed in my environment, and this repo's SRT gate is a GitLab CI stage that does not run on the GitHub mirror. This PR edits three `# nosec B608` justification comments and changes how the annotated queries are built — precisely what that gate exists to check. Low risk (the change strengthens the defense, and the escaping is verified live), but it is an unverified gap rather than a clean pass.

### Recommendation

Merging. The core fix is correct and verified end to end; every finding above concerns the *shape* of the new recovery machinery rather than its correctness. The two 🟡 items are tracked as a follow-up.

PR #635, opened 2026-08-21
fix(template): gate GovCloud Cognito trust principals on partition (#632)

### LINKED ISSUE #632: [Bug]:  CognitoAuthorizedRole GovCloud trust principals break commercial upgrades under a restricted permissions boundary (iam:UpdateAssumeRolePolicy AccessDenied)
### Describe the bug

  CognitoAuthorizedRole's AssumeRolePolicyDocument in template.yaml lists the GovCloud Cognito federated principals (cognito-identity-us-gov.amazonaws.com, cognito-identity.us-gov-east-1.amazonaws.com)
  unconditionally, alongside the commercial cognito-identity.amazonaws.com. These principals are present in v0.6.4 but not in earlier releases (e.g. v0.6.1).

  On a commercial-partition upgrade from a release without them, this addition rewrites the existing role's trust policy, which makes CloudFormation call iam:UpdateAssumeRolePolicy. If the stack is deployed
  with a CloudFormation service role whose permissions boundary does not include iam:UpdateAssumeRolePolicy, the update fails:

```
  User: …/CFServiceRole is not authorized to perform: iam:UpdateAssumeRolePolicy
  on resource: role …-CognitoAuthorizedRole-…
  because no permissions boundary allows the iam:UpdateAssumeRolePolicy action (AccessDenied)
```

  The stack then wedges in UPDATE_ROLLBACK_FAILED — and the rollback can't revert the trust policy either (same permission), so it can't self-recover without --resources-to-skip.

  Why the existing design doesn't fully hold

  The code comment explains the intent — list every partition's principal since "only the one matching the deploy region is exercised by STS; the others are inert." That's true for STS auth at runtime, but
  the inert principals are not inert at deploy time: they are part of the trust-policy document, so adding them changes the document, which requires iam:UpdateAssumeRolePolicy on any existing role. Fresh
  creates are fine (the trust policy is set via iam:CreateRole); only upgrades of an existing role hit this.

  **Impact**

  - Commercial deployments upgrading to v0.6.4 under a restricted CFN-service-role boundary fail and wedge the stack.
  - Recovery requires continue-update-rollback --resources-to-skip CognitoAuthorizedRole, and there's no clean forward path without either widening the org boundary or patching the template.

  **Scope / who's affected**

  Triggers only when all of these hold together:
  1. Commercial partition (aws) — GovCloud is correct/intended and unaffected.
  2. An upgrade of an existing stack that first introduces the GovCloud principals — i.e., the trust-policy-changing update of an existing CognitoAuthorizedRole (the jump from a release without them to one
  with them).
  3. A CFN service role with a permissions boundary that omits iam:UpdateAssumeRolePolicy (a boundary that grants iam:CreateRole/UpdateRole/PutRolePolicy but not UpdateAssumeRolePolicy).

  **Not affected:**
  - Fresh creates — the trust policy is set at iam:CreateRole time; no UpdateAssumeRolePolicy call (clean even in commercial + restricted boundary).
  - Deployments with no boundary or a permissive boundary — the update just succeeds.
  - GovCloud — the principals are the intended configuration.
  - Subsequent updates once already on the new version — the trust policy is stable and won't re-trigger, unless a future release changes it again (which is why a permanent partition guard is preferable to a
  one-time boundary widening).

  **Proposed fix**

  Guard the GovCloud principals behind an AWS::Partition condition so commercial partitions render a commercial-only trust policy (identical to prior releases → no trust-policy change → no
  iam:UpdateAssumeRolePolicy call), while GovCloud partitions still get their principals:
```
  Conditions:
    IsGovCloudPartition: !Equals [!Ref "AWS::Partition", "aws-us-gov"]

  # in CognitoAuthorizedRole.AssumeRolePolicyDocument.Statement:
    - Effect: Allow            # commercial — always present
      Principal: { Federated: cognito-identity.amazonaws.com }
      ...
    - !If
      - IsGovCloudPartition
      - { Effect: Allow, Principal: { Federated: cognito-identity-us-gov.amazonaws.com }, ... }
      - !Ref "AWS::NoValue"
    - !If
      - IsGovCloudPartition
      - { Effect: Allow, Principal: { Federated: cognito-identity.us-gov-east-1.amazonaws.com }, ... }
      - !Ref "AWS::NoValue"
```
  This is behavior-neutral for both partitions at runtime and eliminates the upgrade-time trust-policy churn on commercial.

  **Alternatives considered**

  - Document that the CFN service role needs iam:UpdateAssumeRolePolicy — works, but pushes an org-boundary change onto every restricted-boundary adopter for a principal that's never used in commercial.
  - Leave as-is — repeats the wedge for boundary-restricted commercial upgrades.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### AWS Region

_No response_

### Accelerator Processing Mode

_No response_

_No response_

### Custom Stack Parameters

_No response_

### Custom Configurations

_No response_

### Output of the 'Troubleshoot' agent (if issue is a document processing failure)

_No response_

### DeepWiki answer link

_No response_

### Additional context

_No response_

rstrahan:
Fixes #632.

## The bug

`CognitoAuthorizedRole` has listed the two GovCloud Cognito federated principals unconditionally since **v0.6.2** (`23e27362c`) — not v0.6.4 as the issue title says; v0.6.4 is just where the reporter landed. The affected upgrade path is any **pre-v0.6.2 → v0.6.2+**.

The report's diagnosis is exactly right. The existing code comment says the non-matching principals are "inert" — true for STS at runtime, but not at deploy time. They are part of the trust-policy document, and changing that document on an **already existing** role is a separate `iam:UpdateAssumeRolePolicy` call. So every commercial upgrade rewrote the trust policy for principals commercial never uses, and under a CloudFormation service role or permissions boundary lacking that action the update failed and wedged in `UPDATE_ROLLBACK_FAILED` — unrecoverable without `--resources-to-skip`, because the rollback needs the same action. Fresh deploys were never affected (`iam:CreateRole` sets the trust policy).

## What this PR changes

**1. The fix** — new `IsGovCloudPartition` condition; both GovCloud statements wrapped in `!If [..., !Ref "AWS::NoValue"]`, as proposed in the issue.

Verified churn-free in **both** directions:
- Commercial renders a document byte-identical to `v0.6.1` (I diffed it — the surviving statement is unchanged), so CloudFormation sees no diff and makes no API call.
- GovCloud keeps all three statements, so GovCloud stacks already on v0.6.2+ see no diff either. Both GovCloud statements stay present because the Cognito principal is per-**region**, not per-partition; gating on region would churn deployed GovCloud stacks for no benefit.

**2. The blast radius was wider than reported.** The issue frames condition 3 as a customer permissions boundary that omits the action. But the CloudFormation service role **this repo ships** (`iam-roles/cloudformation-management/IDP-Cloudformation-Service-Role.yaml`) granted `iam:CreateRole`/`UpdateRole`/`PutRolePolicy` and **not** `iam:UpdateAssumeRolePolicy` — the string appeared nowhere in the repo. So anyone following our own documented service-role setup hit this, boundary or not.

That role (and its README) now also grant `iam:UpdateAssumeRolePolicy`, `iam:PutRolePermissionsBoundary` and `iam:DeleteRolePermissionsBoundary`. The partition guard fixes *this* trust-policy change; these grants mean the *next* one — or any change to the `PermissionsBoundaryArn` parameter — can't wedge anyone either.

**3. Why CI didn't catch it.** `scripts/sdlc/validate_service_role_permissions.py` runs on every MR to develop and exists to catch exactly this. It had been **passing vacuously**: the IAM comparison crashed on the first `Fn::If`-wrapped inline policy and a broad `except` printed the `AttributeError` and returned an *empty* action set — so "no missing IAM permissions" meant "nothing was compared". It extracted **0** actions from `template.yaml` before this PR, 44 after. It now walks intrinsics tolerantly, fails loudly on an unparseable template instead of degrading to "nothing required", and additionally derives the IAM **control-plane** actions CloudFormation needs to *manage* the roles our templates declare — not just the actions those roles grant, which was the blind spot that let this class of bug through.

**4. Operator-facing** — `docs/troubleshooting.md` gains the symptom plus the `continue-update-rollback --resources-to-skip` recovery, and the service-role README explains why create-time permissions aren't sufficient.

## Tests

`scripts/sdlc/tests/test_iam_trust_policy_partitions.py` — contains a small three-valued (`True`/`False`/undecidable) CloudFormation condition evaluator, so the assertions are semantic rather than structural. It **renders every IAM role's trust policy in every role-declaring template, per partition**, treating parameter-driven conditions as "either branch reachable" — so the guarantee holds for all parameter combinations, not one sampled deployment. It then asserts:

- no GovCloud/China-only principal survives a **commercial** rendering (18 templates, 43 trust policies);
- the commercial Cognito document matches the pre-v0.6.2 one **byte-for-byte** (a statement count wouldn't catch a reorder that still triggers the API call);
- GovCloud keeps all three principals;
- each statement's OIDC condition-key namespace matches its own principal — those namespaces are literal map keys that can't be built with `Fn::Sub`, so they're hand-copied per statement, and a copy-paste slip yields a trust policy that validates fine but authorizes nobody (the original GovCloud symptom, `InvalidIdentityPoolConfigurationException`);
- the shipped service role grants the update-only actions, and the README hasn't drifted from it.

`test_validate_service_role_permissions.py` pins the validator's tolerant walk and control-plane derivation, including a regression test that extraction is never silently empty.

**Coverage was double-checked against ground truth, which found three gaps in my own first pass** (second commit): it was scanning gitignored `.aws-sam/build/` copies; two role-declaring template groups (`iam-roles/`, `scripts/sdlc/cfn/`) sat outside the globs while the template *count* still looked healthy; and a partition-specific principal could have been smuggled in via `!FindInMap` on `AWS::Partition` (this repo already has a partition Mapping) since the renderer left `FindInMap` opaque. All three are fixed, and `test_every_role_declaring_template_is_scanned` now asks `git grep -l AWS::IAM::Role` for the real list so the globs can't go stale silently.

**Everything is mutation-verified** — each check was confirmed to fail when the thing it guards is broken:

| Mutation | Result |
|---|---|
| Revert the partition guard | 2 tests fail |
| Drop `iam:UpdateAssumeRolePolicy` from the service role | test fails **and** the CI validator reports `❌ Missing IAM permissions` |
| Copy-paste slip in a GovCloud condition-key namespace | namespace test fails |
| Gov principal appended to a role in `patterns/unified/template.yaml` | caught (proves multi-template coverage) |
| Same principal smuggled via `!FindInMap` | caught |
| Remove a glob from `TEMPLATE_GLOBS` | coverage guard fails |

Local: `scripts/sdlc/tests` 240 passed, `lib/idp_sdk/tests/unit` 300 passed, `cfn-lint` clean on both changed templates (only pre-existing `W1028`s elsewhere), `ruff check`/`format` clean, `make check-arn-partitions` clean.

## Upgrade notes

- Anyone **currently wedged**: `aws cloudformation continue-update-rollback --stack-name <StackName> --resources-to-skip <StackName>-CognitoAuthorizedRole`, then upgrade to a release containing this fix. Worth checking drift on that role afterwards — skipping it leaves CloudFormation's recorded state diverged from IAM, and this fix lands them back on the document IAM actually holds.
- Anyone **already on v0.6.2–v0.6.4** with a permissive boundary: upgrading shrinks the trust policy back, which is one more `UpdateAssumeRolePolicy` call. They already have the permission, so it succeeds.
- **GovCloud**: no change in rendered output.

## Note for reviewers

`make test` is currently broken on `develop`, independently of this PR: `feature-platform/seller-entitlement-service/tests` (added in `3de5c47ea`) isn't registered in `scripts/run_all_tests.py`, and the registry guard hard-errors before running anything. I left it alone as unrelated scope and ran the affected suites directly; worth a one-line follow-up.


rstrahan:
Follow-up commit: the first CI run passed, but checking *what* it ran exposed two gaps.

**1. `scripts/sdlc/tests` wasn't in the GitHub Actions gate at all** — so the new trust-policy guards (and the 230 existing tests there, including the API-security cases) ran only on developers' machines and the green check above didn't actually execute them. That's precisely the situation `test-packages-cicd` was created to end, per its own comment, so the suite is now part of it. Verified green with no AWS credentials or config on PATH.

**2. `make test` was hard-erroring on develop before running anything** — `feature-platform/seller-entitlement-service/tests` (added in `3de5c47ea`) was in neither `RUN_ROOTS` nor `QUARANTINE`, and the registry guard treats an unclassified test directory as fatal. It's already run explicitly by `make test-packages-cicd`, i.e. known green headless, so it's now registered. This is the follow-up I flagged in the PR description — folding it in here since it otherwise blocks the full-suite run.

Issue #634, opened 2026-08-21
[Bug]: Model incorrectly extracts based on smudges/artifacts in pdf.


sindhu-achuthan:
### Describe the bug

Describe the bug:

When extracting boolean signature fields from Form XXX (emailed) the model incorrectly returns `true` for signature presence  with high confidence when only smudges, stray marks, or scanning artifacts are present in the signature area. A signature should be a handwritten name or initials - small marks and artifacts should not qualify.

This causes false positive signature detection, which can lead to downstream processing errors where documents are treated as signed when they are not.

### To Reproduce

- Upload the emailed document where the signature field is empty but contains minor smudges or marks

- Configure extraction with a boolean `Signature-of-*****` field

- Run document extraction

- Observe that `Signature-of-*****` returns `true` despite no actual signature

rstrahan:
Thanks for the detailed report — this reproduced exactly as described, and the investigation turned up more than one cause. Summary below, plus the config change you'll want, since **code alone does not fix it**.

## Root cause

Reproducing your case (an unsigned Form 4549 whose signature cell holds only a faint smudge, with a date filled in beside it), the model was inferring "signed" from the **date in the adjacent column**. Its own recorded reasoning:

> "The extraction indicates 'true' for signature presence, which appears correct based on the document structure and **the presence of a date**. However, without being able to visually verify an actual signature mark, confidence is set at 0.85."

Textract actually agreed with you. With `SIGNATURES` enabled it emitted **no** signature block for the cell in question, and a single detection at **11.0% confidence** over a stray mark in the *other* taxpayer's cell. That evidence existed and reached nothing useful:

1. **`SIGNATURE` blocks were dropped from every derived artifact.** They carry a confidence and bounding box but no text, and both `textConfidence.json` (which feeds the confidence pass) and `pageData.json` (which feeds the UI) filtered to `LINE` blocks. The only trace left was the linearizer's bare inline `[SIGNATURE]` token — which is placed by *reading order*, so it landed right after taxpayer 1's date while the detection was in taxpayer 2's cell.
2. **The confidence pass ran blind.** Page images were being dropped whenever `geometry.mode` was `ocr_only` (the default), so a `{DOCUMENT_IMAGE}` placeholder in the prompt rendered as an empty block — for a visually-evidenced field, the one thing that could have caught the error.
[...]

PR #638, opened 2026-08-21
fix(assessment): resolve $ref subschemas when formatting field descriptions

rstrahan:
## Problem

The confidence (assessment) prompt's `{ATTRIBUTE_NAMES_AND_DESCRIPTIONS}` block was **empty of descriptions** for any class whose groups or lists live in `$defs` — which is how the UI's schema editor writes them, and what `docs/few-shot-examples.md`-style hand-authored v0.6 configs use.

`_format_property_descriptions` reads `type` and `description` straight off each property. A `$ref` property carries neither — both live in the referenced `$defs` entry:

```jsonc
"properties": {
  "Signatures": { "$ref": "#/$defs/Signatures" }   // no type, no description
}
```

So `prop_type` is `None`, the `TYPE_OBJECT` branch is skipped (which is also the only branch that walks nested members), and the line rendered is literally:

```
Signatures  	[  ]
```

Real request body captured from a live stack (`Form-4549`, two group properties, both behind `$ref`):

```
<attributes-definitions> HeaderInformation  	[  ]
Signatures  	[  ] </attributes-definitions>
```

The confidence model is asked to judge values whose meaning it was never told. That matters most where the field's *definition* is the whole question — a boolean `"Does the signature of taxpayer 1 on the left lower side of the 2nd page exist?"` arrives as a bare name, and the value got a 0.85.

Extraction was never affected: it passes the raw schema, `$defs` included.

## Change

New `_deref_schema(node, root)` resolves a local `#/$defs/<name>` reference before type/description are read, at every level the formatter walks: property, group member, list `items`, and item column.

- Follows `$ref` chains.
- Sibling keys on the referencing node win over the definition's (a local `description` overrides the shared one).
- Cycles, dangling refs, remote refs, and non-dict nodes all degrade to the previous behavior rather than raising.

## Tests

Three new tests. Mutation-verified — making `_deref_schema` a pass-through fails `test_format_property_descriptions_resolves_refs`.

- `tests/unit/assessment`: 208 passed
- ruff + basedpyright clean

Refs #634


sindhu-achuthan:
## Summary: PR #638 is a PARTIAL fix

### What PR #638 Fixed (✅)
- Added `_resolve_ref()` and `_get_effective_schema()` methods
- **Task creation** now properly resolves `$ref` (line 703)
- Signatures group correctly classified as `group` task type (not `simple_batch`)

### What PR #638 MISSED (❌)  
- `_format_property_descriptions()` (lines 319-368) **NOT updated**
- Assessment **prompt** still has empty attribute descriptions for `$ref` groups
- Model doesn't receive signature field descriptions ("Small marks NOT signatures")

### Evidence from Logs
```
<attributes-definitions> Signatures  	[  ] </attributes-definitions>
```
Child properties (`Signature-of-taxpayer1`, etc.) missing!

### Why signatures still wrongly assessed
1. Extraction correctly returns `"Signature-of-taxpayer1": true` (but maybe wrong)
2. Assessment task is correctly created as `group` type ✅
3. But prompt to Bedrock has **no description** ❌
4. Assessment model can't verify if the signature is real vs a "mark/squiggle"
5. Extraction error passes through

This is a **gap in PR #638** - the task creation was fixed but the prompt formatting was not.

PR #641, opened 2026-08-21
fix(feature-sdk): make marketplace.licenseMode reachable, bake it on both publish paths, regenerate GraphQL types

rstrahan:
## Summary

Follow-up to #636. That change wired `licenseMode`'s reader and one of its two bakers, but not the schema — so the manifest route shipped as **dead code**. Reported from an extension repo that tried to use it; reproduced here, plus a second gap sitting behind it.

## 1. The schema rejected the field before the model saw it

`feature-manifest.schema.json` declares `marketplace.additionalProperties: false` and did not list `licenseMode`:

```
$ idp-feature-cli validate <feature>
✗ feature.yaml is invalid:
  • marketplace: Additional properties are not allowed ('licenseMode' was unexpected)
```

So `manifest.py` reading it and `pack.py` baking it were both unreachable. Worse, the commented example I added to `feature-template/feature.yaml` leads a reader straight into that error.

Now listed with an `enum` of the three values, so a typo fails at **publish** time. The host's install-time drop is the right fail-safe but far too late to be the only check — by then the extension is deployed and the host is quietly using its catalog default instead.

## 2. The token was baked on only one of two publish paths

`pack.py` and `publisher.py` each carried an independent `.replace()` chain over the same three `marketplace.*` tokens. Only `pack.py` learned about the third, so a feature published through `FeaturePublisher` shipped a literal `<FEATURE_LICENSE_MODE_TOKEN>` in its template. Caught by extending the existing "all tokens baked" assertion.

Both paths now call one `bake_marketplace_identity()` helper. It lives in `manifest.py`, next to the fields it reads — `pack.py` already imports `publisher.py`, so it can't live in either without a cycle.

## 3. A diagnostic that gave wrong advice

`registerFeature` told operators to fix *"the feature's manifest (marketplace.licenseMode)"* for an invalid value. The host has no idea where the value came from: the bundled SDK bakes it from the manifest, but an extension is equally free to declare it as a CloudFormation `Mappings` constant, a stack parameter, or anything else its ui-deployer reads — the host only requires it in the `registerFeature` payload. The warning now points at the ui-deployer's source without naming a file.

The developer guide documents that payload-only route explicitly, and says that an extension which already keeps the mode elsewhere should **omit** the manifest field: two declarations of one fact is precisely the drift `licenseModeMismatch` exists to detect, and duplicating it would have the detector firing on a bug introduced by the duplication.

## Tests

Regression tests go through `load_manifest` (which validates) rather than constructing `MarketplaceSpec` directly — reader, baker and schema all have to agree, and testing the dataclass alone is what let this ship:

- every enum value validates and reaches the model
- absence stays `None` (so a caller can tell "said nothing" from "said check nothing"; `pack.py` is what turns absence into `none`)
- a typo is rejected at validate time
- every token, including the new one, is baked on the publish path

All three fixes **mutation-verified** against the shipped state: reverting the schema fails 3 tests, reverting `publisher.py` fails the token-bake test.

- `lib/idp_feature_sdk/tests` — **140 passed**
- `feature-platform/main-stack-extensions/tests` — **202 passed**
- `lib/idp_sdk/tests/unit` — **306 passed**
- `ruff` — clean

## Note

No behaviour change for a deployed stack, and no change to the resolver's authority logic. The host side of #636 was already correct — an extension that sends `licenseMode` in its `registerFeature` payload worked before this PR and works after it. What was broken was only the *bundled SDK's* manifest route for producing that value.


rstrahan:
Rebased onto current `develop` (conflict was `CHANGELOG.md` only — five merged PRs all insert at the top of `### Fixed`; all entries preserved).

Also added a second commit that fixes a CI failure **not introduced by this PR**: `make codegen-check` fails on `develop` itself. #636 added five fields to `schema.graphql` — `FeatureEntitlement.{licenseMode, declaredLicenseMode, catalogLicenseMode, licenseModeMismatch}` and `RegisterFeatureInput.licenseMode` — without regenerating `src/ui/src/graphql/generated/`, so every PR branched off develop has been red on that check for a reason none of them caused. Verified by running the check on a clean `develop` checkout. Generated output only, and the diff is exactly those five fields.

Local verification on the rebased branch:
- `make codegen-check` — ✅ up-to-date
- `make lint-cicd` — all code quality checks passed
- `lib/idp_feature_sdk` + `feature-platform/main-stack-extensions` — 342 passed
- `src/ui` vitest — 221 passed

PR #642, opened 2026-08-21
feat(ocr): enable Textract SIGNATURES by default; fix Textract cost figures

rstrahan:
> **Stacked on #637** — base is `fix/ocr-capture-signature-detections`, so this diff shows only the default flip. Merge #637 first; GitHub will retarget this to `develop` automatically. Without #637 the default would emit only bare, unpositioned `[SIGNATURE]` tokens with no confidence, which is the misleading state that caused #634.

## Change

Default `ocr.features` becomes **`TABLES` + `LAYOUT` + `SIGNATURES`**.

## Is it really free?

Yes — and AWS says so explicitly. From the [Textract pricing page](https://aws.amazon.com/textract/pricing/):

> \*Signatures feature is included free of cost with any combination of Forms , Tables, Queries, and Layout

(the same footnote block carries the equivalent rule for Layout). Worth knowing for future verification: that footnote and the whole rate table live in the pricing widget's markup and are rendered client-side from `pricing/2.0/meteredUnitMaps`, so a plain markdown fetch of the page does **not** contain them — you have to read the raw HTML or the Price List API.

Our own bill corroborates the *mechanism*. Cost Explorer for this account, grouped by usage type:

| Month | Usage types billed |
|---|---|
| 2026-08 | `SyncTablesPagesProcessed` 4,510 pages — **$67.65** |
| 2026-07 | `SyncTablesPagesProcessed` 12,191 · `SyncLayoutPagesProcessed` 7,298 · `SyncFormsPagesProcessed` 20 |
| 2026-06 … 2026-02 | Tables / Layout / Forms / Text only |

`SignaturesPagesProcessed` has never appeared in any month — including August, when config-634 ran `LAYOUT + SIGNATURES`. And August shows Tables with **no** Layout line at all despite `TABLES + LAYOUT` being the default for every one of those pages: AWS emits **no usage type** for a feature that is free in combination. Months that *do* show a Layout line are LAYOUT-only configs.

`_feature_combo()` already encoded this, so metering is unchanged (`TABLES+LAYOUT+SIGNATURES` → `-Tables`).

**The one case that is billed:** `SIGNATURES` *alone*, with none of Forms/Tables/Queries/Layout, costs ~$0.0035/page. A test asserts the shipped defaults never leave SIGNATURES without a paying feature.

## Caveat, documented rather than assumed

**A detection is not proof of a signature.** Textract flags stray pen marks and scan artifacts, usually at low confidence (#634's was 11%). That costs a few prompt tokens on affected pages and gives a model something to over-read. `docs/configuration.md` now tells corpora without signature fields to drop the entry, and notes that removing it costs nothing since the feature was free. The `_feature_combo()` docstring and `base-ocr.yaml` cite the pricing-page rule and point at `config_library/pricing.yaml` for whoever revisits it if Textract's usage types change.

## Bonus: the shipped cost figures were wrong by more than 4×

While editing these lines I found `base-ocr.yaml` and `docs/configuration.md` both state `TABLES`+`LAYOUT` ≈ **$0.065/page**, "~16×" the LAYOUT-only rate. **$0.065 is the Tables+Forms rate.** The real billed rate is **$0.015/page** — verified against this account's invoice: 4,510 pages → $67.65 = exactly $0.015/page — i.e. ~3.75×, not 16×.

So the docs had been overstating the cost of the *default configuration* by 4.3× for anyone deciding whether to keep `TABLES`. Corrected in both places, along with the BDA-vs-Textract comparison built on it (which claimed $0.065 was "often comparable" to BDA's $0.01 flat rate; the honest statement is $0.015 vs $0.010, converging above 1M pages/month where Textract's Tables tier drops to $0.010).

## Side effect worth noting

`lending-package-sample` and `lending-package-sample-govcloud` set **no** `ocr.features` at all, while both READMEs claim "Amazon Textract with LAYOUT, TABLES, and SIGNATURES features" and "Uses SIGNATURES feature to detect signed documents". They inherit the default, so this PR finally makes those READMEs true without touching either config.

## Tests

Four new tests: the shipped default set is exactly `[TABLES, LAYOUT, SIGNATURES]`; it meters as `-Tables` alone; `SIGNATURES` adds no priced component to Forms/Layout/Tables+Forms either; and `SIGNATURES` is never left in the defaults without a paying feature (alone it costs ~$0.0035/page). Mutation-verified — removing the default entry fails the first.

- `make test` (idp_common_pkg): 3086 passed, 55 skipped
- ruff + basedpyright clean

## Opting out

Remove the `SIGNATURES` entry from `ocr.features`. Existing deployments with a **custom** `ocr.features` list are unaffected; those on the stack-managed default pick it up on upgrade.

Refs #634


rstrahan:
**Correction to the original description.** It claimed AWS does not document the Signatures free-in-combination rule and that this PR rested on observed billing rather than published terms. That was wrong — the [Textract pricing page](https://aws.amazon.com/textract/pricing/) states it directly:

> \*Signatures feature is included free of cost with any combination of Forms , Tables, Queries, and Layout

The footnote and rate table live in the pricing widget's markup (rendered client-side from `pricing/2.0/meteredUnitMaps`), so a markdown-converted fetch of the page omits them; reading the raw HTML shows it. The description, `base-ocr.yaml`, the `_feature_combo()` docstring, `docs/configuration.md`, the CHANGELOG and one test docstring now cite the published rule, with the Cost Explorer data kept as corroboration of the mechanism. No behavior change.

PR #652, opened 2026-08-24
feat(ui): flag classes that disagree with ground truth, where the class is shown (#627, #630)

### LINKED ISSUE #627: [Feature]: Improve the visualization of classification errors when reviewing test sets

I'm having a hard time following why my test set scores are low when there's a classification error. I don't have a clear indication in the data view mode of if my extraaction was low scoring b/c it was the wrong class or because it was an incorrect extraction

I want to see clear classificaiton errors displayed after I select a document from the 'lowest scores' category, showing the specific mismatches either at the page or packet level, outside of the evaluation report. In the same way we can toggle on 'show Evaluation' and then select eval mismatches only, I want to be able to easily see the errors in terms of page level classification in the viewer. 

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #630: [Bug]: When viewing evaluation report vs. Document sections after clicking through on documents from the Test studio, classification is hard to understand
### Describe the bug

I am confused about the ground truth vs. the prediction when viewing documents from test studio. It's hard to undersatnd what the true vs. preicted was of the classifications. The evaluation report doesn't seem to agree with what's being displayed, and with what's being shown in 'document sections'. 

Image

### To Reproduce

Create a test set with packets and ground truth
Do the inference and get something wrong
Go to view that document after selecting it based on low scores
See the confusion!

### Expected behavior

We need a clearer 'this is what the document sections were denoted as in the ground truth' vs. 'this is what our run extracted'. 

### Screenshots

_No response_

### AWS Region

_No response_

### Accelerator Processing Mode

_No response_

_No response_

### Custom Stack Parameters

_No response_

### Custom Configurations

_No response_

### Output of the 'Troubleshoot' agent (if issue is a document processing failure)

_No response_

### DeepWiki answer link

_No response_

### Additional context

_No response_

rstrahan:
Fixes #627. Fixes #630.

Both issues (from @sromoam) describe the same gap from two angles: arriving at a low-scoring document from Test Studio's *Documents with Lowest Weighted Overall Scores*, there is no way to tell whether the score is low because a **field** was extracted wrongly or because the **page was classified as the wrong document type** in the first place. The Sections and Pages tables show only what the run predicted, with nothing to compare against, and the only ground-truth view is the evaluation report — a different place, in a different shape.

The comparison was already being computed — `doc_split_metrics.page_details` in `evaluation/results.json` records every page's expected and predicted class — but **nothing surfaced it**. The markdown report renders section-level tables only, so a page misclassification was visible as a number with no way to see which page went where.

> **Revised after review feedback.** The first version of this PR added a standalone *Classification vs Ground Truth* panel to the document page. That was the wrong shape — we already have **View Evaluation Report**, and a panel costs screen space on every document. This version puts the same information where the class is already displayed and adds nothing to a document without ground truth. The panel and its section-pairing/summary parsing are gone.

## What this adds

**1. Document Sections and Document Pages tables.** A section or page whose class disagrees with ground truth carries a **Class mismatch** alert beside its Class/Type value; hovering gives the expected class, the class this run assigned, and the pages involved. Also shown in **Edit mode** next to the class dropdown — that is where the class can actually be corrected, so it is the most actionable place for it.

**2. Visual Editor ("View Data") under Show Evaluation.** The toggle now compares the section's `document_class` as well as its fields, above the field list: this run's class, ground truth's class, and a match/mismatch verdict. Both values are shown whether or not they agree, since the user asked to see the comparison. It reads the same `results.json` the field-level comparison already loads, so it costs **no extra request**, and falls back to the baseline file's own `document_class.type` for a document with no page-level comparison.

**3. All three say why it matters.** A misclassified page was extracted against the **wrong schema**, so its low extraction score is a symptom rather than the cause — stated inline, because that inversion is what made the original reports confusing.

## Two judgement calls worth reviewing

**A section's verdict is derived from its pages, not by pairing section ids.** When the split itself is wrong there is no single counterpart section to compare against — but every page still has a ground-truth class. So a section that merged two ground-truth documents reports *both* classes it spans ("Section spans more than one ground-truth class") instead of an unhelpful "no match", and is flagged **even when one of those classes equals its own**, because there the boundary is what is wrong. Pinned by tests in both the util and the component.

**Silence is the default.** A row whose class matches gets no annotation, and a page ground truth says nothing about is left unannotated rather than flagged — an incomplete baseline is not a misclassification. With no evaluation report there is no annotation *and no request*: `useClassificationComparison` never fetches. So a document with no ground truth renders exactly as it did before this PR.

## Correctness details

- **Page numbers are 1-based.** `page_indices` in `results.json` are 0-based (computed as `page_id - min(page_id)`), so joining on the raw index would annotate every row one page off — worse than not annotating, since the alert names a class. Converted once in the util, and pinned by row-scoped wiring tests (see below).
- **`results.json` is derived from the document's own `evaluationReportUri`** (`.../evaluation/report.md` → `.../evaluation/results.json`) rather than rebuilt from bucket + input key. Both are written to the same prefix by `EvaluationService`, so this follows whatever prefix that document used — and with `versionIdForUri`, a historical run snapshot resolves to *that run's* results.
- **One fetch per document.** The hook runs in `DocumentPanel` and the index is passed to both tables, rather than each panel fetching for itself. The Visual Editor reuses the results it already has.
- A fetch failure (results pruned by retention, an older document) leaves the tables unannotated and logs at debug — there is nothing the user can act on.

## Tests

41 tests, all passing; full UI suite 291 passed; `lint`, `typecheck` and `build` clean.

- `classification-comparison-utils.test.ts` (21) — index keying and 0→1-based conversion; stickler's `"Missing"` placeholder treated as *absence* of ground truth rather than a class named "Missing"; malformed entries skipped instead of indexed under `NaN`; the full verdict matrix including `unknown ≠ mismatch`, the multi-class span, and "class right, some pages uncovered → still a match"; URI derivation (including a key containing `report.md` earlier in the path).
- `ClassMismatchIndicator.test.tsx` (12) — renders nothing with no ground truth, nothing on a match, nothing for an uncovered page; popover contents for a wrong class and for a merged section; the editor variant showing both values, its baseline fallback, and rendering nothing with neither source.
- `PagesPanel.classMismatch.test.tsx` (2) and `SectionsPanel.classMismatch.test.tsx` (2) — the join, asserted **row-scoped**: the annotated row must be the right one. Counting indicators alone would not catch an off-by-one (one alert on the wrong row is still one alert), and an early version of these tests didn't — verified by mutating the page-number join and the section page list and confirming each test fails.

Mutation-verified overall: reverting the `hasGroundTruth` guard, the `unknown`-vs-`mismatch` rule, the multi-class verdict, the page-number join, or the section page list each fails a specific test.

## Docs

- [docs/evaluation.md](docs/evaluation.md#seeing-which-pages-were-misclassified-web-ui) — rewritten for the in-place annotations, including the page-derived section rule and the two deliberate silences.
- [docs/test-studio.md](docs/test-studio.md#results-analysis) — "Diagnosing a low-scoring document", cross-linked from where the workflow starts.
- `CHANGELOG.md`.


rstrahan:
Reworked per feedback — the standalone panel is gone.

Your read was right on both counts: we already have **View Evaluation Report** for the report-shaped view, and a panel costs screen space on every document to answer a question most documents don't have. The information now lives where the class is already displayed:

**(b) Sections and Pages tables** — a **Class mismatch** alert beside a Class/Type value that disagrees with ground truth, hover giving the expected (GT) class, the class this run assigned, and the pages involved. Also in **Edit mode** next to the class dropdown, since that's where the class can actually be corrected.

**(a) Visual Editor under Show Evaluation** — the section's `document_class` compared above the field list: this run's class, ground truth's class, match/mismatch. Both values shown whether or not they agree, since the toggle is an explicit request to see the comparison. It reads the `results.json` the field-level comparison already loads, so it adds **no request**.

Two things worth knowing about the shape it took:

- **A section's verdict comes from its pages, not from matching section ids.** When the split itself is wrong there's no single counterpart section to compare against, but every page still has a ground-truth class. So a section that merged two ground-truth documents names *both* classes it spans, and is flagged even when one of them equals its own — there the boundary is what's wrong, not the label.
- **Nothing is added unless ground truth disagrees.** A matching row gets no annotation; a page ground truth says nothing about is left alone rather than flagged (an incomplete baseline isn't a misclassification); and with no evaluation report there's no annotation *and no request* — the hook never fetches. A document without ground truth renders byte-for-byte as it did before.

Both tables share one fetch (loaded once in `DocumentPanel`, passed down) rather than one each.

41 tests. The wiring tests are row-scoped rather than count-based: an off-by-one in the page-number join still produces exactly one alert, just on the wrong page — the first version of those tests didn't catch that, which is why they now assert *which* row is flagged.

PR #672, opened 2026-08-27
fix(security): eliminate the flagged yaml.load call shape from all six CFN loaders

rstrahan:
## What and why

ACAT reported **Unsafe YAML Load (High)**, finding `ffaca17a-1d02-4fd7-ae2a-56bfabc66497`, against `feature-platform/seller-entitlement-service/tests/test_template_security.py:50`.

**The RCE claim is a false positive** — assessed and being marked NOT_USEFUL:

- The loader subclasses `yaml.SafeLoader`, never `yaml.Loader`/`FullLoader`, so the `python/object`, `python/name` and `python/object/apply` constructors that make `yaml.load` dangerous are not registered.
- Its multi-constructor returns only plain scalars/lists/dicts, so it cannot construct or evaluate code.
- The parsed input is the package's own committed `template.yaml`, read by a unit test — no external input reaches it, and the file is not packaged into the deployed Lambda. Anyone able to edit that template could equally edit the test parsing it.

**What is true** is that `yaml.load(..., Loader=...)` is the shape pattern-based scanners key on *regardless* of the loader's base class, and `# nosec B506` is Bandit's marker, which ACAT does not honour. So the finding would keep resurfacing and keep needing hand triage. This PR removes the shape.

The loader is now driven directly — construct, `get_single_data()`, `dispose()` — which is exactly what `yaml.load` does internally.

## Scope: all six sites, not just the flagged one

The same pattern was copy-pasted six times, so the other five would have generated the same ticket as ACAT's coverage widened:

| Site | Change |
|---|---|
| `feature-platform/.../test_template_security.py` | local loader, call shape removed (the ACAT site) |
| `lib/idp_sdk/tests/unit/test_permissions_boundary_coverage.py` | uses shared helper |
| `lib/idp_sdk/tests/unit/test_webui_rebuild_triggers.py` | uses shared helper |
| `scripts/sdlc/validate_service_role_permissions.py` | local loader, call shape removed |
| `scripts/sdlc/tests/test_iam_trust_policy_partitions.py` | local loader, call shape removed |
| `lib/idp_sdk/idp_sdk/_core/publish.py` | uses shared helper (only non-test site) |

New shared helper: `lib/idp_sdk/idp_sdk/_core/cfn_yaml.py`.

### Why only three sites share it

I set out to dedupe all six and backed off on three, for reasons worth stating:

- **`validate_service_role_permissions.py`** — the `deployment_validation` CI job runs it with only `pip install PyYAML` (`.gitlab-ci.yml:218`). An `idp_sdk` import would break that gate. Its collapse-intrinsics-to-`None` policy is also load-bearing: `_iter_statements` is built around it.
- **`test_iam_trust_policy_partitions.py`** — normalizes intrinsics to long form so it can evaluate what an `Fn::If` guards. Not a shareable policy, and these tests are meant to run from the repo root with nothing installed.
- **`test_template_security.py`** — `ruff.toml:80` bans importing `idp_sdk._core` from outside the SDK, and `feature-platform/**` is not on the exemption list. I did not add one; the duplication is deliberate and commented.

Those three get the call-shape swap only, so their behaviour is unchanged.

## Verification

- **Bandit across all touched files: 0 high, 0 medium, and 0 lines skipped via `#nosec`** — the finding is eliminated, not suppressed. All six `# nosec B506` annotations are gone. Remaining lows are pre-existing `assert_used` in tests.
- No `yaml.load(` call remains anywhere in the repo (`grep`); the only hits are in prose explaining why.
- `make lint-cicd` clean. `make typecheck-pr` clean on all 8 changed files.
- `make test-packages-cicd` green, including the seller-entitlement suite (136), `scripts/sdlc/tests` (240) and `lib/idp_sdk/tests/unit` (327, including 21 new).
- `validate_service_role_permissions.py` still compares non-vacuously — 61 permissions, 32 IAM actions, exit 0. (Its silent-vacuous-pass was the #632 defect, so this mattered to check.)
- `publish.py`: diffed `_extract_function_name` resolutions across `template.yaml`, `patterns/unified` and `nested/api-resolvers` before vs after — **82/82 identical**.

The new tests assert the inertness rather than asserting it in a comment: four real `python/object/apply` / `python/name` / `python/object/new` / `python/module` payloads, plus one that would `touch` an observable marker file, are shown to raise or parse to plain data with no side effect. Switching the base class back to `yaml.Loader` fails them — which is the point, since that is how this reintroduces itself.

## Behaviour deltas (all verified inert)

1. **Deep construction** at the two idp_sdk test sites: a nested intrinsic inside an `!If` now resolves instead of surfacing as an unconstructed node. Strictly more resolved.
2. **`{"!Tag": value}` at the ACAT site**, replacing a policy that rendered `!Sub x` as `'!!Sub x'` (double-banged — the full tag arrived as the suffix). No assertion depended on either form; all 24 are unchanged.
3. **`publish.py` tolerates any `!` tag**, replacing a hand-maintained 16-tag list. Previously an unlisted tag raised, was swallowed by the surrounding broad `except`, and silently cost the function its answer — one `!Transform` and no function name resolves.

## Notes for the reviewer

- `_extract_function_name` in `publish.py` **has no callers anywhere in the repo**. It is dead code, which is why touching it is low-risk; deleting it is a reasonable follow-up but out of scope here. It is the last commit, so it reverts independently.
- No CHANGELOG entry: no user-visible behaviour changes.

### DIFF
--- a/feature-platform/seller-entitlement-service/tests/test_template_security.py
+++ b/feature-platform/seller-entitlement-service/tests/test_template_security.py
@@ -25,29 +25,48 @@ import yaml
 _TEMPLATE = Path(__file__).resolve().parents[1] / "template.yaml"
 
 
+# --- CFN-tolerant YAML loader -------------------------------------------------
+# `!Sub`, `!GetAtt`, `!FindInMap` etc. are not standard YAML, so SafeLoader
+# rejects them. Each intrinsic is preserved as a {"!Tag": value} dict, which
+# keeps the logical id inside a !GetAtt readable to the assertions below.
+#
+# This cannot execute the document it parses, for two independent reasons:
+# SafeLoader (never yaml.Loader/FullLoader) registers no `python/object`,
+# `python/name` or `python/object/apply` constructor, so nothing here can
+# instantiate an object or import a module; and `_intrinsic` returns only plain
+# scalars, lists and dicts. idp_sdk._core.cfn_yaml is the shared version of this
+# loader, with tests that assert the inertness against real RCE payloads — it is
+# duplicated rather than imported because ruff.toml bans reaching into
+# idp_sdk._core from outside the SDK, and this service is deployed standalone
+# into a seller account.
 class _CfnLoader(yaml.SafeLoader):
-    """YAML loader that tolerates CloudFormation's short-form intrinsic tags.
-
-    `!Sub`, `!GetAtt`, `!FindInMap` etc. are not standard YAML, so SafeLoader
-    rejects them. We only need the document's shape, so every unknown tag
-    collapses to a plain string/list.
-    """
+    pass
 
 
-def _any_tag(loader: yaml.Loader, tag_suffix: str, node: yaml.Node) -> Any:
+def _intrinsic(loader: Any, tag_suffix: str, node: yaml.Node) -> Any:
+    tag = "!" + tag_suffix
     if isinstance(node, yaml.ScalarNode):
-        return f"!{tag_suffix} {node.value}"
+        return {tag: loader.construct_scalar(node)}
     if isinstance(node, yaml.SequenceNode):
-        return loader.construct_sequence(node, deep=True)
-    return loader.construct_mapping(node, deep=True)
+        return {tag: loader.construct_sequence(node, deep=True)}
+    return {tag: loader.construct_mapping(node, deep=True)}
 
 
-_CfnLoader.add_multi_constructor("", _any_tag)
+_CfnLoader.add_multi_constructor("!", _intrinsic)
 
 
 @pytest.fixture(scope="module")
 def template() -> dict:
-    return yaml.load(_TEMPLATE.read_text(encoding="utf-8"), Loader=_CfnLoader)  # nosec B506 - _CfnLoader subclasses yaml.SafeLoader; no arbitrary construction
+    # The loader is driven directly rather than through
+    # `yaml.load(..., Loader=_CfnLoader)`. Identical behaviour — that is what
+    # yaml.load does internally — minus the call shape that pattern-based
+    # scanners report as unsafe deserialization regardless of the loader's
+    # actual base class.
+    loader = _CfnLoader(_TEMPLATE.read_text(encoding="utf-8"))
+    try:
+        return loader.get_single_data() or {}
+    finally:
+        loader.dispose()
 
 
 @pytest.fixture(scope="module")
--- a/feature-platform/seller-entitlement-service/tests/test_template_security.py
+++ b/feature-platform/seller-entitlement-service/tests/test_template_security.py
@@ -16,6 +16,7 @@ violation would be a real vulnerability, and says why in the failure message.
 
 from __future__ import annotations
 
+import tempfile
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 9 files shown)


rstrahan:
## Review pass found a regression in this PR — fixed in 525907e5 and d87cb12d

Reviewing my own change turned up one genuine regression and three hardening gaps, all introduced by this PR. Recording them here since the PR description above no longer tells the whole story.

### 🔴 The regression (fixed)

Switching `test_template_security.py` to the repo's `{"!Tag": value}` tag policy **silently disabled a live security assertion**:

```python
resource = str(stmt["Resource"])
assert not resource.rstrip().endswith("/*")   # unfalsifiable once Resource is a dict
```

`str()` of a dict always ends in `}`, so the wildcard check could never fail again — on a resource policy whose `Principal` is `*`. Proven by mutation rather than argued: appending `/*` to the policy's Resource **fails** on `develop` and **passed** on `563fc540`. `_actions()` degraded the same way, which would have made the IAM and KMS wildcard checks vacuous the first time anyone wrapped an action list in `!If`.

The flat-string rendering is now kept exactly as `develop` has it — including the cosmetically ugly `'!!Sub x'` double-bang, because these assertions match on the *ends* of those strings and tidying them is not free. Only the `yaml.load()` call shape is removed.

This is worth stating plainly: the PR as originally pushed would have weakened a security control while claiming to fix a security finding.

### 🟡 Hardening gaps (fixed)

1. `load_cfn_yaml(text, yaml.UnsafeLoader)` **executed its input** — verified, it returned `os.getcwd()`. A module premised on "cannot execute the document it parses" accepted any loader class, making the safety property conventional rather than structural. Non-`SafeLoader` classes are now refused with a `TypeError`.
2. `assert not issubclass(make_cfn_loader(), yaml.UnsafeLoader)` could never fail — in PyYAML those loaders are siblings, not ancestors. Replaced with a check that the code-executing constructors are absent.
3. Removing the `yaml.load()` shape also removed what made Bandit B506 and ACAT fire on the three sites that keep a local loader, so `class CFNLoader(yaml.Loader)` would now draw no objection from any scanner *or* test. `scripts/sdlc/tests/test_cfn_loader_safety.py` is the deliberate replacement for that accidental guard — it matters most for `validate_service_role_permissions.py`, which takes a template path as a CLI argument and runs in CI.

### Verification

Every new guard was mutation-tested, not just run: flipping a loader's base class to `yaml.Loader` fails all three sdlc tests plus the seller-entitlement one, and the payload assertions catch genuine object construction rather than only the base-class change.

Re-verified after the fixes: `make lint-cicd` clean, `make typecheck-pr` clean, `make test-packages-cicd` green (138 seller-entitlement, 246 sdlc, 329 idp_sdk), Bandit 0 high / 0 medium / 0 `#nosec`-skipped lines.

### Still open for the reviewer

`_extract_function_name` in `publish.py` has **no callers anywhere in the repo**. Refactoring dead code is defensible (it keeps the pattern consistent and the commit is independently revertible), but deleting it may be the better answer — happy to do that here or leave it for a follow-up.

rstrahan:
## On the "remove the call shape scanners key on" approach

Worth stating explicitly, since it can read badly out of context: **this is not suppression dressed up as a fix.**

- Six `# nosec B506` annotations were **deleted**, not added — none remain anywhere in the repo.
- Every loader still subclasses `yaml.SafeLoader`. The `python/object`, `python/name` and `python/object/apply` constructors are still simply not registered.
- The shared module now **structurally refuses** a non-`SafeLoader` class (`TypeError`), so the safety property is enforced rather than conventional — verified: `load_cfn_yaml("a: 1", yaml.UnsafeLoader)` raises.
- Shape-level scanner signal is replaced by behaviour-level tests, and those tests were **mutation-verified**, not merely run:
  - `CFNLoader(yaml.SafeLoader)` → `yaml.Loader` in `validate_service_role_permissions.py` fails 3/3 tests in `scripts/sdlc/tests/test_cfn_loader_safety.py`.
  - The same flip in `test_template_security.py` fails `test_loader_cannot_execute_its_input`.
  - Switching that file's `_any_tag` to the `{"!Tag": value}` policy fails `test_the_wildcard_guard_is_live` (`isinstance({'!!Sub': 'execute-api:/*/POST/activate'}, str) = False`) — so the regression caught mid-PR is genuinely re-pinned.
- No `yaml.load(` call remains in the repo; the only hits are prose explaining why.
- The `deep=True` delta is nil in practice: parsing `template.yaml` and `patterns/unified/template.yaml` with the old non-deep loader vs the shared one is **byte-identical**. Likewise the "any `!` tag" change is a latent robustness win — the only tag in the repo's templates outside publish.py's old 16-tag list is `!Condition`, which was already listed.

PR #675, opened 2026-08-27
Feature/UI acceptance testing

sromoam:
*Issue #, if available:*
Not related to an issue, but rather an attempt to provide a mechanism for UAT testing on UX. 

*Description of changes:*
Mechanisms for both deterministic and agentic UAT. 


rstrahan (commented):
## Review

Thanks for this — the design thinking is unusually good, and the commit messages are some of the most useful I've read on this repo. Detailed findings below; one blocking item.

**Size:** +3,355 / -0 across 27 files, 3 commits. Targets `develop` ✅.

I verified locally against the PR head: `ruff check` and `ruff format --check` are clean on `scripts/uat/`, `npm ci && tsc --noEmit` passes in `src/ui/e2e`, and `make check-arn-partitions` / `make check-filtered-scans` both pass.

### Summary

Two new test tiers: (1) a deterministic Playwright UI acceptance suite under `src/ui/e2e/` with a full-lifecycle driver (`scripts/uat/run_uat.py`, `make uat-testing`) that deploys → provisions Cognito users → tests → tears down, plus a nightly workflow; and (2) an experimental agentic usability prototype under `scripts/uat/agentic/`.

### Verdict

| Question | Verdict | Notes |
|---|---|---|
| Good PR? | :warning: | Well-reasoned and thoroughly explained, but two separable features in one PR — commit 1 is nearly ready, commits 2–3 are ~1,350 lines of experimental code with no CHANGELOG entry and no `docs/` coverage |
| Safe / no regressions? | :warning: | Purely additive; nothing existing is touched, and lint/format/typecheck are clean. But the self-deploy path is broken (see blocking item) |
| Good UX? | :white_check_mark: | Selector policy (roles/labels, no `data-testid`) is right; assertion failure messages are genuinely actionable; interaction-cost reporting is a good idea |
| No security issues? | :warning: | Credential model is thoughtfully designed. Two gaps: public-artifact trace upload, and template injection in the workflow |
| Well documented? | :warning: | Layer-1 doc is good but missing frontmatter/license header and unregistered in the docs site; the agentic layer is undocumented; 5 comments reference files not in the repo |
| **Safe to merge?** | :warning: **Changes requested** | The blocking item makes the headline command non-functional |

### Findings

:red_circle: **Blocking** — `scripts/uat/run_uat.py:169,183` — **the self-deploy path cannot work.** `deploy_stack()` passes `HEADLESS=1` to both `make publish` and `make deploy`. In this repo `--headless` means *"no UI/AppSync/Cognito/WAF — for API-only or GovCloud deployments"* (`lib/idp_cli_pkg/idp_cli/cli.py:386`), and `lib/idp_sdk/idp_sdk/_core/template_transform.py:279` puts `ApplicationWebURL` in `outputs_to_remove` while `:95` strips `UserPool`. So `make uat-testing ADMIN_EMAIL=...` — the command in the CHANGELOG, `docs/testing-ui-acceptance.md:27`, and `make help` — spends 30–70 minutes deploying a stack with no Web UI, then dies at `stack_output(..., "ApplicationWebURL")` (line 244) and deletes it in the `finally`. Even if that output existed, `resolve_stack()` would then raise on the missing `UserPool`. Fix: drop `HEADLESS=1` from both commands. This is consistent with the PR description, which only claims validation "against a live APIGateway-hosted stack" — i.e. the `--stack-name` path.

:yellow_circle: **Should fix** — `scripts/uat/run_uat.py:166–187` — `--publish` is a ~40-minute no-op. The publish result is never captured into `template_url`, so the deploy falls through to the `else` branch and appends `FROM_CODE=1`, which builds and publishes all over again. Either capture the published template URL or drop the flag.

:yellow_circle: **Should fix** — `.github/workflows/ui-uat.yml:117–125` — Playwright artifacts can leak the UAT users' credentials into a **publicly downloadable** artifact. `playwright.config.ts` sets `trace: 'on-first-retry'` and `video: 'retain-on-failure'`; traces capture network response bodies (Cognito ID/access/refresh tokens) and `fill()` action parameters. The uploaded `playwright-report/` embeds those traces, and Actions artifacts on a public repo are downloadable by anyone who can view the run. Given 2/12 scenarios currently fail, traces *will* be produced. Suggest dropping `playwright-report/` from the artifact (keep the md/json/junit), or `trace: 'off'` in CI.

:yellow_circle: **Should fix** — `.github/workflows/ui-uat.yml:110` — template injection. `${{ inputs.grep }}` is interpolated straight into a `run:` block. `workflow_dispatch` is write-access-gated so severity is low, but this workflow holds secrets on a public repo and no other workflow here does this. Pass it via `env: GREP: ${{ inputs.grep }}` and reference `"$GREP"`.

:yellow_circle: **Should fix** — the suite ships with 2 of 12 scenarios known-failing (per the description: "10/12 scenarios pass"). Nothing marks them `test.fixme`/`skip` and the workflow has no `continue-on-error`, so the nightly job is red from the first run — which trains people to ignore it. Either fix them, annotate them, or make the job advisory explicitly.

:yellow_circle: **Should fix** — `make uat-lint` (`Makefile:468`) is not wired into `make lint`, `make lint-cicd`, or any CI job, so ~1,000 lines of new TypeScript get no automated typecheck or lint. It passes today; this is about keeping it that way. (For what it's worth, `src/ui`'s eslint `files: ['src/**/*']` and tsconfig `include: ['src/**/*']` correctly exclude `e2e/`, so there's no build breakage.)

:yellow_circle: **Should fix** — `scripts/generate-dep-manifest.sh:186–189` — `NODE_LOCKFILES` is an explicit list (`src/ui`, `docs-site`). The new `src/ui/e2e/package-lock.json` isn't in it, so its 6 packages sit outside the `dep_audit` gate the previous CHANGELOG entry just added. One-line fix.

:yellow_circle: **Should fix** — `scripts/uat/run_uat.py:157` — `new_stack_name()` is `idp-uat-{%m%d-%H%M%S}`, with no year and no random component. If it ever collides with a pre-existing stack, `make deploy` *updates* it and the `finally` block *deletes* it, even though the script didn't create it. You already use `secrets.token_hex(3)` for the user prefix (line 260) — reuse it here.

:yellow_circle: **Should fix** — `scripts/uat/agentic/flows/README.md:60` documents `ddb_item_exists` as a working verify method and the example YAML (line 39) uses it, but `verify.py` implements only `ddb_config_classes` and `ddb_document_status`, so it falls through to `no verifier implemented`. The `preconditions:` key at README line 41 is likewise not read by `flows.py`. Worth marking both as not-yet-implemented.

:yellow_circle: **Should fix** — `scripts/uat/agentic/flows.py:128–137` + `run_prototype.py:169` — a flow YAML with neither `docs` nor `docs_ref` never gets a `"docs"` key, so `spec["docs"]` raises `KeyError` under `--warm`. Use `spec.get("docs")`.

:yellow_circle: **Should fix** — five comments reference files that aren't in the repo: `agentic_uat_design.md` (`browser.py:11`) and `accelerator_issues.md` (`flows.py:59`, `03-configuration.spec.ts:11`, `documents.ts:60,103`). Either include them or inline the substance.

:yellow_circle: **Should fix** — `docs/testing-ui-acceptance.md` is missing the YAML frontmatter (`title:`) and license header that 87 of 89 files in `docs/` carry, and it isn't registered in `docs-site/astro.config.mjs`, so it won't appear on the published site.

:yellow_circle: **Should fix** — the PR claims (CHANGELOG, commit message, `Makefile:437`) that it "closes the open Browser/e2e UI test (Playwright) item in `scripts/sdlc/docs/CI_TEST_COVERAGE.md`", but that file isn't in the diff — line 1039 still reads `- [ ]`.

:green_circle: **Nice to have**

- `Makefile:432–435` — the UAT block was inserted between the "In-place stack upgrade validation (X→Y)" comment and its `stacktest-upgrade:` target, orphaning that comment.
- `.gitignore:45` says `make uat-lint` runs `npm ci`; it actually runs `npm install` (`Makefile:469`), as does `npm_setup()`. Only the workflow uses `npm ci`.
- `browser.py:295` tells the model to "Retry with exact=true", but the `click` tool exposed to the agent (`worker.py:215`) has no `exact` parameter.
- `browser.py:319,342` — `read_text` and the `fill` not-found branch skip `_capture()`, so those steps are missing from the filmstrip the report describes as stage-by-stage.
- `helpers/documents.ts:124` — the comment says the filename assertion confirms "the POST was issued", but Cloudscape `FileUpload` renders the filename token immediately on `setInputFiles`, so this passes regardless of whether the upload succeeded.
- `helpers/documents.ts:81` — matching status tokens as substrings of the whole row text means a filename or column value containing e.g. `OCR` would read as permanently in-flight.
- `04-viewer-role.spec.ts:41` — `waitForLoadState('networkidle')` is fragile against the UI's polling transport; prefer waiting on a concrete element.
- `verify.py:60` hardcodes `Config#default`, but `configuration_manager.resolve_active_version()` shows the active version can be something else, so the oracle would read stale ground truth on a stack with an activated version.
- `src/ui/e2e/` lands inside `src/ui`, which `publish.py`'s `compute_ui_hash()` checksums and `package_ui()` zips wholesale — so harness edits invalidate the UI build cache and the test code ships in the WebUI source zip. Both harmless in effect, but it partly undoes the stated reason for making this a separate package. A top-level `e2e/` would avoid it.
- `create-test-set.yaml:11` has a `TODO`.
- `run_uat.py:229` pre-flight checks `UAT_<ROLE>_USER` but not `_PASSWORD`, though the error message names both.

### Recommendation

**Changes requested.**

The design thinking here is the strongest part: the information barrier enforced by type rather than prompt wording, verification held strictly out-of-band, difficulty measured rather than self-reported, recording the stack name *before* the deploy so a partial deploy is still torn down, and the reasoning about why OIDC federation was avoided — all hard to retrofit later. The deterministic tier closes a real gap and the code is clean.

Before merge: the `HEADLESS=1` fix, plus the two workflow security items.

I'd also suggest splitting commits 2–3 into their own PR. The agentic prototype is genuinely interesting, but it's 1,350 lines with no CHANGELOG entry, no `docs/` page, no documented install path for its `strands`/`playwright` dependencies, and a README describing capabilities that aren't implemented yet — it will review better on its own, and it lets commit 1 land quickly once the blocking fix is in.

rstrahan (changes requested):

sromoam:
Closing for now, will revisit in the coming weeks.

PR #683, opened 2026-08-28
feat(ui): bulk document download from the document list (#679)

### LINKED ISSUE #679: [Feature]: Ability to download bulk documents from the accelerator through the UI

I want to be able to download the outputs of many documents extractions from the 'document list' view, and not have to go one at a time. 

I want an additional button in the document list after selecting multiple documents to 'download extractions' that would downlaod the extractions for every document selected. 

_No response_

_No response_

### Additional context

_No response_

rstrahan:
Closes #679.

Extends the existing per-document ZIP export to a multi-select bulk export driven from the document list, so collecting the extractions for a set of documents no longer means opening each one in turn.

## What this adds

The document list's **Download** menu now covers both the table itself and the artifacts of the selected rows. They're grouped separately because they differ in *scope* — the Excel export covers every filtered row, the ZIP exports cover only the selection:

```
Document list
  ⤓ Table as Excel (142 rows)
Selected documents (3)
  ⤓ All data (ZIP)
  ⤓ Predictions (ZIP)
  ⤓ Baselines (ZIP)
```

This replaces the list's previous icon-only download button (which exported the table to Excel and was labelled just "Download"). Folding both into one labelled menu avoids two similar-looking download affordances whose scopes differ invisibly. Trade-off: the Excel export is now two clicks instead of one.

A bulk archive contains one top-level folder per document, each laid out exactly like a single-document archive (`output/`, `baseline/`, `input/`, `document-attributes.json`, `manifest.json`), plus a root `manifest.json` indexing every document with its folder, file count, and error count.

## Implementation

**Exporter** (`document-export.ts`) — new `exportDocuments()` shares one `JSZip` across documents; `exportDocument()` becomes a single-document wrapper, so the detail page's archive layout, filenames, and manifests are unchanged (its 20 existing tests pass untouched). Three details worth review:

- **Key collisions.** `sanitizeDocumentKey` maps both `a/b.pdf` and `a_b.pdf` to `a_b.pdf`. Harmless for a one-document archive, but the second would have silently overwritten the first in a shared one — repeats now get a numeric suffix.
- **Concurrency.** Fetches were strictly serial; a 50-document export would have taken minutes of pure round-trip latency. They now run through a bounded worker pool (default 5) that starts thunks in order, so issue order stays deterministic.
- **Error attribution.** Soft errors are recorded per document and carry the document key, so one document with a permissions problem doesn't spoil the archive and the completion dialog names the document.

**UI** — selection is hydrated through `getDocument` before export, because list rows carry no `Sections`/`Pages`; if a detail fetch fails, that document still lands in the archive with its attributes rather than vanishing. Every bulk scope confirms first: the dialog states the document count, warns above 25 documents (the archive is assembled in browser memory), and offers the page-image / source-document toggles for the `all` scope. Progress reports documents alongside files (`24 of 60 files processed · 3 of 5 documents`) and **Cancel** stops at the next file.

## Testing

- **UI suite: 33 files, 328 tests, all passing** — 19 new: 9 bulk-exporter (multi-doc layout, root manifest, key collision, per-document error attribution, mixed baseline availability, aggregate progress, concurrency ceiling, abort, empty selection), 7 menu-builder (group labels, counts, disabled reasons, baseline gating), 3 header render tests (click dispatch per scope, and that disabled items don't fire).
- `eslint`, `tsc --noEmit`, `prettier --check` and a full `vite build` are clean.
- **Not exercised against a live stack.** The presign + fetch path is unchanged from the shipped single-document export, but the bulk flow itself hasn't been run in a browser against real S3.

## Notes for reviewers

- **No role gate.** The bulk export is available to any user who can see the list, matching the ungated per-document download. Worth a conscious yes/no, since it makes pulling many documents' extractions a one-click action.
- Frontend-only: no template, IAM, or backend changes.

Docs: `docs/web-ui.md` gains a "Bulk download from the document list" section; CHANGELOG updated under `[Unreleased]`.


rstrahan:
Thanks — both 🟡 findings were real, and 🟡1 in particular was a good catch. Pushed e421be5 addressing all seven.

## 🟡 1 — Baselines 403: confirmed, fixed, and verified live

Your read was correct, and I verified both halves against the live `us-west-2` stack rather than by inspection alone:

```
$ aws iam simulate-principal-policy --policy-source-arn <IDP1-CognitoAuthorizedRole> \
    --action-names s3:GetObject --resource-arns arn:aws:s3:::idp1-evaluationbaselinebucket-.../any/key.json
  implicitDeny
$ ... --resource-arns arn:aws:s3:::idp1-outputbucket-.../any/key.json
  allowed
```

So the client-signed baseline URL was always going to 403. And the resolver route does work — invoked the live `GetFileContentsResolverFunction` with `fieldName: getFilePresignedUrl` on a real baseline object, then fetched the URL it returned:

```
StatusCode: 200 → {presignedUrl, contentType, size}
$ curl "$url"  →  http_status=200 bytes=2514  (valid baseline result.json)
```

**Fix:** presigning is now bucket-aware. `presignForExport()` signs Output/Input in-browser (no extra round trip, and those grants exist) and routes everything else — baseline, unknown buckets, and the case where `settings` hasn't loaded — through `getFilePresignedUrl`, exactly as `SectionsPanel.handleDownload` does.

I went with the resolver rather than adding the baseline bucket to `CognitoAuthorizedRole` for two reasons: the IAM route would hand every authenticated user blanket read on ground-truth data, and it would only take effect after a stack update, whereas the resolver already holds `S3ReadPolicy` on that bucket behind its own allow-list, so this fix works on already-deployed stacks. Worth noting this was **broken in the shipped single-document download too** (0.6.5), not just the new bulk path — so it's now a `### Fixed` CHANGELOG entry, and `docs/web-ui.md` documents the split signing route instead of claiming everything is signed client-side.

5 new tests cover the routing, including the resolver-returns-no-URL case and the mixed archive where predictions sign in-browser while baselines go through the resolver.

## 🟡 2 — Silent hydration failures and unbounded fan-out: both fixed

- **Bounded**: hydration now runs in chunks of `HYDRATION_CHUNK_SIZE = 5` (matching the exporter's fetch concurrency) instead of one `getDocumentDetailsFromIds(allKeys)` call. I chose sequential chunks over per-key thunks through the existing pool deliberately: each call triggers `setDocumentsDeduped`, so per-key would mean ~100 list re-renders where chunking means ~20.
- **Visible**: your suggested `!detailByKey.has(row.objectKey)` check is in. Misses become `preflightErrors` — a new `ExportOptions` field — which the exporter folds into progress events, the returned errors, the root manifest's `errors[]`, the owning document's `errorCount`, *and* that document's own `manifest.json`. So the completion dialog names the document and the archive itself no longer reads as complete when it isn't. Two tests pin the plumbing, including that the failure appears in the affected document's manifest and nowhere else.

## 🟢 3–7

- **3** — `isBaselineAvailable` is now exported from `document-export.ts` and used by all three call sites.
- **4** — `controller.signal.aborted` is checked between hydration rounds, so Cancel takes effect there too; the button relabels to "Cancelling…" and disables while the in-flight step drains.
- **5** — Progress now covers hydration: `ExportProgress.phase` is `'preparing' | 'exporting'`, and the modal reads "3 of 12 documents loaded" during hydration rather than sitting at 0% describing files it isn't processing yet.
- **6** — Good point, un-coupled: the dropdown is no longer disabled wholesale. Only the ZIP items are held (`disabledReason: 'A download is already in progress'`), so the Excel export stays available. Test added.
- **7** — Header switched to Apache-2.0 to match the directory.

## On your two answers

Agreed on both, and thanks for the reasoning on the role gate — "the identity role already grants blanket GetObject on input/output to every authenticated user, so a UI gate would be theatre" is the right frame, and the `simulate-principal-policy` output above independently confirms it. Left ungated, no hard cap added.

**Validation:** 336 UI tests across 34 files pass (8 new here, 27 total for the feature), plus `eslint`, `tsc --noEmit`, `prettier --check` and a full `vite build`. Still no end-to-end browser run against a live stack — but the specific failure that caveat was hiding is now closed at the API level with the evidence above.

Issue #676, opened 2026-08-28
[Bug]: BDAOCRProject blocks every unified-pattern deployment to GovCloud — no Condition, no parameter to disable


karolytanczos-govcio:
### Summary

`patterns/unified/template.yaml` creates a `Custom::BDAOCRProject` at deploy time with no
`Condition` and no parameter to disable it. GovCloud Bedrock Data Automation rejects the
project it creates, so `PATTERNSTACK` fails and the root stack rolls back — regardless of
deploy mode (`--govcloud` or `--headless`) and regardless of the configured `ocr.backend`.

### Version

Present on `develop`, and in `v0.6.4` / `v0.6.5`.

### The failure

```
Received response status [FAILED] from custom resource. Message returned:
An error occurred (ValidationException) when calling the CreateDataAutomationProject
operation: Sync project does not support video/audio/document modality in
Standard Output Configuration
```

`PATTERNSTACK` reports `The following resource(s) failed to create: [BDAOCRProject]`, and the
root stack rolls back with ~30 further resources cancelled.

### Availability is not capability

BDA *is* available in `us-gov-west-1` — the same account holds LIVE data-automation projects
with full `standardOutputConfiguration.document` blocks (DOCUMENT/PAGE granularity, bounding
boxes, markdown output). What GovCloud does not support is this **sync** project with
document modality in the standard output configuration.

So this is not "BDA is missing in GovCloud"; it is one project shape that is unsupported.

### Why it is unavoidable

```yaml
BDAOCRProject:
  Type: Custom::BDAOCRProject
  Properties:
    ServiceToken: !GetAtt BDAOCRProjectFunction.Arn
```

No `Condition`, and no template parameter gates it. `--headless` does not remove it either —
a headless deploy reaches the same failure, it simply fails earlier on a parameter mismatch
in some configurations and never gets there.

### The intended behaviour is already documented in the template

The consuming environment variable carries this comment:

```yaml
# ARN of the stack-scoped BDA OCR project (empty in regions without
# Bedrock Data Automation; the 'bda' backend then errors clearly).
BDA_OCR_PROJECT_ARN: !GetAtt BDAOCRProject.ProjectArn
```

Graceful degradation for BDA-less regions is clearly intended — but the unconditional
resource makes that path unreachable. The OCR service already handles the empty case
(`ocr/service.py` resolves `bda_project_arn` from config or the env var, and the `bda`
[...]

PR #684, opened 2026-08-28
fix(headless): strip SuppressAdminInvite, which broke every headless deploy

rstrahan:
Found while investigating an aside in #676: *"`--headless` … simply fails earlier on a **parameter mismatch** in some configurations and never gets there."* @karolytanczos-govcio was right that something else was wrong. It is worse than "some configurations" — it is **all** of them.

## The bug

`b301a1fa7` (2026-07-16) added a `SuppressAdminInvite` condition that tests `AdminEmail` against a CI sentinel to suppress the Cognito invite email. The headless transform removes the `AdminEmail` **parameter** but was never taught to remove that condition.

A `Condition` referencing a deleted parameter is not dead weight CloudFormation ignores — it rejects the entire template up front:

```
Template format error: Unresolved dependencies [AdminEmail].
Cannot reference resources in the Conditions block of the template
```

**So `--headless` has been broken for six weeks, failing before creating a single resource.** That's the "fails earlier" symptom in #676.

Verified against the real CloudFormation `ValidateTemplate` API in both directions:

- the offending shape → `ValidationError: Unresolved dependencies [AdminEmail]`
- after the fix → the transformed template's **full** `Parameters` (61) + `Conditions` (34) blocks validate clean

## The fix

One line: `SuppressAdminInvite` joins `conditions_to_remove`.

It is safe because the condition was **already inert** in headless — its only consumer is `AdminUser`'s `MessageAction`, which the same transform removes, and I confirmed **no resource is gated on it** (`grep -c 'Condition: SuppressAdminInvite'` → 0). So this is a pure unblock with no behavioural change. `GovCloudTemplateTransformer` is untouched: it keeps the full UI, so `AdminEmail` and the condition both remain valid there.

## Why CI missed it for six weeks — and the two gates that close it

This is the more important half of the PR.

The headless template *is* validated during `publish --headless`, but only by a **live CloudFormation call that no CI job makes**. And the one offline gate that lints it, `test_real_template_headless_survives_govcloud_region_cfn_lint`, deliberately discards every finding except `E3006` ("resource type absent from region") as out of scope — so the `E1020` for this removed-parameter reference was **filtered out by design**.

Two offline, credential-free gates now close that hole:

1. `test_headless_transform_leaves_no_unresolved_parameter_reference` — structural: for every parameter the transform removes, assert no surviving `Ref`/`Fn::Sub` points at it. Covers **Conditions, Rules and Outputs**, not just Resources, which is what made this bug invisible. Needs neither cfn-lint nor credentials.
2. `test_real_template_headless_has_no_cfn_lint_errors` — asserts **zero Error-level** cfn-lint findings on the transformed real template. Warnings stay ungated on purpose: headless legitimately leaves unused conditions and unreachable `Fn::If` branches behind (11 of them today), so gating warnings would be flaky while gating errors is precise. Measured: 1 error before this fix, 0 after.

Both fail against the pre-fix transformer (verified by stashing the source): 2 failed / 7 passed.

## Testing

331 passed in `lib/idp_sdk/tests/unit`; `make fastlint` clean; `typecheck_pr_changes` 0 errors.

⚠️ Not deployed end-to-end — we have no GovCloud access right now, and this affects `--headless` in **every** region, not just GovCloud. The evidence is the real `ValidateTemplate` result above plus the two new gates. A commercial `--headless` deploy would exercise it fully if someone wants belt-and-braces before release.

## Relationship to the other GovCloud PRs

Independent of #681 and #682 — different root cause, different code path. It does mean **#676 needed two fixes to close for headless users**: the BDA partition gate (#681) and this. Merges cleanly with both except for the usual `CHANGELOG.md` conflict at the top of `### Fixed` (keep all entries); the `template_transform.py` edits are in a different class from #682's and auto-merge.


rstrahan:
## Added: real CloudFormation validation of the headless template in CI

The original PR fixed the bug and added two **offline** gates. This adds the missing **integration-tier** gate, closing an item that was already an open TODO in `scripts/sdlc/docs/CI_TEST_COVERAGE.md`:

> - [ ] **`--headless` template-transform smoke.** At minimum, run the `HeadlessTemplateTransformer` + cfn-lint in the fast gate to catch transform breakage without a deploy.

`validate_headless_template` transforms the **packaged** `.aws-sam/idp-main.yaml` and validates it through real CloudFormation `ValidateTemplate`.

**Why this and not a headless deploy probe.** Three reasons it's the right cost/value point:

- It **reuses the template the publish step already built**, so it costs one S3 put + one `ValidateTemplate` call — not a second ~1h SAM build.
- Validating the **packaged** artifact closes a gap the offline tests structurally cannot: they transform the *source* `template.yaml`, which differs from what publish uploads (SAM expansion, nested-stack URLs).
- It is **exactly the check `publish --headless` already performs** — no CI job ever invoked it.

**Failure handling.** It never raises, and its result is folded into the verdict at the *end* rather than short-circuiting, so a bad headless template doesn't cost us the primary suite's signal. Applied at the verdict specifically because `stack_success` is *assigned* (not and-ed) by the primary suite, so setting it False early would be silently overwritten.

**Scope, stated plainly.** This proves the headless template is well-formed and every reference resolves. It does **not** prove a headless stack stands up. A real deploy probe would need the Probe framework to accept a template *path* (today every probe deploys the same template with different parameters) plus ~1h — left as a tracked follow-up in the coverage doc.

**Tests:** 8 new, no AWS and no subprocess. The `ValidateTemplate` rejection is asserted as a failure using the **verbatim `AdminEmail` error** from this bug; the packaged template is reused rather than rebuilt; the artifact lands beside the main template with bucket/prefix/region derived from the published URL; and all four failure paths (missing template, transform failure, boto3 raise, unparseable URL) return rather than raise. `254 passed` in `scripts/sdlc/tests`, lint and typecheck clean.

**Also corrected in the docs:** the note claiming the `jobsapi` probe covers the transform now explains that it was once *named* the "headless" probe (hence the leftover `-headless` stack names in the reaper fixtures) and was renamed precisely to stop implying that coverage.

### Coverage after this PR

| Tier | Gate | Runs |
|---|---|---|
| Fast (offline, no AWS) | structural dangling-parameter check | every MR |
| Fast (offline, no AWS) | cfn-lint zero-`Error` on transformed template | every MR |
| Integration (real AWS) | `ValidateTemplate` on the **packaged** headless template | develop + MRs targeting develop |
| — | headless stack **deploy** | ❌ still not covered (tracked) |

PR #690, opened 2026-08-29
feat(config): Configuration Profiles with revision history

### LINKED ISSUE #689: Enhancement: split Configuration Versions into Configuration Profiles + Revisions (config history, RBAC scoping, Test Studio pinning)
## Summary

`Configuration Version` currently does three orthogonal jobs at once, and the resulting overload is confusing for users and awkward for RBAC:

| Axis | What it actually is | Who should control it |
|---|---|---|
| **Identity / tenancy** | The RBAC object (`allowedConfigVersions`), the document-visibility partition, the target of a test set's `boundConfigVersion` | Admin |
| **Lineage** | "the same thing, later" — a prompt iteration, a model swap, an upgrade-driven change | Author, within scope |
| **Selection** | What new documents process under (`IsActive`), what a test run uses | Author, within scope |

Because lineage has no home, users express it by **minting new identities** (`usecaseA_v1`, `usecaseA_v2`, …). Each such name is a new RBAC object, a new document-visibility partition, a new confidence-curve bucket, and a new row an Admin must add to every scoped user's `allowedConfigVersions`.

## Problems this causes today

- **A scoped Author cannot iterate without destroying history.** `saveAsVersion` is Admin-only (`nested/api-resolvers/src/lambda/configuration_resolver/index.py:272`) while a plain `updateConfiguration` on an in-scope version *is* permitted (`:253`). So an Author's only move is an in-place overwrite — no undo, no diff-against-yesterday, no rollback.
- **An Author scoped to use case A cannot create or test a new iteration of A** without an Admin minting the name for them, even though it is entirely within their remit.
- **Lineage-as-identity already happens automatically.** `apply_feature_config_preset` creates versions like `sample-health-insurance-review-v0.1.6`; every feature upgrade adds a name Admins must re-scope users onto.
- **Stack upgrades are destructive or ignored.** Managed versions are overwritten in place; custom versions are never updated and the only remedy is "Reset to Default", which clobbers all customization. There is no way to see *what* an upgrade changed.
- **Terminology is unrecoverable.** "Version" already means *immutable numbered snapshot* for [document versions](docs/document-versions.md) and test set versions. Config is the odd one out, where "version" means the named entity itself — hence the nonsense phrase "versions of a config version".

## Proposal

Split the concept in two:

- **Configuration Profile** — the named entity that exists today (`default`, `Production`, `lending`). The RBAC object, the document partition, the activation target. Admin-created.
- **Revision** (`r7`) — an immutable numbered snapshot of one profile's configuration, cut on every save. Content, not an access-control object. Author-created within their scope.
- **Configuration** keeps its current meaning: the *content* (OCR/classification/extraction settings, classes, prompts).

The invariant: **revisions are content, profiles are access-control objects. Authors move content; only Admins mint objects.**

The design principle that keeps it cheap: **the profile head stays the only thing the runtime resolves by default**, so the ~189 files referencing `config_version` / `ConfigVersion` are unaffected, and `allowedConfigVersions` / `ConfigVersion` / `--config-version` keep working unchanged.

Full design, ERD, exact keys, permission-matrix delta, file-by-file change list, and rejected alternatives: **`docs/proposals/config-profiles-and-revisions.md`** (added by the first PR).

## Why this fixes the RBAC problem

Largely **without a permission-model change**. The only reason `saveAsVersion` is Admin-gated today is to stop history being destroyed; once every save cuts a revision, an in-scope Author save is inherently non-destructive. An Author scoped to `lending` can then iterate, diff, roll back, and (in a later phase) test revisions of `lending` — while remaining unable to see or touch any other profile.

## Additional wins

- Stack upgrades **cut** `default` r(n+1) instead of overwriting → the upgrade becomes a diff you can read, and rollback is publishing the previous revision.
- "Reset to Default" stops being a blunt clobber.
- Feature presets cut a revision of one managed profile instead of minting `…-v0.1.6`, `…-v0.1.7`, … names.
- Test Studio can pin `(profile, revision)` alongside its existing test-set version pin, making a metric delta attributable to *config revision* vs *ground truth* vs noise.

## Decisions taken

| Decision | Choice |
|---|---|
| Entity name | **Configuration Profile** (bare "Profile" inside config UI; API identifiers always qualified `configProfile*`, since `profile` already means *user* profile via `getMyProfile`) |
| Revision body storage | S3 bodies in `ConfigurationBucket` + tiny DynamoDB pointer items, so revision blobs never inflate a `ConfigurationTable` scan |
| Retention | Cap per profile (last N) + always keep published, labeled, and test-run-pinned revisions |

## Phasing

- **Phase 0** — terminology + docs (no behavior change).
- **Phase 1** — revisions cut on save, history panel (view/compare/restore/label/delete), retention pruning, managed-profile revisions on upgrade. Self-contained: runtime still uses the profile head.
- **Phase 2** — revision *selection*: `Document.config_revision` through the pipeline, Test Studio pinning and cross-revision compare, confidence-curve fingerprinting, CLI/SDK flags.
- **Phase 3** — profile *families* (group profiles under a use case, scope by family). Only if Phases 0–2 leave real demand.

**Phases 0 + 1 are in scope for the first PR.** Phases 2 and 3 to follow as separate issues.

## Riding along in the first PR

Two independent fixes on lines this work already touches:

- **Fail-closed document scope filter** — `list_documents_gsi_resolver/index.py:306` currently passes documents with **no** `ConfigVersion` to scoped users (fails open), so a scoped user sees unstamped documents.
- **Prefix-glob `allowedConfigVersions`** — allow entries like `lending-*` as an interim stopgap for deployments already using `usecaseA_v1/v2/v3` naming.

## Out of scope

`getDocument`, Knowledge Base queries, and Athena analytics remain unscoped by config version — already tracked under Known Limitations in [docs/rbac.md](docs/rbac.md). This work neither fixes nor worsens them.

rstrahan:
Closes #689 (Phases 0 and 1).

Splits "configuration version" into a **Configuration Profile** (the named, access-controlled entity) and its **Revisions** (immutable numbered snapshots of that profile's configuration), so keeping a previous configuration no longer requires minting a new profile name.

Design doc, ERD, rejected alternatives and the Phase 2/3 plan: **`docs/proposals/config-profiles-and-revisions.md`**.

## Why

`Config#<name>` was doing three orthogonal jobs: identity/tenancy (the RBAC object, the document partition), lineage ("the same thing, later"), and selection (what runs now). Because lineage had no home, users expressed it by minting new identities (`usecaseA_v1`, `usecaseA_v2`, …) — and each new name is a new access-control object an admin must grant to every scoped user.

The sharpest consequence: **a scoped Author could not iterate at all.** `saveAsVersion` is Admin-only (`configuration_resolver/index.py:272`) while a plain in-scope `updateConfiguration` is permitted, so their only move was a destructive overwrite — no undo, no diff-against-yesterday, no rollback.

## What changed

- **Every save cuts a revision.** A new **Revision history** panel per profile lists them newest-first with who saved it and when; you can compare any two (reusing the existing field-level diff, including its word-level inline rendering), restore, label, or — as an admin — delete.
- **Restore is forward-only.** The restored configuration is saved as a *new* revision, so the state it replaced stays inspectable. History is never rewritten.
- **Stack deployments cut a revision of `default`** instead of overwriting it silently, so an upgrade's configuration changes are diffable and rollback is one restore. A save that changes nothing records nothing, so repeated deployments cannot push real history out of the retention window.
- **Terminology**, in UI and docs only: profile = the named entity, revision = a snapshot of one. **No API, CLI, or stored field was renamed** — `getConfigVersions`, `versionName`, `--config-version`, `ConfigVersion`, `allowedConfigVersions` all still refer to profiles. `docs/configuration-versions.md` remains as a stub so existing links keep working.

## RBAC

The invariant: **revisions are content, profiles are access-control objects. Authors move content; only Admins mint objects.**

| Operation | Admin | Author (in scope) | Viewer (in scope) | Reviewer |
|---|---|---|---|---|
| View / compare revisions | ✅ | ✅ | ✅ | ❌ |
| Restore / label a revision | ✅ | ✅ | ❌ | ❌ |
| Delete a revision | ✅ | ❌ | ❌ | ❌ |
| Create a profile (`saveAsVersion`) | ✅ | ❌ | ❌ | ❌ |

Scope is checked at the **profile** for all five operations — a revision is never its own RBAC object, so there is one rule to get right rather than five. Each operation is also gated server-side in `_OPERATION_REQUIRED_GROUPS` (defence in depth) and registered in `scripts/api_rbac_expectations.yaml`.

## Riding along (authorized side fixes)

- 🔒 **Fail-closed document scope.** Both document-list resolvers and the document-chat processor admitted a document with **no** `ConfigVersion`, so every document processed before config-version stamping was visible to — and chattable by — every scoped user. All three now fail closed. ⚠️ **Scoped users will see fewer documents after this upgrade**; anything that disappears was unstamped, and reprocessing it under a profile in their scope restores visibility. Admins and unscoped users are unaffected.
- **Glob patterns in `allowedConfigVersions`** (`lending-*`), for deployments that already encode iterations in profile names. An interim convenience; revisions make it unnecessary for new deployments.
- **One shared scope matcher** (`idp_common/config_scope.py`). The two document-list resolvers carry no `idp_common` layer (hottest UI query, kept dependency-free), so they vendor it verbatim with a **drift test** — a scope matcher that differs per call site is exactly how the fail-open gap arose.
- **Active-profile resolution no longer scans** the configuration table on every queued document; activation writes a pointer item read with one `get_item`, with the paginated scan as fallback. `__active` is now a reserved profile name (and refused as a profile name, so nobody can claim the sentinel).

## Storage

| Item | Key | Holds |
|---|---|---|
| Profile head | `Config#<profile>` | unchanged working config + `LatestRevision` / `PublishedRevision` |
| Revision index | `ConfigRevIndex#<profile>` | one small entry per retained revision |
| Revision body | `config_revisions/<profile>/<nnnnnn>.json.gz` | the configuration that revision recorded |
| Active pointer | `Config#__active` | the active profile name |

Bodies live in S3 because `ConfigurationTable` is HASH-only, so listing profiles requires a `Scan`, and DynamoDB bills a scan on **full item size regardless of `ProjectionExpression`** — revision blobs in the table would make the profile list more expensive with every save. Metadata is one index item per profile so listing history is a single `get_item`; appends use native `list_append` (cannot lose a concurrent append) and revision numbers are allocated with an atomic `ADD`, so two simultaneous saves get distinct numbers. `ConfigRevIndex#` deliberately does not match `begins_with(Configuration, "Config#")`, so a revision can never leak into the scope-filtered profile dropdowns.

History is **best-effort**: if a revision cannot be recorded the save still succeeds (logged at WARNING). Losing a history entry is recoverable; refusing a save is an outage. With no `CONFIGURATION_BUCKET` configured the store is disabled and everything behaves as before.

Retention: last `CONFIG_REVISION_CAP` (default 20) per profile, plus the published revision and anything labeled or pinned by a test run.

## Migration

No backfill, no format change. On a profile's first save after the upgrade, the configuration that was there **before** that save is cut as its first revision, so enabling history does not lose the state history exists to protect. Profiles never saved after the upgrade simply show no revisions.

## Verification

**Unit/lint/typecheck**
- Full battery (`make test`, 54 roots): **5,435 passed**, 3 failed — all `test_z3_orchestrator.py`, all `No module named 'z3'`, confirmed identical on pristine `develop` in a worktree. 0 new regressions.
- New tests: 32 for revision lifecycle (moto, real DynamoDB+S3 semantics), 26 for the scope matcher, 29 for the resolver operations (group gate, profile scope, reserved names, argument validation), 4 for vendored-copy drift.
- `make fastlint` clean; `basedpyright` clean on every new module (the 4 remaining repo-wide errors are in files this PR does not touch, 2 reproduced on `develop`).
- `make api-test-static`: 0 FAIL (the 2 WARN are the pre-existing documented GAP-01/GAP-02).

**Live deploy test** — published and `update-stack`'d onto a real stack in us-west-2, `UPDATE_COMPLETE`, then exercised against the **deployed** Lambdas:
- The upgrade cut `default` r1 (pre-history snapshot) + r2 (deployed config) and the same for each managed profile; bodies landed in `config_revisions/`, index items in the table, and neither showed up in the profile list.
- **The diff between those two revisions was real and previously invisible**: 144 differing leaf paths, including ~24 extraction field renames (`NumberofAutoAccidents` → `Number of Auto Accidents`). Before this change there was no way to see that an upgrade had done that.
- Author (non-Admin) restore and label succeeded; Reviewer was refused server-side; the published revision refused deletion; Admin delete of an older revision removed both the index entry and the S3 body; deleting a profile removed its whole history.
- A restore was verified to change the head configuration and to land as a new revision with the earlier one intact.
- Two documents processed end-to-end to `COMPLETED` (6 pages each), one before the active pointer existed (scan fallback) and one after (pointer path), both stamped `ConfigVersion=default`.
- All test artifacts were cleaned up afterwards.

## Not in this PR

- **Phase 2** — revision *selection*: `Document.config_revision` through the pipeline, Test Studio run pinning and cross-revision comparison, confidence-curve fingerprinting, CLI/SDK flags.
- **The feature-preset conversion.** `apply_feature_config_preset` still mints `<feature>-v<semver>` profile names. It writes raw sparse rows directly with `put_item`, has no `idp_common` layer or bucket access, and its version names are load-bearing for feature hook registration — converting it needs its own migration story. Tracked in §10 of the proposal.
- `getDocument`, Knowledge Base queries and Athena analytics remain unscoped by config version (pre-existing, documented in `docs/rbac.md`).


rstrahan:
## Second deploy-test pass (validates the no-op suppression)

The `fix(config): a save that changes nothing records no revision` commit came out of the *first* deploy test, so the stack was re-published and updated a second time to verify it against the deployed artifact.

`IDPRel066` (us-west-2) → **`UPDATE_COMPLETE`**, all five nested stacks complete.

`UpdateDefaultConfig` ran at 16:15:30 in that second deployment. Revision state afterwards, read back through the **deployed** resolver:

```
revisions of default after TWO deployments:
  r2 published=True  by=stack-deployment  notes='Updated by stack deployment'
  r1 published=False by=system            notes='Configuration as it stood before revision history was enabled'
```

Counters stayed at `latest=r2 published=r2` for `default` and all four managed profiles (`docsplit`, `fake-w2`, `ocr-benchmark`, `realkie-fcc-verified`), and no new object appeared under `config_revisions/default/`. Without the fix, that deployment would have added an identical r3 to every one of those five profiles — which is how repeated no-op upgrades would have pushed a user's real history out of the 20-revision retention window.

Everything in this PR has now been exercised on a live stack.

rstrahan:
## UX corrections (3rd commit)

Reviewing the shipped panel surfaced four things worth fixing before merge:

- **"Revision History" was already taken.** `VisualEditorModal.tsx:1381` and `GroundTruthVisualEditor.tsx:610` use that exact label for the edit history of an extracted/labeled value. The panel is now **Configuration revisions**. Grepping `version` collisions is what motivated this whole change; not grepping `revision` nearly reintroduced the same overloading one level down.
- **Two compare buttons read identically** but compared different axes — now **Compare profiles** (across profiles) and **Compare revisions** (within one).
- **History was only reachable from the profiles table**, so a user who had just saved something they regret had to leave the editor to undo it. Added *Actions → Configuration revisions…* to the editor's own action bar.
- **The diff rendered below the revisions table**, so a long history had to be scrolled past to reach it. It now opens in its own modal, matching `DocumentVersionsPanel`.

UI lint, `tsc --noEmit`, and the 344 UI tests are green.

PR #692, opened 2026-08-29
feat(config): select and pin a Configuration Profile revision (Phase 2)

### LINKED ISSUE #691: Enhancement: select and pin a Configuration Profile revision (Phase 2 — processing, Test Studio, pickers)
Follow-up to #689 (Phase 2 of `docs/proposals/config-profiles-and-revisions.md`). Depends on #690, which delivers Phases 0 and 1.

## What Phase 1 left out

#690 makes every configuration save non-destructive: a Configuration Profile now keeps immutable numbered **revisions** you can view, compare, restore, label, and delete. But the runtime still always resolves the profile **head**, so a revision cannot be *selected*:

- **Test Studio picks a profile only.** `TestRunner.tsx:139` sends `configVersion: <profile name>`; there is no revision in the picker, the run payload, or the run record. So you cannot run one test against `lending@r5` and another against `lending@r7` — which is the single most valuable thing revisions could buy, because it is how you tell "this prompt change helped" from "the ground truth moved".
- **Every other picker is the same** — Upload Documents, Reprocess, Generate draft labels, Discovery, Multi-doc Discovery, Capacity Planning. All select a profile; all get whatever revision is current.
- The only workarounds today are ugly: restore r5 → run → restore r7 → run (mutates the profile between runs, so concurrent processing sees whichever is current), or Admin `Save as Version` to fork a revision into a new profile — which is exactly the identity proliferation the design exists to remove.

## Scope

**Resolution.** `Document.config_revision` alongside `config_version`, and a `revision` parameter on `get_config()` / `ConfigurationManager.get_configuration()`. Resolution order: explicit revision → the profile's `PublishedRevision` → the head.

Deliberately **not** a composite `lending@7` identifier: that string is equality-compared against `allowedConfigVersions` in the document-list and configuration resolvers, so an unparsed composite would either bypass or break scope. `ConfigVersion` stays the profile name; the revision travels as its own field.

**Pinning at queue time.** The queue sender stamps the resolved revision on the document, so a save mid-flight cannot change the configuration under an in-flight document — the same reasoning that made config-version pinning necessary in #599.

**Test Studio.** Pin `(profile, revision)` on the run record alongside the existing test-set `activeReference` pin, mark the revision so retention cannot prune it out from under a retained comparison, and surface the revision in the run list, results, and comparison view. This completes the three-way attribution the test-set design already aspires to: a metric delta becomes attributable to *config revision* vs *ground truth* vs noise.

**Confidence curves.** `curve_store` keys curves per config version today because "confidence means different things across models and prompts". Keying per revision would cold-start the curve on every prompt tweak; keying per profile silently mixes incomparable semantics across a model swap. Key by `(profile, confidence-relevant fingerprint)` instead, so revisions that do not touch the model/assessment configuration keep accumulating observations while a model swap branches the curve.

**Pickers.** A revision selector next to each profile selector, defaulting to the current revision and offering older ones.

**CLI/SDK.** `--config-revision` where `--config-version` is accepted.

## Out of scope

- Phase 3 (profile *families* — grouping profiles under a use case and scoping by family). Revisit only if Phases 0-2 leave real demand.
- The `apply_feature_config_preset` conversion (still mints `<feature>-v<semver>` profile names) — see §10 of the proposal for why it needs its own migration story.

rstrahan:
Closes #691. **Stacked on #690** — this PR targets `feature/config-profiles-and-revisions`, not `develop`, so the diff shows only Phase 2. Retarget to `develop` once #690 merges.

Phase 1 made configuration revisions *recordable*. This makes them *selectable*: you can now run one test against `lending` r5 and another against r7, which is the comparison revisions exist to make possible.

## The gap this closes

Everything ran under whatever the profile held at the time. `TestRunner.tsx` sent `configVersion: <profile name>` and nothing else, so a run recorded only *which profile* it used — and a later save silently changed what that name meant. The workarounds were restore-then-run (mutating the profile between runs) or forking a revision into a new profile, which is exactly the identity proliferation the design removes.

## Selection

`Document.config_revision` travels alongside `config_version` through S3 object metadata (`config-revision`), the SQS body, the Step Functions input, and the tracking table (`ConfigRevision`). `get_config(version=…, revision=…)` loads that revision's stored body; all 16 pipeline call sites pass it.

Two deliberate decisions:

- **No composite `lending@7` identifier.** That string is equality-compared against `allowedConfigVersions` in the document-list and configuration resolvers, so an unparsed composite would either bypass or break scope. The profile name stays the profile name; the revision is its own field.
- **A missing pinned revision raises rather than falling back to the head.** A run that silently used the wrong configuration *looks successful*, and its numbers then enter a comparison — worse than a failed run. Retention protects any revision a run pinned, so this only happens after an explicit delete.

## Pinning at queue time

The queue processor — the single chokepoint every execution passes through — stamps the profile's current revision on every document, whether or not a revision was chosen. Without it, a save made while a document is in flight changes the configuration underneath it: extraction under r7, assessment under r8, and a result corresponding to no single configuration. Same reasoning that made config-version pinning necessary in #599.

A profile with no revision history pins nothing and consumers fall back to the head, which is exactly the pre-revision behavior.

## Test Studio

A run now records the pinned revision alongside the test-set version it already pinned, which makes a metric delta attributable instead of ambiguous:

| Both runs | A difference means |
|---|---|
| Same test-set version, different config revision | the configuration change moved it |
| Same config revision, different test-set version | the ground truth moved |
| Both differ | nothing conclusive — re-run holding one fixed |

Capturing a revision also marks it **exempt from retention pruning**, because a comparison is only readable while both runs' configurations still exist and the default window is 20 revisions.

## UI

One shared `ConfigRevisionSelector` sits beside the profile picker in **Test Executions**, **Upload Documents**, **Reprocess**, and **Generate draft labels**. It defaults to the profile's current revision, labels it (`Current (r7)`), and **renders nothing when the profile has no history** — a dropdown whose only entry is "Current" is noise, not a choice, and that is the state of most deployments until someone saves a configuration. Changing profile clears a stale revision selection, since r5 of `lending` has nothing to do with r5 of `claims`.

The revision is displayed next to the profile everywhere it already appeared: document list, test-run list, test results, test comparison, and CSV/JSON exports.

## CLI/SDK

`--config-revision 7` on `process` / `run-inference`, threaded through `BatchProcessor` and stamped as object metadata. The SDK's `batch.process(config_revision=…)` mirrors it.

## Confidence curves: consciously NOT re-keyed

The proposal calls for keying curves by `(profile, confidence fingerprint)` so a prompt tweak keeps accumulating observations while a model swap branches. I did not wire it, because **partial wiring is worse than none**: the three call sites are a scoring run (which knows the pinned revision), a reviewer's save (which knows only the profile), and the review-effort estimate (likewise). Fingerprinting only the site that has the information would write scoring observations to `curve#p@fp` while review observations kept landing on `curve#p`, so the estimate — which reads one key — would stop seeing scoring data. That is a regression against today, where both land together.

What is here instead: every revision records a **confidence fingerprint** (a hash of the extraction model/sampling plus assessment configuration), so a follow-up can branch curves correctly and cheaply. The interim guidance — reset the affected curve after a model or assessment change, using the existing `CurveStore.reset` — is documented in `docs/test-studio.md` and `docs/configuration-profiles.md`.

## Verification

- Full battery (`make test`, 55 roots): **5,456 passed**, 3 failed — all `test_z3_orchestrator.py`, all `No module named 'z3'`, confirmed identical on pristine `develop`. 0 new regressions.
- New tests: 7 for pinned-revision resolution (raises on a missing revision, `get_config` passthrough, published-revision resolution), 4 for queue-time pinning (including that an unusable value degrades to no pin rather than breaking the workflow start), 7 for test-run pinning and the retention mark, 8 for the revision picker's render-nothing cases, 2 for the confidence fingerprint.
- `make fastlint`, UI lint, `tsc --noEmit`, 352 UI tests, `make api-test-static` (0 FAIL) all clean; API validation spec regenerated.
- Live deploy test in progress on the same stack as #690 — results posted as a comment.

## Not in this PR

- Phase 3 (profile *families*).
- `apply_feature_config_preset` still mints `<feature>-v<semver>` names — see §10 of the proposal.
- Discovery panels deliberately have no revision picker: they *save* a discovered schema into a profile, and saving into an old revision is not a meaningful operation.
- Capacity Planning has no revision picker (it estimates from a configuration; add if asked).


rstrahan:
## Live deploy test — and two real defects it caught

Published and `update-stack`'d onto the same us-west-2 stack as #690 (three cycles, each `UPDATE_COMPLETE`). Two defects surfaced that no unit test could have caught; both are fixed in this PR with regression tests that fail on the pre-fix code.

### Defect 1 — the test runner had no `idp_common` layer

`_capture_config` gained a `ConfigurationManager` import to read a pinned revision's configuration and mark it exempt from retention pruning. `TestRunnerFunction` carries no `idp_common` layer, so the import raised `ImportError`, the surrounding `except Exception` swallowed it, and **both features silently no-op'd**: the run recorded `ConfigRevision=1` while capturing the *wrong* configuration body, and `r1` stayed prunable (`pinned=false`) despite a run depending on it.

The unit tests mock `ConfigurationManager`, so they passed either way. Added `scripts/sdlc/tests/test_resolver_layer_coverage.py`, which compares each function's source imports against the layers its template declares — it fails on the pre-fix template and passes on the fixed one, and it also asserts the two deliberately layer-free document-list resolvers stay that way.

### Defect 2 — captured revision configs are floats, DynamoDB wants Decimal

With the layer in place, **every** `startTestRun` that pinned a revision failed:

```
TypeError: Float types are not supported. Use Decimal types instead.
  index.py:576 in _store_test_run_metadata -> table.put_item(Item=item)
```

A revision body is JSON, so it carries Python floats (`temperature: 0.0`, thresholds), and it is written straight into the run's DynamoDB item. The pre-existing path never hit this because it read the config *from* DynamoDB, which returns `Decimal` already. Fixed on capture, with a test asserting the captured config contains no float values.

### Verification after both fixes

**Two documents, same profile, different revisions, concurrently** — from the OCR function's own log:

```
Loaded configuration profile 'revsel-test' revision r2   <- unpinned upload
Loaded configuration profile 'revsel-test' revision r1   <- uploaded with config-revision=1
```

Both reached `COMPLETED` (6 pages each) with `ConfigVersion=revsel-test` and `ConfigRevision=1` / `2` on their tracking records. The unpinned document was pinned to r2 automatically at queue time — the profile's current revision — which is the protection against a save landing mid-flight.

**Two test executions of one test set against two revisions of one profile** — the thing this PR exists for:

```
bench-tiny_form-20260829-191721: profile=revsel-test r=1 captured_notes='REVISION ONE'
bench-tiny_form-20260829-191728: profile=revsel-test r=2 captured_notes='REVISION TWO'
```

Each run recorded its own revision *and* captured that revision's actual configuration, so the comparison view has the real inputs for both sides.

**Retention pins** after those runs: `r1 pinned=true`, `r2 pinned=true` — neither can be pruned out from under a retained comparison. Deleting the published revision is still refused (`Cannot delete r2: it is the current configuration`).

All live test artifacts (scratch profile with its revisions and S3 bodies, three test runs, two documents and their outputs, tracking rows) were cleaned up afterwards.

rstrahan:
## 4th commit — one vocabulary, and Phase 3 closed

Follow-up agreed after reviewing the shipped feature. No functional change.

- **"Revision" is no longer overloaded.** The extraction-value editors labelled their field-level edit log **"Revision History"** — sitting right beside Configuration Profile pickers, while a revision is an immutable numbered snapshot of a configuration. That tab is now **Edit history**, which is what its own `_editHistory` data and `EditHistoryTab` component already called it.
- **Documents and test sets keep their `versions`.** I checked the cost of renaming for symmetry: 6 public GraphQL operations, ~7 field names across 37 files, and stored DynamoDB sort keys (`version#000001`) — a data migration plus a breaking API change, for no user benefit. The collision that hurt was *inside one noun* ("config version" meaning the container and its lineage at once), not across nouns.
- **The whole vocabulary is now in one table** in `docs/configuration-profiles.md`, linked from `document-versions.md` and `test-studio.md`, with the rule that keeps it stable: the noun that owns the snapshot picks the word.
- **Phase 3 (profile families) closed** in the proposal, with the two observable signals that would reopen it, so it stops reading like committed work.

Battery re-run: **5,536 passed**, same 3 pre-existing `z3` environment failures. UI lint, `tsc`, and 352 UI tests green.

Issue #713, opened 2026-08-31
[Bug]: evaluation _convert_to_rich_values auto-unwraps any single-key result, which can misfire


rstrahan:
`idp_common/evaluation/service.py` `_convert_to_rich_values` contains a wrapper-detection heuristic: if the result has a **single non-metadata key** whose value has >= 2 confidence children, it treats that key as a wrapper and unwraps it.

That is a guess about intent based on shape alone. Any extraction whose class schema legitimately has one top-level object property will be silently unwrapped, changing the key paths the evaluation compares against the baseline.

Noticed while assessing what would happen to downstream consumers if extraction output were ever wrapped in a list — the heuristic would misfire on a genuine wrapper too, in confusing ways. It predates that work and is a latent issue on its own.

Suggested: gate the unwrap on something explicit (a declared marker, or the class schema saying so) rather than inferring it from key count.

rstrahan:
Fixed in #725 (merged to `develop` as `4728f00bb`).

Verified before/after: the new tests fail without the source change and pass with it — evidence is in the individual commit messages. `make test` (57 roots), `make lint-cicd`, typecheck and the UI suite (379 tests) all green on the combination, not just individually.

Issue #712, opened 2026-08-31
[Bug]: listDocumentsByDateRange appears to be dead code in the UI


rstrahan:
`listDocumentsByDateRange` / `ListDocumentsByDateRange` has no consumers in `src/ui/src` outside the generated operations and one test:

```
grep -rl 'listDocumentsByDateRange|ListDocumentsByDateRange' src/ui/src --include='*.ts' --include='*.tsx' \
  | grep -v 'graphql/generated|graphql/operations|__tests__'    # no matches
```

The document list uses `listDocuments`. The operation, its `.graphql` selection set, and the backend resolver are all still maintained — including being updated whenever a `Section` field is added — for no consumer.

Either it is intended for external API users (in which case it is worth documenting as such, since it is currently indistinguishable from dead UI code) or it can be retired.

rstrahan:
Investigated as part of #725. **Recommend closing as invalid — it is not dead code.**

`listDocumentsByDateRange` is unreferenced in `src/ui/src` (the narrow claim here is correct), but it is a **live API operation** repo-wide. Searched the whole repo, excluding `node_modules`/`.git`, for `listDocumentsByDateRange`, `DocumentsByDateRange`, `documents_by_date_range` and `list_documents_range`:

| Where | What |
|---|---|
| `nested/api-resolvers/src/api/schema.graphql` | schema definition |
| `nested/api-resolvers/src/lambda/list_documents_range_resolver/index.py` | resolver, with reviewer + config-version scope filtering |
| `nested/api-resolvers/template.yaml` | function + wiring |
| `.../http_api_dispatcher/api_validation_spec.json` | dispatcher validation |
| `scripts/api_rbac_expectations.yaml` | a dynamic authorization case **invoked with real arguments** by `make api-test` |
| `docs/rbac.md` + **six** `security/test-results/*/rbac-dynamic.md` | published security results |

Deleting it would have removed a tested authorization surface and invalidated six published security result sets.

The issue's secondary suggestion — "worth documenting as such" — was valid, so #725 adds a header comment to `ListDocumentsByDateRange.graphql` recording that it has no UI consumer on purpose.

[...]

Issue #711, opened 2026-08-31
[Bug]: Neither UI Section interface derives from the generated GraphQL types, so a new field is silently dropped


rstrahan:
There are two hand-written Section shapes in the UI and **neither** derives from `src/ui/src/graphql/generated/schema-types.ts`:

- `src/ui/src/types/documents.ts` -> `Section`
- `src/ui/src/components/sections-panel/SectionsPanel.tsx` -> a second, **local** `SectionItem` (this is the one the table actually uses)

So adding a field to the GraphQL `Section` type and regenerating changes nothing about whether the UI can read it, and nothing fails if both interfaces are missed. This is the same silent-breakage class that previously dropped confidence alerts from the historical-version view.

Evidence that it drifts: `types/documents.ts` was already missing `Excluded`/`ExclusionReason` before this was noticed.

Mitigation added meanwhile: guards over the four generated operation texts assert the field is present in each query. But the type-level gap remains — deriving both from the generated types (e.g. `Pick<>` over the generated `Section`) would make it a compile error instead of a runtime blank.

rstrahan:
Fixed in #725 (merged to `develop` as `4728f00bb`).

Verified before/after: the new tests fail without the source change and pass with it — evidence is in the individual commit messages. `make test` (57 roots), `make lint-cicd`, typecheck and the UI suite (379 tests) all green on the combination, not just individually.

Issue #706, opened 2026-08-31
[Bug]: Document.metadata is never serialized, so metadata['failed_page_exceptions'] is silently dropped between Lambdas


rstrahan:
`Document` has a `metadata` field, but it is absent from both `Document.to_dict` and `Document.from_dict` (`idp_common/models.py`).

Since `serialize_document` and the compression path both go through `to_dict`, anything in `metadata` exists only for the lifetime of a single Lambda invocation and never crosses a Step Functions hop. `metadata["failed_page_exceptions"]` is written expecting to be read later, and never is.

Verified: `"metadata"` does not appear in `Document.to_dict`, while the dataclass field does exist.

This is the same class of defect as the per-page `document_boundary` signal, which was being stashed via `setattr(page, "metadata", ...)` — an attribute that is not a dataclass field and therefore never survived serialization either. That one was fixed by adding a declared field; this needs the same treatment, or the writes should be removed if the data genuinely isn't needed downstream.

rstrahan:
Fixed in #725 (merged to `develop` as `4728f00bb`).

Verified before/after: the new tests fail without the source change and pass with it — evidence is in the individual commit messages. `make test` (57 roots), `make lint-cicd`, typecheck and the UI suite (379 tests) all green on the combination, not just individually.

PR #725, opened 2026-09-02
Wave 1: six silent-failure bugs, each with a guard against recurrence

rstrahan:
Five independent bugs from the #694 review's side findings. Every one is the same
shape: something written by one stage, then quietly dropped or mis-read before any
consumer saw it — so nothing failed, and the wrong answer looked like a normal one.

Each was developed in an isolated worktree, and for each I verified the new tests
**fail without the source change and pass with it** — a test that passes either way
proves nothing. Evidence is in the individual commit messages.

## What is fixed

| Issue | Symptom |
|---|---|
| [#706](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/706) | `Document.metadata` was in neither `to_dict` nor `from_dict`, so `metadata["failed_page_exceptions"]` never crossed a Step Functions hop |
| [#704](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/704) | Section `Excluded`/`ExclusionReason` were never persisted, so the "Skipped" badge could never render |
| [#705](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/705) | The document-name-regex short-circuit ignored `sectionSplitting` — the same defect as #686 |
| [#713](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/713) | A class with a single top-level group scored with **no confidence at all**, reported as "nothing to measure" |
| [#711](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/711) | Neither UI `Section` shape derived from the generated GraphQL types, so a new backend field was invisible to the UI |

## Three findings worth reading

**#704 was worse than "not written".** Extraction calls `update_document_section`
for *every* section including excluded ones, and that write is
`SET #Sections[i] = :section` — a whole-map replace. Omitting the keys was
**erasing** the flags classification had already persisted, so even adding a writer
elsewhere first would have left the badge flickering and vanishing.

**#706 could have turned a dropped field into a hard failure.**
`classification/service.py` stores a live `Exception` in `metadata`. Including it
verbatim would put a non-JSON-serializable object into `serialize_document`'s
uncompressed branch, which hands `to_dict()` straight to Lambda's response
serializer. Hence the explicit coercion, with a test asserting the payload survives
a plain `json.dumps`.

**#711's guard found three real drifts already on `develop`**, now fixed:
`GetDocumentVersion` was missing `Excluded`/`ExclusionReason`; `OnUpdateDocument`
and `ListDocumentsByDateRange` were missing `ProcessingIssues`.

## #712 investigated and NOT actioned — recommend closing as invalid

`listDocumentsByDateRange` is unreferenced in `src/ui/src`, but it is a live API
operation repo-wide: schema, resolver, template wiring, dispatcher validation
spec, an RBAC authorization case invoked with real arguments by `make api-test`,
`docs/rbac.md`, and six published `security/test-results/*/rbac-dynamic.md`.
Deleting it would have removed a tested authorization surface. A header comment now
records that it has no UI consumer on purpose.

## Guards added (the durable part)

- **#711**: `SECTION_FIELDS` is `Record<keyof GqlSection, true>`, so a new schema
  field is a **compile error**, plus a drift test over every `Sections { … }`
  selection set. I verified both independently by injecting a probe field into
  `schema.graphql`, regenerating, and confirming `tsc` errors *and* the test fails
  — then reverting and confirming zero traces remain.
- **#705**: the three-way `sectionSplitting` rule is factored into one shared
  predicate with #686, so the two paths cannot diverge again.
- `make test` no longer counts `.claude/worktrees/` as unregistered test roots —
  five concurrent agent worktrees made the whole gate fail for a reason unrelated
  to the code.

## Cost note

`sectionSplitting: llm_determined` is the default, so a **name-matched multi-page**
document now makes classification inference calls where it made none. Single-page
matches keep the zero-inference path; `disabled` or `page` is the explicit opt-out.
Same shape as the #686 note already in the CHANGELOG.

## Verification

- `make test` — **all 57 test roots pass**
- `make lint-cicd` — clean, including GraphQL codegen freshness
- `typecheck_pr_changes.py --base develop` — 0 errors
- UI: `npx vitest run` 39 files / 379 tests; `tsc --noEmit` and `npm run build` clean

## Deliberately left for later

- `maxPagesForClassification` still overrides `sectionSplitting` for every config
  (`_apply_limited_classification_to_all_pages` collapses to one all-pages
  section) — pre-existing, orthogonal, flagged in a code comment, worth its own
  issue.
- `Document.metadata` is still not persisted to DynamoDB; #706 is scoped to the
  Step Functions hop and a tracking-table attribute is a separate design decision.
- ⚠️ **#713 caveat:** affected classes reported *no* calibration before and report
  real numbers now, so calibration metrics on previously recorded runs are
  non-comparable for those classes. Accuracy and field scores stay comparable.


rstrahan:
### Added since the PR was opened: #708 + #707

The fifth workstream landed. Two more items, both verified independently rather than taken on report.

**#708 — a retired Bedrock model is no longer offered.** `us.anthropic.claude-3-5-haiku-20241022-v1:0` is end-of-life but was still selectable. Worked the `add-model` checklist in reverse across the template enums, `pricing.yaml`, the bedrock client's cachepoint set, `update_configuration`'s EU mapping, and both doc tiers.

The interesting finding is that hard-remove-vs-deprecate was a **false dichotomy**: every model field in `IDPConfig` is a plain `str`, and the ConfigSchema `enum` drives the UI dropdown only — nothing revalidates a stored config against it. So removing it from the enum *is* "accepted-but-hidden"; there was no deprecation shim to build.

I verified the one real consequence myself, because it is a deployment-risk claim. `merge_utils._load_valid_bedrock_models()` builds `valid_models` from `pricing.yaml`, so `config-validate` now reports a config pinning it as invalid. Measured on the merged branch:

```
stored config still LOADS at runtime : True     <- no deployment breaks
config-validate on retired model     : valid=False
    -> extraction.model has invalid model ID: us.anthropic.claude-3-5-haiku-...
control, current model               : valid=True   <- the check is not always-false
```

And I confirmed the blast radius by grep: `validate_config` is reachable only from `idp-cli config-validate` and the SDK's `validate()` — **not** from the stack-update custom resource and **not** from the config save path. So no deployment can wedge on it. That is the loud-signal outcome #708 asked for.

**#707 — a parity guard between `IDPConfig` and the CFN-embedded UI schema.** 336 config fields vs 279 UI paths, reconciled by an allowlist so every gap is a deliberate, written decision instead of an invisible one. It extends `test_config_schema_order.py`'s loader rather than adding a second one, which preserves the "does not leak constructors onto the global `yaml.SafeLoader`" property by construction — and that loader is now covered by `test_cfn_loader_safety.py`'s tripwire, which it wasn't before.

I proved it fails on a real gap by injecting a field into `ExtractionConfig` and re-running:

```
E  extraction.parent_probe_knob  (declared on ExtractionConfig)
FAILED test_every_config_field_has_a_ui_control_or_a_written_reason
1 failed, 10 passed
```
Probe reverted; tree confirmed clean.

The 55 exceptions are split into three *distinct verdicts*, which is the part worth reviewing: `NOT_USER_SETTABLE` (28 — derived, machine-written, or runtime plumbing), `EXEMPT_SUBTREES` (5), and **`KNOWN_UI_GAPS` (22 — "there arguably should be a control and there isn't")**. That last list is now a concrete, shrinkable backlog rather than an unknown; it includes the entire confidence self-healing ladder, which is on by default and YAML-only. `test_no_stale_exceptions` makes the lists self-cleaning.

**Also found, not fixed** (reported so they aren't lost):
- `discovery.output_format.sample_json` is a **dead knob** — rendered by the UI, present in `base-discovery.yaml`, but `DiscoveryConfig` has no such field and the sample is hardcoded in code. The shipped default is also stale relative to the hardcoded one.
- The repo's own CHANGELOG states `claude-3-7-sonnet` was retired by Bedrock, yet it is still in `KnowledgeBaseModelId`, `pricing.yaml` and the EU map — same open question for two more Claude 3.x IDs. Not removed (unverified live, outside #708 scope). Adding one is now a one-line entry in `RETIRED_MODEL_IDS`.
- `_validate_model_ids` only checks ocr/classification/extraction/assessment/summarization, so a retired model in `chat`, `evaluation.llm_method`, `discovery.*` or `agents.*` still slips through `config-validate`.

**Re-verified on all six fixes together:** `make test` all 57 roots, `make lint-cicd` clean, typecheck 0 errors.

PR #731, opened 2026-09-02
feat(classification): real class confidence, on by default, and the measurement that justifies it (#673)

### LINKED ISSUE #673: [Feature]: Add confidence scores for document classification

Currently, document classification returns `confidence: 1.0` (hardcoded) regardless of how confident the LLM actually is in the classification decision. This contrasts with the extraction workflow which uses the Assessment service to return meaningful confidence scores (0-1) for each extracted field.

Users cannot distinguish between a high-confidence classification (clear W-2 form) vs. a low-confidence classification (ambiguous form that could be multiple types).

Have the classification LLM return a confidence score (0-1) for its classification decision, similar to how extraction assessment works:

1. __Modify classification prompt__ to request confidence score alongside the classification result
2. __Parse confidence from LLM response__ in `ClassificationService._classify_page()`
3. __Populate `page.confidence`__ with actual LLM-provided score
4. __Optional: Add secondary classification__ with confidence (e.g., "80% Form W-2, 15% Form XXXX")

_No response_

_No response_

### Additional context

This would enable:

- Automated escalation of low-confidence classifications to human review
- Better analytics on classification accuracy

rstrahan:
Closes #673.

Classification pinned `confidence: 1.0` on every page and section, so a clear W-2
and an ambiguous form that could have been three things were indistinguishable.
Two values were already being produced and thrown away, and the fix splits into
"stop discarding what we have" and "ask for something better, then check whether
it means anything".

## Commit 1 — keep what the model already reports

- **`confidence`** was documented in `docs/classification.md` (with example
  output) and parsed by nothing.
- **`classification_reason`** is asked for by the **default** prompt, so its
  output tokens were bought on every page of every document and dropped.

Both are now parsed and threaded through to persistence, the API and the UI.

**The semantic decision this rests on: `None` means NOT SCORED, and is never
serialized as a number.** `Page.confidence` defaulted to a fabricated `0.0`/`1.0`
and `Section.confidence` to `1.0`, and that value reached the reporting lake's
long-standing `section_confidence` column where it was indistinguishable from a
measurement. Now:

| Situation | Confidence |
|---|---|
| Model returned one | the model's value |
| `mode: off`, or a custom prompt that asks for none | *not scored* (null) |
| **BDA mode**, blueprint matched | BDA's matched-blueprint confidence |
| Name-regex / single-class / page-content-regex / human correction | `1.0` — a real assertion |
| Invalid class replaced by `invalidClassFallback` | *not scored* (the class is ours, not the model's) |
| Pages beyond `maxPagesForClassification` | *not scored* (class extrapolated, not predicted) |
| SageMaker/UDOP backend | *not scored* (endpoint returns none) |
| Classification errored | `0.0` |

Section confidence is the **minimum** across its pages, and unscored if any page
is: a mean hides the one page the classifier was unsure about, which is the page
a reviewer needs.

**BDA mode is scored immediately and for free** — BDA's matched-blueprint
confidence *is* the classification confidence there (the blueprint decides the
class) and was already read for the HITL decision, just never stored.

Two defects fixed on the way:

- `update_document_section` replaces the whole section map, so it now writes
  `Confidence` — otherwise extraction erased what classification had stored
  moments earlier (the trap `ProcessingIssues` already documents).
- The classification cache-hit branch dropped `document_boundary` entirely; both
  branches now share `_apply_page_result`.

Plumbing mirrors the `document_boundary` commit (d6a85b1a4): declared fields on
`Page`, both DynamoDB page writers plus the read path, the run snapshot, GraphQL
`Page.ClassConfidence` / `ClassReason` / `ClassCandidates` and
`Section.Confidence`, the versions-resolver allow-list, all four operations that
select Pages/Sections, and the Pages/Sections tables in the UI.

## Commit 2 — top-K, and the measurement that justifies it

`classification.confidence.mode: topk | verbalized | off` — **default `topk`**.

`topk` is the recommended mode and the shape the issue asked for: ranked
candidate classes with probabilities ("80 % W-2, 15 % 1099"). A single
self-reported number is ~0.95 on everything, which is worse than no score because
it invites automated escalation on noise; making the model enumerate and rank
alternatives forces it to distribute probability mass instead (Tian et al., *Just
Ask for Calibration*, EMNLP 2023 — the same reasoning behind extraction's
`G1`/`P1` path). The runner-ups are stored per page and listed in the UI's "Why
this class?" popover.

Resolution decisions worth reviewing:

- A page is scored on the probability of the class **actually stored**, not the
  top of the list — they differ when the model's `class` contradicts its own
  ranking, and the top probability would then describe a class the page was never
  given. A stored class absent from its own candidate list leaves the page
  unscored rather than inferring a number from the leftover mass.
- Out-of-vocabulary candidates are dropped: a class that cannot be stored cannot
  be reported.
- Probabilities are rescaled **only** when they sum to more than 1.0. A
  distribution cannot exceed 1, whereas a sum below 1 legitimately means
  "possibly some other class".
- `top_k` is clamped to `[2, len(classes)]`.

The instruction blocks are editable config, spliced **before** the
document/cache-point marker so they stay inside the prompt-cache prefix — the same
rule as extraction's bbox block, and it matters more here because classification
runs per page. Composed only for `multimodalPageLevelClassification`; the holistic
method (segment list, not one object per page) warns and is left to ask in its own
prompt, which commit 1 already parses.

**Measurement — the actual gate on the default, and on ever acting on these numbers.** The evaluation
report's per-page rows now carry `predicted_confidence` beside the `correct` flag
they already had, and the benchmark harness turns the pair into
`class_calibration_separation` = mean(conf | right) − mean(conf | wrong), with
`class_accuracy` and `n_class_scored_pages` beside it. `aggregate.py --compare`
flags a −0.03 move as a regression, same as the field-level separation.
Near-zero separation means the model is equally confident when it is right and
when it is wrong; the docs say so, and say not to route work on it.

## Commit 3 — measured it, then turned it on

Ran this on a live stack (`IDPBench066`, us-west-2) against `docsplit` — RVL-CDIP-derived
packets over 13 deliberately confusable classes — 298 pages per model per mode,
sequential runs, no failures. Full paper:
[`docs/benchmarking/classification-confidence.md`](docs/benchmarking/classification-confidence.md);
raw per-page data in `benchmarks/results/v0.6.7/clsconf/`.

**Does asking cost anything?** (`off` vs `topk`, same documents)

| | Nova 2 Lite (default) | Claude Haiku 4.5 |
|---|---|---|
| Class accuracy, off → topk | 0.832 → 0.846 (+0.013) | 0.859 → 0.852 (−0.007) |
| Classification step cost | +17.5 % | +14.9 % |
| Classification as share of total doc cost | **3 %** | 16 % |
| **Total document cost / page** | $0.03072 → **$0.03062 (−0.3 %)** | $0.03426 → $0.03474 (+1.4 %) |

Accuracy: no systematic effect — opposite signs, both inside single-run noise. Cost:
+17.5 % of a step that is 3 % of the bill is ~+0.5 % of total, which is why the measured
total came out *negative*. **That is why `topk` is now the default** rather than opt-in.

**Is the score worth having?** (both in `topk`)

| | Nova 2 Lite | Claude Haiku 4.5 |
|---|---|---|
| Calibration separation | +0.044 | **+0.207** |
| Mean confidence when **wrong** | 0.903 | 0.741 |
| Distinct confidence values (298 pages) | **6** | 11 |
| Errors caught / pages reviewed | 43 % / 8 % @0.90 | **73 % / 11 % @0.85** (94 % precision) |

Nova 2 Lite answered **exactly 0.95 for 269 of 298 pages**, including 23 of its 46 errors
— a coarse two-level flag, not a probability, and no threshold reaches half its mistakes.
Haiku 4.5 grades properly. So the default makes a score *available* everywhere; it does
not make it equally *informative* everywhere, and the docs say exactly that.

**Side finding worth a follow-up (not fixed here):** Haiku 4.5 reported **zero** prompt-cache
reads or writes on this path while Nova served ~934 input tokens/page from cache. Haiku *is*
in `CACHEPOINT_SUPPORTED_MODELS` and the prompt does carry `<<CACHEPOINT>>`, so the probable
cause is a cacheable prefix below Anthropic's minimum — probable cause, **not** a diagnosis.

## Compatibility

- ⚠️ **`section_confidence` in the reporting lake changes from a constant `1.0` to a real
  value.** A query that filtered or averaged on it will see different results — the old
  numbers were not measurements.
- ⚠️ **Classification costs ~0.5 % more per document** on the default classifier, because
  every page is now asked for its ranked candidates. `classification.confidence.mode: off`
  restores the previous behaviour exactly (the prompt is untouched, not merely cheaper), and
  it is settable in the config editor, not just YAML.
- Serialized output is byte-identical for anything unscored (keys omitted, as
  `document_boundary` does), so in-flight documents and historical payloads round-trip
  unchanged.
- BDA mode was already scored for free and is unaffected by the mode.

## Verification

- 4775 tests green (idp_common unit + SDLC gates), 379 UI tests, 83 benchmark tests.
- `make lint` clean; `typecheck_pr_changes.py` clean (2 warnings, both confirmed
  pre-existing on `develop`); static API RBAC scan 0 FAIL.
- The config-UI parity gate caught the new block having no control **twice** — once when it
  was added, and again when flipping the default removed its exemption. Both times it named
  the exact missing fields.
- End-to-end on the live stack: `ClassConfidence` / `ClassReason` / `ClassCandidates` persist
  to DynamoDB, come back through the API, and `predicted_confidence` reaches the evaluation
  report.
- **Not run:** the full GitLab CI pipeline (deliberate — GitHub runs a subset today).

## Deliberately not in scope

Confidence-driven **escalation** (routing low-confidence pages to review) is deferred
indefinitely. The measurement above is also the argument for not shipping it blind: the right
threshold is model-specific (0.90 Nova, 0.85 Haiku on this corpus) and on the default
classifier it would reach only 43 % of errors. The score, the runner-ups and the calibration
metric are all in place for whoever wants to build it deliberately.


rstrahan:
## Benchmark: the numbers are in, and they change the guidance

Ran this branch on a live stack (`IDPBench066`, us-west-2) against the `docsplit`
test set — RVL-CDIP-derived packets over 13 deliberately confusable classes — in
`topk` mode, **298 pages per model**, sequential runs, no failures. Full paper:
[`docs/benchmarking/classification-confidence.md`](docs/benchmarking/classification-confidence.md);
raw per-page data in `benchmarks/results/v0.6.7.dev1/clsconf/`.

**The two classifiers are equally accurate and not remotely equally calibrated.**

| | Nova 2 Lite (default) | Claude Haiku 4.5 |
|---|---|---|
| Page classification accuracy | 0.846 | 0.852 |
| **Calibration separation** | **+0.044** | **+0.207** |
| Mean confidence when **wrong** | 0.903 | 0.741 |
| Distinct confidence values (298 pages) | **6** | 11 |
| Errors caught / pages reviewed, best threshold | 43 % / 8 % @0.90 | **73 % / 11 % @0.85** (94 % precision) |
| Classification cost / page | $0.00090 | $0.00573 |

Nova 2 Lite answered **exactly 0.95 for 269 of 298 pages**, including 23 of its 46
errors — so no threshold reaches half its mistakes. That is the failure mode the
design warned about, now measured. Top-K did pull 22 pages onto 0.85 (18 genuinely
wrong), so ranked candidates help; they do not rescue it. Haiku 4.5 produced a
graded 0.50–0.98 distribution and supports real review routing.

**What this changes in the PR:** nothing in the code — it validates the two
decisions the design turned on. `off` stays the default (the score is not free and,
on the default model, not very informative), and the measurement pipeline is the
thing that makes the choice decidable rather than a matter of taste.
`docs/classification.md` now states these numbers instead of warning in the
abstract.

**End-to-end verified on the live stack** (not just unit tests): per-page
`ClassConfidence`, `ClassReason` and `ClassCandidates` persist to DynamoDB, the
ranked alternatives come back through the API, and the evaluation report carries
`predicted_confidence` per page, which is what the harness scores.

**One side finding worth a follow-up:** Haiku 4.5 reported **zero** prompt-cache
reads or writes on this path, while Nova served ~934 input tokens/page from cache.
Haiku 4.5 *is* in `CACHEPOINT_SUPPORTED_MODELS` and the prompt does carry
`<<CACHEPOINT>>` (with the confidence block deliberately spliced *before* it), so the
probable cause is a cacheable prefix below Anthropic's minimum for that model —
reported as a probable cause, not a diagnosis. Consequence: on a small class
vocabulary, don't assume caching softens the cost of an Anthropic classifier; the
6.4× above is what you pay.

**Limits, stated plainly:** one run per model, one corpus, 20 packets. The
0.7-point accuracy gap is *not* a claim of difference. The calibration gap (4.7×,
plus a 269-page spike at a single value) is far larger than plausible single-run
noise, but it is still one sample of one corpus — which is exactly why the harness
prints `class_calibration_separation` for you to re-measure on your own documents.

PR #743, opened 2026-09-02
feat(ui): give class confidence its own sortable column (#673)

rstrahan:
Follow-up to #731 (issue #673). The confidence score shipped as plain text appended to the
class name — `invoice 95.0%` in one cell — which put two unrelated values in one place and
gave no hint that the percentage was clickable.

## What changes

Both document-detail tables get a **Class conf.** column beside `Class/Type`:

| | Presentation | Why |
|---|---|---|
| **Document Sections** | neutral badge | A section score is an aggregate (the minimum across its pages) with no reasoning of its own. Deliberately **not** a link: a trigger that opens nothing is worse than plain text. |
| **Document Pages** | link-styled percentage → **"Why this class?"** | The number is the affordance for the model's reasoning and its ranked alternatives. Uses Cloudscape's `Popover triggerType="text"` — link styling, keyboard focus and aria for free — matching the `95% margin` column in Test Studio, replacing the hand-rolled dotted underline. |

```
DOCUMENT PAGES
┌─────────┬──────────────┬─────────────┬───────────┐
│ Page ID │ Class/Type   │ Class conf. │ Thumbnail │
├─────────┼──────────────┼─────────────┼───────────┤
│ 1       │ invoice      │ 95.0%       │  [img]    │  ← link: reason + candidates
│ 2       │ form         │ 62.0%       │  [img]    │  ← badge: scored, nothing behind it
│ 3       │ handwritten  │ —           │  [img]    │  ← not scored
└─────────┴──────────────┴─────────────┴───────────┘
```

**Sorting is the functional half, and it needed enabling.** Both tables were
`sortingDisabled`, so the `sortingField`s they already declared never rendered a control.
Now:

- sorting starts **unset**, so the default view stays in page order (page order is
  meaningful here);
- the table renders a **sorted copy** — `tableItems` also feeds the page text editor's
  next/previous navigation, which must stay in document order however the table is sorted;
- unscored rows sort **last in both directions**, because sorting least-confident-first is
  the point and a wall of `—` at the top would defeat the click.

Side effect, stated deliberately: the columns that already declared `sortingField`
(Page ID, Class/Type, Section ID, Page IDs, Instances) become sortable too. That is what
those declarations were for.

## Two restraints, both choices

- **No colour banding at any value.** There is no configured classification threshold —
  extraction fields have `hitl.confidence_threshold`, the class has nothing — so
  green/amber/red would assert a pass/fail the product has not defined. It would also
  mislead concretely: in the #731 benchmark the default classifier answered exactly 0.95
  for **269 of 298 pages**, including half its errors, so banding would paint a coarse
  two-level flag as a calibrated traffic light.
- **`—` for unscored**, never `0%` (absence of a measurement is not a measurement of zero)
  and never blank (which reads as a rendering bug in a column with values in every other
  row).

## Also

- The adjacent per-field column is reworded **"Low Confidence Fields" → "Low-conf.
  fields"**. Two neighbouring columns both reading "confidence" for different measurements
  is the confusion this layout exists to resolve, not create.
- One incidental hunk: `ruff format` normalised two blank lines in
  `tests/unit/bedrock/test_sizing.py`, which had drifted on `develop`. Not mine, but
  `make lint-cicd` needs it.

## Verification

- 408 UI tests pass, including **14 new** ones pinning the presentation contract: own
  column, the label and the score no longer sharing a cell, clickable only when there is
  something behind the number, `—` for unscored, identical styling at high and low values,
  and the sort control existing.
- `make lint` clean (eslint + prettier + tsc + codegen check).
- Not visually confirmed in a browser — the panels are behind Cognito auth on a deployed
  stack, so the presentation is pinned by tests rather than a screenshot.


rstrahan:
## Review (self-review — read it as such)

Reviewed against the repo's `pr-review` checklist. Worth stating plainly: this is my own
PR, so this is a weaker signal than an independent review. The findings below are the ones
it actually caught, and they were real.

### 🔴 Two defects found, both fixed in c03d20f0

**1. Descending sort dragged every unscored row to the top.** The panel sorted ascending
and reversed the array, which inverts the comparator's nulls-last rule along with the
ordering — so the PR body's claim "unscored rows sort last in both directions" was false in
one of the two directions.

**2. Columns declaring `sortingField` sorted nothing.** The hand-rolled sort only applied a
column's `sortingComparator`, so Page ID / Class/Type / Section ID / Page IDs rendered a
sort control that did nothing. That is *worse* than the `sortingDisabled` it replaced,
because it looks functional. This also falsified the "those columns become sortable too"
line in the original description.

A second attempt at (1) — swapping the comparator's arguments instead of reversing —
reproduced it in a new form: with arguments swapped, "null goes after" reads as "null goes
before". Two failed attempts was the signal to stop hand-rolling and use `useCollection`
from `@cloudscape-design/collection-hooks`, already used by `TestResultsList` and
`DocumentPickerModal`. It honours both `sortingField` and `sortingComparator` and owns the
direction.

**Accepted wart, now documented and pinned instead of claimed away:** `useCollection`
negates the comparator for descending, so unscored rows lead in that one direction. No
single ordering keeps nulls last both ways. Ascending — the direction that matters, since
least-confident-first *is* the review workflow — puts them last, and the descending
behaviour matches every other Cloudscape table with missing values.

### 🟡 Two minor findings, fixed

- `NotScored` was exported but used only inside its own module — needless API surface.
- `variant` was bolted onto the call signature (`Props & { variant?: … }`) instead of
  declared in the props interface.

### ✅ Checks that came back clean

- **Target branch** `develop` ✅.
- **No index-keyed row logic** in either panel (`grep` for `items[`, `.map((x, i)`), so
  enabling sorting cannot corrupt edit-mode state — every handler keys on `item.Id`.
- **The page text editor still gets document order**: `pages={tableItems}`
  (`PagesPanel.tsx:643`), not the sorted copy, so next/previous navigation is unaffected by
  how the table is sorted. Same for the Sections download/reprocess helpers.
- **Design system only** — Cloudscape `Badge`/`Popover`/`Box`, no ad-hoc styling; the
  clickable trigger is `triggerType="text"`, so link affordance, keyboard focus and aria
  come from the component rather than a hand-rolled `<span>`.
- **No security surface**: UI-only, no IAM, no ARNs, no new dependency (collection-hooks
  was already a dependency), no `dangerouslySetInnerHTML`, no logging of anything.
- **No debug leftovers** — no `console.log`, `TODO`, `FIXME`, or commented-out code in the
  diff.
- **Docs + CHANGELOG** updated, including the corrected sorting claim.
- **Tests**: 414 UI tests pass, 20 of them new — the presentation contract (own column,
  label and score no longer sharing a cell, clickable only when something is behind the
  number, `—` for unscored, identical styling at high and low values) plus the sorting
  contract including `0.0` sorting as a score rather than as absent.

### Answers to the six questions

1. **Good PR?** Yes — single concern, targets `develop`, description explains the choices
   and now matches the code.
2. **Safe? Regressions?** The two real risks were the sorting defects above, now fixed and
   pinned. No backend, API, schema or config surface is touched.
3. **Good UX?** This is the point of the PR: the score is separated from the label,
   visibly clickable where there is something to click, sortable for the review workflow,
   and honest (`—`, no invented traffic light).
4. **Security?** No exposure — UI-only.
5. **Documented?** `docs/classification.md`, `CHANGELOG.md`, and docstrings carrying the
   *reasons* (including why no colour banding).
6. **Safe to merge?** Yes once CI is green. **Not visually confirmed in a browser** — the
   panels sit behind Cognito on a deployed stack, so presentation is pinned by tests, not a
   screenshot. Deploying to a stack afterwards is the check this review cannot make.

Issue #710, opened 2026-08-31
[Enhancement]: Advanced extraction sends the class schema three times per request (~72% redundant tokens)


rstrahan:
In Advanced (agentic) extraction the class schema is transmitted **three times** in every request:

| # | Where | Payslip class |
|---|---|---|
| 1 | prose `{ATTRIBUTE_NAMES_AND_DESCRIPTIONS}` substituted into the task prompt (`_format_schema_for_prompt`) | ~1,485 tok |
| 2 | `_build_system_prompt` appends `"Expected Schema:\n" + model_json_schema()` to the **system** prompt (`agentic_idp.py`, `complete_prompt = f"{final_prompt}\n\nExpected Schema:\n{schema_json}"`) | ~2,595 tok |
| 3 | the `toolSpec.inputSchema` Strands derives from `extraction: model_class` | ~1,612 tok |
| | **total ~5,692 tok where ~1,612 would suffice** | **~72% redundant** |

Copies 1 and 2 are the same information in two serializations (cleaned class schema vs. generated `model_json_schema()`). Measured on `config_library/unified/lending-package-sample` -> `Payslip` at ~4 chars/token.

(A fourth copy is stored in agent state as `extraction_schema_json`, but that is only sent if the schema-reminder tool is invoked, so it is not a per-request cost.)

**Two important caveats, so this isn't over-sold:**

1. **All three copies sit inside the prompt-cache prefix.** `{ATTRIBUTE_NAMES_AND_DESCRIPTIONS}` precedes `<<CACHEPOINT>>` in the default task prompt, and system + tools render earlier still. On a repeated-class workload these are cache reads at ~10% of input price, so the dollar impact is far smaller than the token count suggests.
2. **Where it does bite is context-window pressure** — ~4K redundant tokens per request consume budget that `context_buffer` / `shard_token_budget` actively manage, so it can push a document into sharding earlier than necessary.

**This is a quality risk, not a free win.** Restating a schema in prompt text often improves adherence — models tend to follow system/user instructions more reliably than a bare tool schema — so the redundancy may be load-bearing rather than accidental. Suggested approach: drop copy **2** first (largest, and most redundant with the toolSpec since both derive from `model_json_schema()`), keep copy 1, and A/B completeness and accuracy on the benchmark suite rather than reasoning about it.

rstrahan:
[...]

Issue #709, opened 2026-08-31
[Bug]: Four shipped config presets have property names Bedrock Converse rejects in a tool schema


rstrahan:
Bedrock Converse rejects `toolSpec.inputSchema` property keys that don't match `^[a-zA-Z0-9_.-]{1,64}$`:

```
ValidationException: tools.0.custom.input_schema.properties:
Property keys should match pattern '^[a-zA-Z0-9_.-]{1,64}$'
```

So **names containing spaces are rejected**, as are keys longer than 64 chars. Measured across all 94 class schemas in `config_library/unified/*` — 4 violate it, all with top-level offenders:

| Preset | Class | offending names | at top level |
|---|---|---|---|
| `bank-statement-sample` | `Bank Statement` | 6 | 3 |
| `lending-package-sample` | `Homeowners-Insurance-Application` | 36 | 13 |
| `lending-package-sample-govcloud` | `Homeowners-Insurance-Application` | 36 | 13 |
| `rvl-cdip-package-sample` | `questionnaire` | 6 | 6 |

e.g. `"Account Number"`, `"Statement Period"`, `"Policy Number"`, `"Purchase Date and Time"`.

**Harmless today** because nothing sends a toolSpec on the simple extraction path — which is precisely why it went unnoticed. It becomes a hard blocker for any work that puts the class schema on the wire as a tool schema.

Two traps worth recording:
1. The check is **top-level only** — a bad key nested inside an object property, inside `array.items`, or inside a referenced `$defs` is **accepted**. Verified by probe. So wrapping a class in a list makes a rejected schema silently start passing, with unsanitized names, relying on that shallow validation.
2. Relying on the shallow check is a forward-compatibility trap: if AWS ever makes it recursive, configs that worked start failing.

Any fix should sanitize **recursively** and map names back on the response, reusing the existing `X_AWS_IDP_ORIGINAL_NAME` constant and the pydantic generator's alias machinery (which already handles `"Date of Birth"` -> `Date_of_Birth`).

rstrahan:
**Fixed in `develop` — ships in v0.6.7.**

[...]

Issue #653, opened 2026-08-25
[Bug]: Classification sectionSplitting: llm_determined fails with Claude Sonnet 5 - outputs "start" for all pages


sindhu-achuthan:
### Describe the bug

When using Claude Sonnet 5 for document classification with `sectionSplitting: llm_determined`, the model incorrectly outputs `document_boundary: "start"` for every page of multi-page forms (e.g., Form-XXXX). This causes each page to be split into a separate section instead of being grouped together as a single document.

Claude Sonnet 4.5 correctly outputs `document_boundary: "continue"` for subsequent pages of the same form.

Expected behavior (verified with Claude Sonnet 4.5):

**From CloudWatch logs:**

[INFO] Parsed classification response as json: {
  'classification_reason': 'This document is clearly Form XXXX Page 1...',
  'class': 'Form-XXXX', 
  'document_boundary': 'start'
}
[INFO] Page 1 classified as Form-XXXX

[INFO] Parsed classification response as json: {
  'classification_reason': 'This document is clearly Page 2 of Form XXXX...Since this is page 2 of 2, it continues the Form XXXX document that began on page 1.',
  'class': 'Form-XXXX', 
  'document_boundary': 'continue'  ← CORRECT
}
[INFO] Page 2 classified as Form-XXXX
[INFO] Created 1 sections using LLM boundary detection (sectionSplitting=llm_determined)

Result: 1 section containing both pages ✅

Actual behavior (with Claude Sonnet 5):

Both pages return:

{
  "class": "Form-XXXX",
  "document_boundary": "start"  ← Page 2 should be "continue"
}

Result: 2 sections - each page incorrectly split into separate section ❌

Environment:

- Version: v0.5.3
- Region: GovCloud (us-gov-west-1)
- Broken model: `us-gov.anthropic.claude-sonnet-5`
- Working model: `us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0`

Why `sectionSplitting: disabled` is NOT a workaround:

Setting `sectionSplitting: disabled` groups ALL pages into a single section. This fails for document packages containing multiple forms (e.g., Form-XXX + Form-YYYY-A + Form-ZZZZ in one PDF). Each form must be extracted as a separate section with its own schema.

Workaround:

Use Claude Sonnet 4.5 for classification:

classification:
  model: us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0
[...]

PR #744, opened 2026-09-02
fix(forced-tool): make WS-05 actually work — schema metadata, wrapper-key data loss, and harness fixes

rstrahan:
Everything here was found by *running* the benchmark A/B that #741 made runnable. `force-on` failed live twice, in two different ways, and the investigation turned up three more problems.

## 1. Forced tool use sent schema metadata Bedrock rejects

Bedrock **meta-validates** `toolSpec.inputSchema`. An IDP class schema sets `$id` to the document-class *name*, so `Policy Application Form` produced `$id: "Policy Application Form"` — spaces, not an RFC 3986 URI-reference:

```
ValidationException: The json schema definition at
toolConfig.tools.0.toolSpec.inputSchema is invalid ...
$.$id: does not match the uri-reference pattern
```

Every section failed. `x-aws-idp-*` extensions were also going on the wire, including `x-aws-idp-examples` few-shot payloads — tokens spent on directives the model cannot act on.

Now stripped recursively (they constrain nothing, so removing them can't change what's asked). Keys inside `properties` are never filtered — a document may legitimately have a field called `id`. The client guard gained the matching check so another caller fails locally with an actionable message instead of a `ValidationException` from inside the retry ladder, after billing.

**Why #709's tests missed it:** they validated property *names*, and `$id` isn't one. No test asserted the wire schema's top-level keys.

## 2. Forced tool use then silently destroyed a 100-row list

Past the schema fix, Sonnet 5 answered the forced call with the whole extraction nested under an invented key:

```json
{"fields": {"Account_Number": "...", "Transactions": [ ...100 rows... ]}}
```

`fields` isn't a declared property, so off-schema-key handling dropped it and with it everything. The section stored `inference_result: {}`, reported `parsing_succeeded: true` and `honored: true`, raised no error, and **COMPLETED**.

| arm | n | failures | completeness recall |
|---|---|---|---|
| `force-off` | 6 | 0 | **1.0** every run |
| `force-on` | 6 | 0 | **0.167 mean** (0.0 three times on the 100-row doc) |

Cause is this module's own contract: the tool is named `emit_extracted_fields` and asked for "the fields". The description now requires top-level properties and names the wrappers not to use; `unwrap_tool_payload` recovers the payload anyway, because a prompt instruction is a request not a guarantee.

The unwrap is deliberately narrow — exactly one key, not a declared property, non-empty dict value, payload sharing a key with the declared properties — and compares against **both** authored and sanitized names, or a class whose every property was renamed would never match. A single-property class, and a single-key reply whose key *is* declared (a group field), are untouched. Both mirror cases are tested.

**Three explanations ruled out by measurement first**, recorded in the test docstring so nobody re-runs them:
- the sanitize/restore round trip — 100 rows survive it
- `$ref`/`$defs` in the tool schema — direct Converse calls returned 100 rows for the ref'd **and** dereferenced form, identical output tokens
- the live task/system prompt — 100 rows with and without it

> **This is not a property of tool use.** Advanced (agentic) extraction has always used tool-based structured output and measured completeness recall **1.0 across all 12 runs** of the #710 restatement A/B on the same stack.

The existing `min_population_ratio` check *did* fire (`populated only 0/4 schema fields (0%) ... possible silent extraction loss`) — as a warning, per the default `fail_action: warn`, left unchanged here.

## 3. `tests/pytest.ini` shadowed the package config — bare `pytest` ran live Bedrock

| | `addopts` | `agentic` marker |
|---|---|---|
| `lib/idp_common_pkg/pytest.ini` | `-m "not integration"` | registered |
| `lib/idp_common_pkg/tests/pytest.ini` | **none** | **none** |

pytest resolves config by walking up from the test path, so the `tests/` copy won for everything beneath it. `-m "not integration"` never applied, so a bare `pytest tests/...` collected and ran the two live-Bedrock tests. CI escaped only because `run_all_tests.py` passes `-m` explicitly. Removed, with guards: nothing may shadow it again, and the *running session* must actually exclude `integration`.

## 4. The pre-#653 control prompt was inside the release regression grid

My own bug from #741. `split-llm-legacyprompt` carries the **pre-fix** prompt and exists only as `boundaryctl`'s control, but I put it in `core_cells` so a suite could name it — silently adding it to `core`, `corefast`, `coresynth` and `full`. Controls now live in `control_cells:`; `corefast` drops 20 → 19 cells.

## 5. The harness was quadratic, and its integrity check false-positived

- **Launch gate:** `inflight` was never pruned, so every launch decision re-polled *every run launched so far* — one DynamoDB query each. A 12-run suite averaged 2.5 min/run; a 30-run suite degraded to 8. `corefast` (171 runs) would have issued ~14,000 polls and taken ~14 hours. Now ~1 launch/min, flat.
- **`@file:` axis values** (the frozen pre-#653 prompt) put file *contents* in the config while the index records the literal `@file:…`. `verify_config_axes` compared them verbatim and aborted `boundaryctl` on configs it had just generated correctly.
- Two new axes, `class_confidence` and `ocr_dpi`, filling documented gaps and used to eliminate both as confounds in the #653 factorial.

## Verification

- Full gate: **all 58 roots**, 4614 passed, 13 skipped
- 6 of 22 forced-tool service tests fail without the wrapper fix; the `$id` fix verified against the real failing class
- `make ruff-lint`, `make format` clean

`extraction.forced_tool` is off by default, so no shipped default changes — this makes the opt-in work at all.


rstrahan:
Re-measured on real labeled corpora. **The "buys nothing" verdict in `config-guidance.md` §7 does not hold** — it came from six synthetic runs.

322 paired documents via Test Studio, stack at the v0.6.7 tag, arms verified to differ in exactly one *stored* config key:

| | `ocr-benchmark` (9 classes, Sonnet 4.6) | `realkie-fcc-verified` (Sonnet 5) |
|---|---|---|
| paired | 293 launched / **282 scored** | **40/40** |
| accuracy Δ | −0.0003 (t=−0.34), sign p=0.42 | +0.0039 (t=+0.27), p=0.17 |
| cost/doc | $0.0286 → **$0.0279 (−2.3%, t=−4.06)** | +$0.007 (ns) |
| honored | **282/282 sections** | **108/108** |
| new failures | **0** (failure set byte-identical across arms) | 0 |

**Forcing is accuracy-neutral and cost-neutral-to-slightly-cheaper.** No measured reason to avoid it; I am *not* proposing a default change, because the win is small and corpus-specific.

⚠️ **The cost win is not caching.** Decomposed: **87% is fewer output tokens** (−34/doc — a tool call is terser than prose JSON with preamble and fences); the entire input+cache shift nets 13%, because +1,221 cache reads at 0.33/MTok nearly cancel −113 uncached input at 3.30/MTok.

What forcing *did* do, confirmed as a prediction registered before the run: it pushes short-prefix classes over Claude's **1,024-token minimum cacheable prefix**. `GLOSSARY` and `SHIFT_SCHEDULE` returned `cacheRead=0` across all 41 of their documents with forcing off, and cached with it on (0 → 1,839 and 0 → 1,901 tok/doc). Extraction read share 28.4% → 48.1%. **14% of that corpus was paying full price on its entire prompt prefix on every request, with nothing reporting it.** Mechanism, the measured boundary, and the +24.9% penalty a one-off document pays for a cache it never reuses: `docs/benchmarking/prompt-caching.md`.

PR #779. Separate bug found in the same run (not a forcing effect, identical in both arms): #778.

PR #751, opened 2026-09-03
fix(classification): boundary rules mis-split a repeated running header (#750)

### LINKED ISSUE #750: [Bug]: #653 boundary rules mis-split documents with repeated running headers — CI Step 8 (Nuveen) is 0/10
### What happened

CI Step 8 ("Agentic extraction with large tables", `samples/Nuveen.pdf`) fails with:

```
❌ FundInformation count mismatch: expected 532, got 193
❌ Step 8: Agentic extraction failed: Agentic extraction test failed: expected 532 fund items, got 193
```

**Extraction is not the problem.** The document reached `COMPLETED` in 173.9 s with
no failures and the correct document class. `193` is *exactly* the cumulative row
count of pages 1–6 of the 16-page fund table:

| pages | rows | cumulative |
|---|---|---|
| 1 | 24 | 24 |
| 2–5 | 34 each | 160 |
| 6 | 33 | **193** |
| 7–16 | 32–35 | 532 |

Page 7 was classified `document_boundary: "start"`, so the single document was
**split into multiple sections**, and the test reads only the first one:

```
find … -path '*/sections/*/result.json' -type f | head -1   →  sections/1/result.json
```

Each section's rows are complete; only the section count is wrong.

### Root cause

`e6cff6800` ("fix(config): sync every pinned classification prompt with the #653
boundary rules") replaced `scripts/sdlc/config/nuveen.yaml`'s pinned
`classification.task_prompt` with the #653 `<boundary-detection-rules>` block. Its
own commit message predicted this shape of failure: *"row completeness stays 100%
and the document still reaches COMPLETED. Only the section count changes."*

The Nuveen document defeats those rules:

1. Every continuation page reprints the title, logo **and the full column-header
   block**, and carries only a **bare page number** (`7`), never `Page 7 of 17`.
   Rules 1–2 (PAGINATION) therefore never fire.
2. Rule 3 (OPENING HEADER BLOCK ⇒ `start`) is evaluated **before** rule 4
   (CONTINUATION EVIDENCE ⇒ `continue`), and the repeated running header satisfies
   it.
3. The tie-breaker (*"prefer 'start' for a page that carries a complete opening
   header block"*) and the `CRITICAL - consecutive documents of the same type`
   clause both push the same way.
4. `contextPagesCount: 0` (the shipped default) means the classifier never sees the
   preceding page, so rule 4's *"a table or list continuing from a previous page"*
   has no evidence to fire on.

The classifier says so in its own reasoning (Nova 2 Lite, page 15):

> *"The page number '15' at the top right indicates this is part of a multi-page
> document, **but the presence of the full title** and structured data table
> **confirms this is the start of the document's content**."*

### Measured, offline, 10 reps per cell

Per-page `document_boundary` decisions replayed through `converse` with the real
prompt/model/inference settings (`us.amazon.nova-2-lite-v1:0`, `temperature 0`),
scored against known ground truth. `small_narrow` / `paginated_3pg` /
`twodocs_2x20` are the existing benchmark boundary fixtures
(`benchmarks/matrices/doc_matrix.yaml`).

| prompt / setting | Nuveen (want 1) | small_narrow (want 1) | paginated_3pg (want 1) | twodocs_2x20 (want 2) |
|---|---|---|---|---|
| pre-#653 prompt | 3/3 | 2/3 | 3/3 | 3/3 |
| **current (post-#653)** | **0/10** | 7/10 | 10/10 | 10/10 |
| + "running header is not an opening block" | — | 4/10 ↓ | 10/10 | 10/10 |
| + "a bare page number > 1 is decisive" | 3/3 | 2/10 ↓ | 10/10 | 10/10 |
| + both of the above | 10/10 | 0/10 ↓ | 10/10 | 10/10 |
| `contextPagesCount: 1`, prompt unchanged | 10/10 | 10/10 | 10/10 | **5/10 ↓** |

Two things follow:

* Step 8 is **0/10**, i.e. deterministically broken, not flaky — though *which*
  page flips is random (CI hit page 7; local reps hit 5, 6, 8, 12, 14, 16), so the
  observed row count varies run to run.
* **Every generic lever regresses one of the two directions #653 balances.**
  Prompt edits that fix the repeated-running-header shape cost the over-split
  case; `contextPagesCount: 1` fixes both over-split cases but halves accuracy on
  the over-merge case. This is not a wording bug with a free fix.

### Impact

* **CI:** Step 8 cannot pass. It has no diagnostic value in its current shape
  either — it reports a *classification* defect as an *extraction* defect, which
  is why the pipeline's generated RCA chased "verify agentic pagination over the
  full 532-item table" and concluded "root cause not captured".
* **Product:** any document whose continuation pages reprint a running header and
  paginate with a bare number — multi-page brokerage/fund tables, bank and
  transaction statements — is at risk of silent mis-splitting under the shipped
  default prompt. It is silent because completeness stays 100 % per section and
  the document still reaches `COMPLETED`.

### Expected behaviour

`samples/Nuveen.pdf` yields **one** section containing all **532**
`FundInformation` rows, and Step 8 distinguishes "extraction lost rows" from
"classification split the document".

### Notes

The other red lines in the failing pipeline (`Step 5`, `Step 14`, `exit code -9`)
are teardown artifacts: the buckets were deleted at 16:46:40 and the
still-running steps were then SIGKILLed. Only Step 8 failed for cause.

Refs #653, #737, commit `e6cff6800`.

rstrahan:
Fixes #750.

## What was wrong

CI Step 8 failed with `expected 532 fund items, got 193` on `samples/Nuveen.pdf`.
**Extraction lost nothing.** `193` is exactly the cumulative row count of pages
1–6 of the 16-page fund table — the classifier had split one document, and the
step read only the first section:

```
find … -path '*/sections/*/result.json' -type f | head -1
```

`e6cff6800` gave `scripts/sdlc/config/nuveen.yaml` the #653
`<boundary-detection-rules>`. This document defeats them:

- every page reprints the title, the logo and the full column-header block —
  which itself carries a date (`NAV as of 10/31/2024`), one of the page-1
  specifics rule 3 looks for;
- it paginates with a **bare** number (`7`), so rules 1–2 (PAGINATION) never
  fire;
- rule 3 (opening header block ⇒ `start`) is evaluated **before** rule 4
  (continuation evidence ⇒ `continue`);
- `contextPagesCount: 0` (the default) leaves rule 4 with no preceding page.

The classifier says so in its own reasoning:

> *"The page number '15' at the top right indicates this is part of a multi-page
> document, **but the presence of the full title** and structured data table
> **confirms this is the start** of the document's content."*

It is silent — each fragment stays 100 % complete and the document still reaches
`COMPLETED`. Only the section count changes, which is why the pipeline's
generated RCA spent its whole budget on "verify agentic pagination over the full
532-item table" and concluded "root cause not captured": the assertion had named
the wrong subsystem.

## The fix

The rules' own `PRECEDENCE` clause — a `BOUNDARY:` sentence in the class
description, which overrides the generic rules **for that document type only**.
No change to the shared prompt, no new config key, no code path.

## Why not fix the shared prompt

Because it was measured, and every generic lever buys this shape at the cost of a
direction #653 already balances. Per-page `document_boundary` decisions replayed
through `converse` with the real prompt / model / inference settings
(`us.amazon.nova-2-lite-v1:0`, `temperature: 0`), scored against exact ground
truth, **10 runs per cell**. `small_narrow` / `paginated_3pg` / `twodocs_2x20`
are the existing boundary fixtures in `benchmarks/matrices/doc_matrix.yaml`.

| prompt / setting | Nuveen (want 1) | `small_narrow` (want 1) | `paginated_3pg` (want 1) | `twodocs_2x20` (want 2) |
|---|---|---|---|---|
| pre-#653 prompt | 3/3 | 2/3 | 3/3 | 3/3 |
| **shipped rules (before this PR)** | **0/10** | 7/10 | 10/10 | 10/10 |
| + *"a running header is not an opening block"* | — | 4/10 ↓ | 10/10 | 10/10 |
| + *"a bare page number > 1 is decisive"* | 10/10 | 2/10 ↓ | 10/10 | 10/10 |
| + both | 10/10 | 0/10 ↓ | 10/10 | 10/10 |
| `contextPagesCount: 1`, prompt unchanged | 10/10 | 10/10 | 10/10 | **5/10 ↓** |
| **class-description `BOUNDARY:` (this PR)** | **10/10** | untouched | untouched | untouched |

Two things worth noting from that table:

- Step 8 was **0/10** — deterministically broken, not flaky. *Which* page flipped
  was random (CI hit page 7; local reps hit 5, 6, 8, 12, 14, 16), so the reported
  row count varied run to run, which is what made it look intermittent.
- `contextPagesCount: 1` is a legitimate lever for a corpus with no back-to-back
  duplicates. It is documented as such, with its cost, rather than made the
  default.

## Verification

**Live, end-to-end, on a deployed stack** with the committed config
(`us-west-2`, `samples/Nuveen.pdf`, `mode: advanced` agentic extraction):

```
document_class: Estimated2024AnnualTaxableDistributions
   section 1: pages 1-17, 532 rows
sections: 1
TOTAL FundInformation rows: 532
VERDICT: PASS          (573s, COMPLETED, no failures)
```

**Offline, 10 reps, committed config as-shipped:** 1 section / 532 rows in 10 of
10 runs.

**No regressions:**

- `scripts/tests` + `scripts/sdlc/tests` — **587 passed**. The 1 failure + 7
  errors in `test_headless_template_gate.py` are pre-existing (`idp_sdk` not
  installed locally) and identical before and after — verified by stashing.
- `lib/idp_common_pkg/tests/unit/classification` — 186 passed, 1 skipped.
- `benchmarks/tests` — 115 passed, 4 skipped.
- `ruff format --check` / `ruff check` clean on every touched file under the
  CI-pinned ruff `0.15.13`. (One `TID251` in `codebuild_deployment.py` is
  pre-existing and untouched by this PR — confirmed by stashing.)

## Also in here

**Step 8 can no longer mis-name its own failure.** It sums the list across
**every** section — numerically ordered, not `find` order, which was incidental —
and checks section count as a **separate** assertion:

```
    section 1: pages 1-6, 193 rows
    section 2: pages 7-16, 339 rows
❌ Classification over-split the document: expected 1 section, got 2
```

so a boundary defect reports as a boundary defect, and genuine extraction
truncation keeps its own `expected 532 fund items, got 400` message.

**Two guards, both verified to fail on a seeded regression** (not just to pass
today):

- `scripts/tests/test_nuveen_boundary_precedence.py` — the `BOUNDARY:` sentence
  is present and names `"start"`/`"continue"` explicitly, **and** the
  `PRECEDENCE` clause it depends on still exists. Nothing else checks a class
  *description*; the existing prompt-copy guard only covers
  `classification.task_prompt`. Removing `PRECEDENCE` would silently disable
  every class-level boundary instruction in the product.
- `scripts/sdlc/tests/test_step8_section_aggregation.py` — the aggregation
  (`193 + 339 == 532`), 0-based → 1-based page reporting, `null`/missing field
  tolerance, numeric section ordering (`1, 2, 10`), and that a mis-split and lost
  rows produce **different** error strings.

**Docs, both tiers:** `docs/classification.md` gains a *"Known failure mode:
repeating running headers"* section with the measured table and the copy-pasteable
class-description pattern; `lib/idp_common_pkg/idp_common/classification/README.md`
documents `PRECEDENCE` as the supported per-class override and where the working
example lives.

## Considered and rejected

A synthetic benchmark fixture for this shape (`running_header` on the
`bank_statement` generator: title + as-of date + bare page number reprinted per
page). It was built and measured, and it **did not reproduce** — 10/10 correct at
both 3 and 12 pages. Something more specific to the real document is required
(the 9-column landscape layout, the tall multi-line header block, the fund-family
groupings interleaved in the table). Shipping a fixture that passes would have
been a fake guard, so it was dropped; the regression guard is the real document
in CI Step 8, now asserting both dimensions.

## Not a bug

The other red lines in the failing pipeline (`Step 5`, `Step 14`, `exit code -9`)
are teardown artifacts — buckets were deleted at 16:46:40 and the still-running
steps were SIGKILLed. Only Step 8 failed for cause.


rstrahan:
## Live A/B on one stack: the control reproduces the CI number exactly

Same deployed stack, same code, same document — only the config differs.

**Control (`develop`'s `nuveen.yaml`, no `BOUNDARY:` sentence):**

```
   section 1: pages 1-6,   193 rows      <-- what `head -1` measured in CI
   section 2: pages 7-7,    34 rows
   section 3: pages 8-8,    35 rows
   section 4: pages 9-9,    34 rows
   section 5: pages 10-10,  34 rows
   section 6: pages 11-12,  68 rows
   section 7: pages 13-13,  34 rows
   section 8: pages 14-14,  34 rows
   section 9: pages 15-17,  66 rows
sections: 9   TOTAL rows: 532            (COMPLETED, 249s, no failures)
```

**This PR:**

```
   section 1: pages 1-17,  532 rows
sections: 1   TOTAL rows: 532            (COMPLETED, 573s, no failures)
```

Three things this pins down beyond the offline measurements:

1. **`193` is reproduced to the row, live.** Section 1 is pages 1–6 in both the
   failing CI run and this control — the same first split point, from an
   independent stack.
2. **Extraction was never at fault.** The control totals **532 across its 9
   sections**: every row was extracted, in every fragment. The old assertion
   reported that as `expected 532 fund items, got 193`.
3. **The fix is what changed it.** Same stack, same commit, same document; the
   only difference is the class description.

The control is also 3× cheaper in wall-clock (249s vs 573s) because nine small
sections extract in parallel — worth knowing, because "it got faster" is not a
signal that anything improved here.

Issue #746, opened 2026-09-03
[Bug]: `WorkflowErrorsAlarm` never fires — uses non-existent metric `ExecutionsFailedCount`


sirirako:
## Summary
`WorkflowErrorsAlarm` in `template.yaml` is configured against `AWS/States` metric
`ExecutionsFailedCount`, which does not exist. The correct Step Functions metric is
`ExecutionsFailed`. As a result the alarm can never receive datapoints and never
alarms on failed workflow executions — the primary "high error rate" signal for the
solution is silently non-functional.

Secondarily, neither `WorkflowErrorsAlarm` nor `SlowExecutionsAlarm` sets
`TreatMissingData`, so with the default (`missing`) they sit permanently in
`INSUFFICIENT_DATA` when there is no traffic, which masks the wrong-metric bug and
adds console noise.

## Impact
- **No alerting on workflow failures.** Operators relying on `WorkflowErrorsAlarm`
  (wired to the `Workflow Alerts` SNS topic) get no notification when Step Functions
  executions fail. This is a monitoring gap in production.
- The dashboard is unaffected — the "Workflow Executions Failed" widget already uses
  the correct `ExecutionsFailed` metric (`template.yaml` ~line 6867). Only the alarm
  is wrong, so the dashboard and the alarm disagree about the same signal.

## Root cause
`AWS/States` publishes `ExecutionsFailed` (count of failed executions per state
machine). There is no `ExecutionsFailedCount` metric, so CloudWatch has nothing to
evaluate.

Reference: [Step Functions CloudWatch metrics — Execution metrics](https://docs.aws.amazon.com/step-functions/latest/dg/procedure-cw-metrics.html)
(`ExecutionsStarted`, `ExecutionsSucceeded`, `ExecutionsFailed`, `ExecutionsAborted`,
`ExecutionsTimedOut`, `ExecutionTime`).

## Location
`template.yaml`, `WorkflowErrorsAlarm` (~line 6392) and `SlowExecutionsAlarm`
(~line 6409).

## Fix
Point the alarm at `ExecutionsFailed` and add `TreatMissingData: notBreaching` to
both alarms (no traffic = no failures = OK, not `INSUFFICIENT_DATA`).

```diff
   WorkflowErrorsAlarm:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmDescription: !Sub "Alert when workflow errors exceed ${ErrorThreshold} in 5
[...]

PR #767, opened 2026-09-04
feat(monitoring): alarms + dashboard for SQS queue backlog and DLQ depth (#761)

### LINKED ISSUE #761: [Feature]: alarms + dashboard coverage for SQS queue backlog and DLQ depth

# Feature request: alarms + dashboard coverage for SQS queue backlog and DLQ depth

## Problem
The solution's DLQs and queue backlog have **no monitoring**. Today `template.yaml`
ships only three CloudWatch alarms — `WorkflowErrorsAlarm`, `SlowExecutionsAlarm`,
`BedrockServiceOutageAlarm` — and the dashboard's only SQS widget plots
**throughput** (`NumberOfMessagesReceived` / `NumberOfMessagesDeleted` per minute on
`DocumentQueue`).

There is no alarm and no dashboard metric for:
- **DLQ depth** — `DocumentQueueDLQ` and `QueueSenderDLQ` are completely invisible.
  A message here is a document that exhausted retries (`maxReceiveCount: 1000` ≈ 8 hrs)
  and failed to process. Operators get no signal.
- **Queue not draining** — no view of `ApproximateAgeOfOldestMessage` /
  `ApproximateNumberOfMessagesVisible`, so a wedged consumer (QueueProcessor not
  pulling, concurrency counter stuck, poison message looping) goes unnoticed.

This is a production-monitoring gap: documents can silently fail (DLQ) or stop
flowing (stalled queue) with nothing firing.

## Why the existing alarms don't cover it
- `WorkflowErrorsAlarm` (`ExecutionsFailed`) only fires for executions that actually
  **start and fail**. A document stuck in the queue (never dequeued) or dead-lettered
  before an execution starts never produces an `ExecutionsFailed` datapoint.
- `SlowExecutionsAlarm` (`ExecutionTime`) only sees **completed** executions.
- So both are blind to the "stuck in / failed out of the queue" class of problem.

## Proposed change

Add CloudWatch alarms (wired to the existing alerts SNS topic) and dashboard widgets
for the SQS layer. SQS auto-emits these metrics in `AWS/SQS` — no custom metrics needed.

### 1. DLQ-not-empty alarm (high value, no tuning)
Alarm when any message lands in a DLQ — unambiguous, rare, actionable.

```yaml
DocumentQueueDLQAlarm:
  Type: AWS::CloudWatch::Alarm
  Properties:
    AlarmDescription: "Messages in the document DLQ (documents that failed after retries)"
    Namespace: AWS/SQS
    MetricName: ApproximateNumberOfMessagesVisible
    Dimensions:
      - Name: QueueName
        Value: !GetAtt DocumentQueueDLQ.QueueName
    Statistic: Maximum
    Period: 300
    EvaluationPeriods: 1
    Threshold: 1
    ComparisonOperator: GreaterThanOrEqualToThreshold
    TreatMissingData: notBreaching
    AlarmActions:
      - !Ref AlertsTopic
```
(Recommend the same for `QueueSenderDLQ` — documents that failed even to be enqueued.)

### 2. Queue-not-draining alarm (parameterized threshold)
Alarm on the **age of the oldest message**, not raw depth. Under the DynamoDB
concurrency-gating, a deep-but-moving queue is normal under load; a *growing oldest-
message age* is the real "not draining" signal.

```yaml
DocumentQueueBacklogAlarm:
  Type: AWS::CloudWatch::Alarm
  Properties:
    AlarmDescription: !Sub "Oldest queue message older than ${QueueBacklogAgeThresholdSeconds}s (queue not draining)"
    Namespace: AWS/SQS
    MetricName: ApproximateAgeOfOldestMessage
    Dimensions:
      - Name: QueueName
        Value: !GetAtt DocumentQueue.QueueName
    Statistic: Maximum
    Period: 300
    EvaluationPeriods: 2
    Threshold: !Ref QueueBacklogAgeThresholdSeconds
    ComparisonOperator: GreaterThanThreshold
    TreatMissingData: notBreaching
    AlarmActions:
      - !Ref AlertsTopic
```
with a new parameter (default sized above a typical burst-drain time):
```yaml
QueueBacklogAgeThresholdSeconds:
  Type: Number
  Default: 600   # 10 min
  Description: Alarm when the oldest DocumentQueue message exceeds this age (queue not draining).
```

### 3. Dashboard widgets
Add to the CloudWatch dashboard alongside the existing throughput widget:
- `ApproximateNumberOfMessagesVisible` for `DocumentQueue` (backlog depth), and
- `ApproximateNumberOfMessagesVisible` for `DocumentQueueDLQ` / `QueueSenderDLQ` (DLQ depth),
- optionally `ApproximateAgeOfOldestMessage` for `DocumentQueue`.

## Design note (why age, not count)
Because `QueueProcessor` gates on the DynamoDB concurrency counter, messages
intentionally wait in the queue while concurrency is saturated — so a raw
"`ApproximateNumberOfMessagesVisible > N`" alarm would be noisy under normal bursts.
`ApproximateAgeOfOldestMessage` distinguishes "deep but draining" (healthy) from
"stalled" (actionable). Keep the threshold configurable so deployments can tune it to
their observed drain times.

## Verification
1. Deploy; confirm the new alarms reach `OK` (with `notBreaching`) when idle.
2. Send a message to a DLQ (or force `maxReceiveCount` exhaustion) → `DocumentQueueDLQAlarm` → `ALARM` + SNS notification.
3. Pause the consumer / saturate concurrency so oldest-message age climbs past the threshold → `DocumentQueueBacklogAlarm` → `ALARM`.
4. Confirm normal bursts that drain within the threshold do **not** trip the backlog alarm.

_No response_

_No response_

### Additional context

_No response_

rstrahan:
Closes #761.

## The gap

`WorkflowErrorsAlarm` and `SlowExecutionsAlarm` both read the **state machine**, so neither can see a document that never got an execution: a dead-lettered document emits no `ExecutionsFailed`, and a document still waiting in the queue emits no `ExecutionTime`. Of the eleven DLQs in `template.yaml`, two were alarmed (`WorkflowTrackerDLQ`, `DataMartRollupDLQ`); `DocumentQueueDLQ` and `QueueSenderDLQ` were not, and nothing watched the queue for a stall. The only SQS dashboard widget plotted throughput.

## What this adds

| Alarm | Fires when | Notes |
|---|---|---|
| `DocumentQueueDLQAlarm` | First message in `DocumentQueueDLQ` | The document exhausted `maxReceiveCount` 1000 × 30 s visibility ≈ **8 hours** of retries |
| `QueueSenderDLQAlarm` | First message in `QueueSenderDLQ` | Worse: never enqueued, so **no tracking record** — absent from the Web UI rather than shown as failed |
| `DocumentQueueStalledAlarm` | Oldest message older than `QueueStalledAgeThresholdSeconds` **and** zero dequeues, for 30 min | Metric math; new parameter, default `1800` |

Both DLQ alarms also set `OKActions`. DLQ depth does not decay on its own, so the recovery notification is how you know someone actually drained the queue rather than muted the alarm — matching the existing `DataMartRollupDLQAlarm`.

Two dashboard widgets: **Document Queue Backlog** (visible, in-flight, and age-of-oldest on a right axis with a horizontal annotation at the threshold) and **Dead-Letter Queue Depth** (the three ingest-path DLQs on one graph).

## Two deliberate departures from the issue as filed

**1. Not a queue-depth alarm, and not an age-only alarm either.** The issue already rejected depth; age alone fails for the same reason, and more sharply than it first looks. `QueueProcessor` refuses a message by letting its 30 s visibility timeout lapse (`src/lambda/queue_processor/index.py`) — the message is never deleted and re-sent — so `ApproximateAgeOfOldestMessage` measures time since the **original send** and climbs monotonically for the whole saturation window. A healthy stack draining a bulk upload passes any fixed age threshold. The circuit breaker compounds it: on a Bedrock outage the processor pushes visibility out to `CircuitBreakerRecoveryTimeoutSeconds`, so age climbs while `BedrockServiceOutageAlarm` is already reporting the real cause. The alarm therefore requires age **and** zero `NumberOfMessagesDeleted`, in one metric-math expression (the shape `BedrockServiceOutageAlarm` already uses).

**2. A 30-minute window and a 1800 s default, not 10 minutes.** A message leaves `DocumentQueue` only when a workflow slot frees up, so a healthy stack saturated by long-running documents can legitimately dequeue nothing for as long as its slowest document takes. Six periods instead of two, because an alarm that cries wolf gets muted, which costs more than the detection latency. Thirty minutes of *zero* dequeues is where "slow" and "stuck" stop being distinguishable from outside — and deserves attention either way, since it is then either a wedge or a capacity shortfall. The alarm description and docs say so, and point at `MaxConcurrentWorkflows` as the other remedy.

## `FILL(m2, 0)` is load-bearing

SQS publishes `NumberOfMessagesDeleted` only when deletions occur, so the **totally stalled** case — the one this alarm exists for — produces no datapoint. Without `FILL` the expression evaluates to missing at exactly that moment and `TreatMissingData: notBreaching` reads the silence as healthy: an alarm guaranteed to stay `OK` through its own incident. That is [#746](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/746)'s failure mode reached by a different route, invisible in review (the expression looks right) and at deploy (CloudWatch accepts it), so `test_stalled_queue_alarm_fills_its_throughput_metric` pins it. Verified non-vacuous: removing the `FILL` fails the test.

## Verification

- **Metric-math expression validated against the live CloudWatch API** (read-only `get-metric-data`, not just cfn-lint): accepted, and with no data on either input it returns `0.0` for all 288 periods — so an idle stack reads `OK` rather than `INSUFFICIENT_DATA`, and `TreatMissingData` is belt-and-braces.
- Dashboard JSON parses after `!Sub` substitution; 16 widgets, both new ones at `y: 42`. The `DashboardMerger` groups widgets by view type and offsets each section, so there is no collision with the pattern dashboard.
- `make fastlint` clean. `cfn-lint template.yaml` produces no new findings (the two pre-existing `E3043` nested-parameter findings and two `W1028` are unchanged).
- `pytest lib/idp_sdk/tests/unit` → 407 passed, 35 skipped. `scripts/sdlc/tests` → 1 failure + 7 errors, all `ModuleNotFoundError: No module named 'idp_sdk'`, confirmed identical on `develop` (environment, not this change).

**Not deploy-tested.** The alarms' transitions (`ALARM` on a DLQ message, and the stall path) still want the live verification the issue describes.

## Out of scope

The other seven DLQs (`DiscoveryDLQ`, `JobTrackerDLQ`, `PostProcessingDecompressorDLQ`, and the four test-file-copy ones) remain unalarmed — same defect class, but feature-scoped rather than on the ingest path. Worth a follow-up issue.


rstrahan:
## Update — review findings addressed (commit 2df6f9a)

A clean-context review of this PR turned up a **blocking** defect that would have made this feature ship inert, plus three pieces of wrong operator-facing text I wrote. All verified against the live default stack before changing anything.

### 🔴 No alarm notification was ever delivered on a default deployment

`AlertsTopic` is encrypted with the stack CMK on **every** deployment, but the key-policy statement granting CloudWatch use of that key was wrapped in `!If [CircuitBreakerEnabledCondition, ...]` — and `CircuitBreakerEnabled` defaults to `"false"`. So every alarm action failed with *"CloudWatch Alarms does not have authorization to access the SNS topic encryption key"*, while the alarms transitioned to `ALARM` normally in the console and on the dashboard.

Confirmed on a live default stack (`CircuitBreakerEnabled=false`) — `describe-alarm-history --history-item-type Action` on `SlowExecutionsAlarm`:

```
2026-09-03T13:46:46  Failed to execute action arn:aws:sns:...:IDP1-AlertsTopic-...
2026-09-02T19:40:22  Failed to execute action arn:aws:sns:...:IDP1-AlertsTopic-...
2026-08-28T16:08:48  Failed to execute action arn:aws:sns:...:IDP1-AlertsTopic-...
```

This is #746's report-OK-by-silence shape one layer further out: the alarm resources are all perfectly well-formed, and the notification path is what's broken — which is why no existing gate caught it and why deploy testing was the only thing that would have. It affects the six pre-existing alarms too (`WorkflowErrorsAlarm`, `SlowExecutionsAlarm`, `ConcurrencyCounterDriftAlarm`, `WorkflowTrackerDLQAlarm`, `StaleOutputPurgeFailedAlarm`, `DataMartRollupDLQAlarm`), so **#761's stated goal was unreachable without fixing it** and it belongs in this PR rather than a follow-up. The grant is now unconditional, `test_cloudwatch_can_decrypt_the_alarm_topic_key_unconditionally` rejects any condition on it (mutation-verified: re-adding the `!If` fails it), and there's a ⚠️ CHANGELOG + docs note that upgraders should expect notifications to start arriving.

### Three corrections to text an operator reads mid-incident

1. **`QueueSenderDLQ` cannot be redriven.** It's a Lambda *async-invoke* DLQ, and `StartMessageMoveTask` supports only DLQs configured as the dead-letter queue of another SQS queue — the console offers no Redrive button. The description said redrive "works too". Removed; re-upload is the only recovery. (The equivalent claim on `DocumentQueueDLQAlarm` is correct — that one *is* a true SQS DLQ.)

2. **"No tracking record, absent from the UI" is wrong in the likelier case.** `create_document` runs *before* `send_message` (`src/lambda/queue_sender/index.py:215` vs `:231`), so a failed **enqueue** — the likelier failure in a queue named for the sender — leaves a row wedged at `QUEUED` in the Web UI. Only a failed `create_document` leaves no row. Sending an operator to look for an absent document that's actually sitting in the list is worse than saying nothing; both states are now named, likelier first.

3. **The circuit breaker is not designed out of the stall alarm.** I claimed the `AND` condition prevents a Bedrock-outage false positive. It doesn't: an `OPEN` processor calls `extend_visibility_for_outage` and returns failure *without deleting* (`src/lambda/queue_processor/index.py:724-731`), so age climbs **and** deletions stay at zero — the alarm fires alongside `BedrockServiceOutageAlarm` exactly as an age-only alarm would. Reframed as intended behaviour to correlate: suppressing it would mute the queue signal precisely when documents pile up. The `AND` still buys the deep-but-draining case, which was always the main point.

### Also

- `OKActions` on `WorkflowTrackerDLQAlarm` — my own argument for it applied verbatim, and without it two of the three DLQ lines on the new widget notify on recovery and one doesn't.
- Documented the **real ~1 hour detection latency** at the defaults (age threshold must be met *first*, then the 30-minute window runs on top), not 30 minutes.
- Documented the 6-hour metric-expiry caveat: in the detached-consumer-plus-no-new-uploads case, SQS stops publishing, `FILL(m1, 0)` yields 0 and the alarm returns to `OK` while still stuck. It fires first, so no notification is lost, but a later `OK` shouldn't be read as resolved.
- Hardened the stall-alarm guard: `FILL(m1)`, the `< 1` comparison direction, and the `QueueStalledAgeThresholdSeconds` interpolation are now all pinned — each breaks the alarm the same silent way.
- Drive-by: `docs/configuration.md` said `ExecutionTimeThresholdMs` defaults to `30000`; it's `300000`.

Left as-is deliberately: `DatapointsToAlarm: 6` alongside `EvaluationPeriods: 6` (redundant but documents intent), and the long widget titles (matches existing practice in this file).

**Re-validated:** `make fastlint` clean, `cfn-lint` no new findings vs `develop`, `pytest lib/idp_sdk/tests/unit` → 408 passed / 35 skipped, dashboard JSON still 16 widgets, all doc anchors resolve.

**Still not deploy-tested** — and given what the review found, the deploy verification from #761 (a message into each DLQ, one stall, and confirming SNS mail actually arrives now) is worth doing before merge.

Issue #753, opened 2026-09-03
[Bug]: A multi-record section that returns a single object is still undetected — the common half of #565 ask 3


rstrahan:
Split out of #715 (which now covers only the **Synthesize mode** schema transform) and the unfinished half of #565's accepted ask 3.

## What happens

When one section spans several records of the same class and the model returns a **single object** for the first record, nothing anywhere reports it. The section is `SUCCESS`, `ProcessingIssueCount: 0`, the document is `COMPLETED`, `instance_count` is `1`, and there is no warning. That is the original complaint in #565, and it is the *common* branch — not the rare one.

Two live runs against a deployed stack (Sonnet 5, `sectionSplitting: disabled` to force one section over a multi-record document, single-record class):

| Run | Result |
|---|---|
| 3-patient document | one object — Alice Anderson only; Brian Baker and Carol Chen silently lost, `instance_count=1`, no warning |
| 2-patient document | one object — Dana Doyle only; Evan Ellis silently lost |

So the model prefers a single object over an array. #687's array recovery is correct and worth having, but the branch that triggers it is the one where **nothing had been lost yet**.

## Why the shipped signal cannot catch this

- `extraction/service.py:2196-2201` — the `extraction_multi_instance_detected` warning is gated on `instance_count > 1`.
- `extraction/service.py:4638-4662` — `instance_count` is only ever set above 1 by (a) recovery of a top-level JSON array or (b) a class declaring `x-aws-idp-instance-array` (Designate mode).
- A single object for an unflagged class therefore lands on `instance_count = 1` by construction. The design plan's decision table states this outcome explicitly ("Unflagged class, single object → `1`, no extra signal"), so it is a known gap rather than an oversight — but it is the gap that matters most to the reporter.

## Why this gets worse, not better

`extraction.forced_tool.enabled` (`extraction/forced_tool.py`, `ForcedToolConfig` in `config/models.py:429`) puts the class schema on the wire as a forced Converse tool. It is **off by default today and gated on a measured win**, so nothing has regressed yet. But a forced tool whose input schema is a single-record object makes a top-level array *structurally impossible* — the model must pick one record. Turning it on converts the remaining loud case into a silent one and removes the only signal that exists. Detection should land before forced tool use is enabled by default, otherwise enabling it is a data-loss regression against #565.

## Scope — detection only

[...]

PR #774, opened 2026-09-04
feat(extraction): make the duplicated prose schema droppable (#710) — HOLD for 0.6.8, pending #775

rstrahan:
Finishes [#710](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/710), which was closed as completed after only the **first** of the two steps its own body proposes. (Reopened, with the status correction, [here](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/710#issuecomment-5545637545).)

## What was actually shipped before this

#710 proposes: "drop copy **2** first … keep copy 1, and A/B completeness and accuracy on the benchmark suite". Copy 2 shipped as `extraction.agentic.restate_schema_in_system_prompt` — **defaulting to `true`**, so a default deployment still sends all three copies and the issue's headline number (~5,692 tok on lending `Payslip` where ~1,612 would suffice) is essentially unchanged. This adds the second step: the prose copy substituted at `{ATTRIBUTE_NAMES_AND_DESCRIPTIONS}`.

## The knobs

```yaml
extraction:
  forced_tool:
    drop_prose_schema: false            # Simple. Default: unchanged behaviour.
  agentic:
    prose_schema: full                  # Advanced. full | minimal | names.
```

**Three states on the Advanced path, not two** — because the two paths are not equally safe to de-duplicate. Verified against `config_library/unified/lending-package-sample`:

| | Simple (`forced_tool`) | Advanced (agentic) |
|---|---|---|
| tool schema built from | the class schema, via `sanitize_tool_schema` | a **generated Pydantic model** |
| per-field descriptions | kept | kept |
| root class description | kept | **lost on all 6 classes** |
| nested-group descriptions | kept | **lost** |
| lending `Payslip` | **37/37** descriptions survive | **34/37** |

`sanitize_tool_schema` strips only non-constraining document metadata (`$id`/`$schema`/`$anchor`/`$comment`/`id`) and `x-aws-idp-*`, so on the Simple path the prose carries **nothing** the toolSpec does not and dropping it is a clean deletion. `model_json_schema()` has nowhere to put a description belonging to an *object*, and **four of the preset's six classes declare nothing but a root description** — `names` would leave those with no natural-language guidance at all. Hence `minimal`, which restates exactly what the round trip loses (class + group descriptions) plus the attribute names.

Per request on `Payslip` at ~4 chars/token: prose `full` **1,485** → `minimal` **372** → `names` **176**; toolSpec 1,173 (simple) / 1,612 (agentic).

## Exposure: default behaviour is unchanged, and that is asserted rather than assumed

- `prose_schema: full` renders `json.dumps(cleaned_schema, indent=2)` — byte-identical to the previous code path (`test_full_is_the_indented_json_schema`).
- The prompt is byte-identical with forcing on vs off (`test_forcing_on_alone_does_not_change_the_prompt`), so the WS-05 A/B still describes shipped behaviour.
- `drop_prose_schema` is **double-gated**: inert unless `forced_tool.enabled` is also true, which is itself off by default.
- A default deployment writes **no** `metadata.prose_schema` block (`test_the_default_records_nothing`).

Two cosmetic deltas everyone sees: `Config#default` gains the two keys at their current-behaviour values on upgrade (so it cuts a config revision), and `metadata.forced_tool` gains a `prose_schema` string — only where forcing is already on.

## The correctness-critical part: ordering

Forcing is skipped automatically for routes that cannot carry a `toolConfig` (custom Lambda hook, GPT-5.x Responses API). Dropping the prose on such a request would leave the model with **no schema at all** — a silent degradation, not a token saving.

So the `force` decision moves from `_invoke_extraction_model` (*after* the prompt is built) to `_initialize_extraction_context` (*before*), and both decisions read the same boolean. Model resolution is factored into `_resolve_extraction_model` so two call sites cannot diverge. A requested-but-not-applied drop records `metadata.prose_schema.kept_reason`, for the same reason `forced_tool.skip_reason` is recorded: an A/B in which "no effect" and "never applied" look identical is unreadable.

## Deliberate scope limits

- **Templates are not rewritten.** Only the substitution changes. `names` is not the empty string — the placeholder sits inside `<attributes>` in user-editable prompt text whose next sentence refers to "a list of attribute names", so an empty substitution would leave dangling prose and make that sentence false.
- **The assessment/confidence path is untouched.** It builds its own descriptions (`assessment/service.py:_format_property_descriptions`) and sends no tool, so dropping the schema there would remove the model's only description of the fields.
- **Escalation keeps the full prose.** A `field-subset` escalation reuses the original prompt with a *subset* data model, so the agent would otherwise see only field names; the subset schema is appended in full. It is the recovery path for an already-failed result — the wrong place to economize.
- **The #753 multi-instance probe** is in the wire schema and so in the prose today. A drop is safe only because it also rides the toolSpec, *including its question* — asserted, because a probe reaching neither copy would make the count structurally impossible while everything still reported success.

## Benchmark wiring

`prose_drop` / `prose_schema` axes; a `prose_schema_cells` group kept **out** of `core_cells` (five more cells would grow the release gate by half for a knob that ships off, and `prose-adv-names` is a control rather than a recommendation); `proseschema` and `prosecost` suites. `analyze.py` now rolls up `prose_schema_modes` from section metadata, the way `forced_tool_honored_rate` makes a forcing arm readable.

Neither axis is *swept*: `default_cell` holds `extraction_mode: simple` + `forcing: 'off'`, where `prose_drop` arms would behave identically and `prose_schema` would be ignored outright — both would confidently report "no effect" for a knob that never applied.

## The A/B — measured, gates pass

Ran on stack `IDPBench066` at this branch's commit, extraction model Sonnet 4.6.
`proseschema` suite: 5 cells × 3 docs (`valuenoise_100`, `longdesc_100`, `small_narrow`)
× 3 repeats = **45 runs, 0 failures**. All three Advanced arms hold
`restate_schema_in_system_prompt: off`, so they differ by ONE copy of the schema rather
than two.

| arm | n | failures | completeness recall | cell accuracy | cost | cost CV | rendering applied |
|---|---|---|---|---|---|---|---|
| Simple `prose-keep` (default) | 9 | 0 | **1.0** sd 0.0 | **1.0** sd 0.0 | $0.165 | 0.08 | forced-tool honored 1.0 |
| Simple `prose-drop` | 9 | 0 | **1.0** sd 0.0 | **1.0** sd 0.0 | $0.158 | 0.06 | 15 sections `names`, **0 kept** |
| Advanced `prose-adv-full` (default) | 9 | 0 | **1.0** sd 0.0 | **1.0** sd 0.0 | $0.591 | 0.51 | — |
| Advanced `prose-adv-minimal` | 9 | 0 | **1.0** sd 0.0 | **1.0** sd 0.0 | $0.595 | 0.63 | 13 sections `minimal` |
| Advanced `prose-adv-names` | 9 | 0 | **1.0** sd 0.0 | **1.0** sd 0.0 | $0.614 | 0.53 | 14 sections `names` |

Gate order per #710's own warning — **(1) completeness: no regression on any arm.
(2) accuracy: no regression on any arm. (3) cost: reported, not claimed.**

The *rendering applied* column is the same instrument as `forced_tool_honored_rate`:
`prose_schema_modes` counts sections that actually rendered each mode and
`prose_schema_kept` counts those where a requested drop was **not** applied. All arms
show 0 kept, so every arm genuinely ran — which matters, because `drop_prose_schema` is
deliberately inert with forcing off and would otherwise report a confident "no effect".

Two things reported rather than buried:

- **`scalar_accuracy` deviates — in the DEFAULT arm.** `prose-keep` scored 0.833 vs 1.0
  for all four other arms: 3 of its 9 runs missed one of **two** scalar fields, the known
  two-field-denominator artifact the `midetectlong` suite exists for. It points the way
  that would make dropping look *better*, and `cell_accuracy` (typed per-row truth) is
  1.0 sd 0.0 everywhere — so it is **not** evidence that dropping helps, and it is in the
  table rather than omitted.
- **The cost numbers do not support a claim.** The Simple pair moves −4.8% at CV
  0.06–0.08 — the tightest cost pair in §7 — but at n=9 that is still inside sampling
  error (t≈1.4). The Advanced arms sit at CV 0.51–0.63 and their *ordering is the
  opposite of the token ordering*, which is what non-deterministic agent turn count looks
  like. `config_matrix.yaml`'s own rule is `repeats >= 5` for a cost claim; this grid ran
  at 3 and the dedicated `prosecost` suite has **not** been run. The `kv_form` arm (a
  different class, run separately) was also not run. Both stated in §7.

**The defensible figure is the per-request static saving** — prose `full` 1,485 →
`minimal` 372 → `names` 176 tok on lending `Payslip`. All copies are prompt-cache reads
at ~10% of input price, so the payoff is context-window headroom (fewer shards), not the
bill. Written up in `docs/benchmarking/config-guidance.md` §7; scored data committed
under `benchmarks/results/v0.6.7/proseschema/`.

**Defaults stay as they are.** One synthetic 100-row corpus is not grounds for changing
what every deployment sends. A future default flip should be gated on "no
completeness/accuracy regression + a real static saving" — not on measured dollars.

## Merge of `develop`, and a semantic conflict it exposed

The CHANGELOG conflict is resolved by keeping develop's entry (#773 rewrote
`[Unreleased]` into short grouped entries) and folding this branch's content into
develop's existing #710 bullet, rather than carrying two.

The substantive part: develop's Prompt Preview work (`66480515f`) added a Tool Schema tab
and started counting the system-prompt restatement, "so the cost of turning this on is
visible before you do". It substitutes `{ATTRIBUTE_NAMES_AND_DESCRIPTIONS}` with the full
cleaned schema unconditionally — so with `prose_schema: minimal|names` it would render and
count a prompt no deployment sends, **overstating** it. That commit's own docstring says a
wrong bullet is worse than no bullet; the same holds for a wrong token total. So the
preview now honours both knobs (`proseSchemaModeFor` / `renderProseSchema`, mirroring the
backend including the `$defs` second pass), and unlike the Expected Schema block this one
is **exact** rather than an approximation — the backend renders it from the same cleaned
class schema with no generator in between. PromptPreview tests 37 → 53.

## Tests

69 new unit tests across `test_prose_schema.py` (renderers, plus the two claims above asserted as tests so a future change to `sanitize_tool_schema` or the Pydantic generator fails loudly) and `test_prose_schema_service.py` (the live section path: dropped-only-when-forced, in all three of its shapes; what reaches `bedrock.invoke_model`; probe survival; metadata; escalation). Full unit suite 5082 passed; `make ruff-lint` clean; `make typecheck` at the pre-existing baseline (2 errors / 46 warnings, unchanged).

## Docs

Both tiers, plus the measured-results page:
- `docs/extraction-and-confidence.md` — §Dropping the duplicated schema, §Forced tool use
- `lib/idp_common_pkg/idp_common/extraction/README.md` — the "three times" section, reframed as "three times **by default**" with the per-path loss table and the ordering constraint
- `docs/benchmarking/config-guidance.md` — axes table now; §7 write-up in the follow-up commit
- `CHANGELOG.md` — the previous `restate_schema_in_system_prompt` entry is merged into one #710 entry describing the final shipped shape (both are intra-cycle, so a user upgrading sees one feature)


rstrahan:
## Holding this for 0.6.8 — the benefit it claims does not exist in the code yet

Not a problem with this PR's implementation, and not an underpowered measurement. While
deciding whether more benchmark data would help, I checked the *mechanism* by which
schema de-duplication is supposed to pay, and it isn't there:

* `extraction/sharding.py` — `plan_shards(page_texts, token_budget=…)` budgets against
  **OCR page text only**. Schema/prompt bytes are not an input to shard planning at any
  point.
* `bedrock/sizing.py` — `compute_sizing_plan` derives that budget as
  `max_input × (1 - context_buffer)` minus an output reserve and an image reserve.
  **Prompt overhead is subtracted nowhere** — the prose schema, the system-prompt
  restatement, the toolSpec and few-shot examples are all absorbed by the blanket
  `context_buffer` (default 0.30).
* So reclaiming ~1,100 tokens leaves ~1,100 more tokens unused inside a reserve already
  ~60,000 tokens wide on a 200K-window model, and `max_pages_per_shard` (default 5) closes
  shards on page count regardless.

**There is therefore no shard-count, cost or latency mechanism behind either knob today.**
That is the explanation for the A/B result below — the measurement was fine, there was
nothing to find.

That claim ("the real benefit is context-window headroom, which is what pushes a document
into extra shards") originates in #710 itself and had already shipped into
`restate_schema_in_system_prompt`'s docs before this branch existed. It is corrected on
`develop` by **#776**, and this branch has merged that correction and applied it to its own
copies, so nothing here reintroduces it. The follow-up that would *create* the mechanism —
make the shard budget subtract measured prompt overhead — is **#775**.

## What the A/B found (unchanged, and still worth having)

45 runs on `IDPBench066`, Sonnet 4.6, 5 cells × 3 docs × 3 repeats, **0 failures**.
Completeness recall and cell accuracy both **1.0 sd 0.0 in all five arms** — so
de-duplicating the schema is *safe* on both paths, including the aggressive `names`
rendering. No arm showed a cost or latency benefit that survives its own variance.

Full numbers and the honest caveats are in the PR description and in
`docs/benchmarking/config-guidance.md` §7. The scored data is committed under
`benchmarks/results/v0.6.7/proseschema/`.

## Why hold rather than close

The code is sound, tested (69 new backend tests, PromptPreview 37 → 53), default-off with
"unchanged" *asserted* rather than argued, and it carries three things worth keeping:

1. `metadata.prose_schema` + the `prose_schema_modes` harness rollup — without these the
   question cannot be answered on a customer corpus at all.
2. The ordering fix: forcing and the prose decision now come from one `force` boolean
   computed before the prompt is built, so dropping the prose on a route that cannot carry
   a toolSpec (Lambda hook, GPT-5.x) is structurally impossible.
3. The Prompt Preview correction, which stops the preview overstating a shortened prompt.

What it lacks is a *reason for a user to turn it on*. That arrives with #775, not with more
runs of this grid — so this rides to 0.6.8 behind it rather than shipping two config knobs
whose documented payoff is currently unreachable.

## Open question for 0.6.8, not blocking

If it does ship, `extraction.agentic.prose_schema` is currently a visible dropdown for
**every** Advanced-mode user (its only gate is the parent's `extraction.mode == advanced`),
while `forced_tool.drop_prose_schema` is already double-gated behind the experimental
Schema Enforcement toggle. Given the null result, the `prose_schema` dropdown probably
wants the `integrated_confidence_strategy` treatment — kept in the config model and usable
via YAML/`config-upload` (which is how the benchmark harness drives it anyway), but not
surfaced in the config UI until there is something to recommend.

One more measurement is running for the record: the `prosecost` suite sequentially
(`--max-inflight 1`, repeats 5) to see whether the Simple path's 159 s → 109 s wall-clock
gap survives without the queueing confound. If it does, that is a user-facing reason to
ship that does not depend on #775; if it evaporates, the case rests entirely on #775.

rstrahan:
## Follow-up: the cost and latency gaps were queueing artifacts

I re-ran the Simple pair **sequentially** (`prosecost`, `--max-inflight 1`, `repeats: 5` —
which also satisfies `config_matrix.yaml`'s own `repeats >= 5` rule for a cost claim), to
test the one signal that had pointed anywhere. Both apparent gaps collapse:

| arm | n | fail | recall | cell acc | wall_s | wall CV | cost | cost CV |
|---|---|---|---|---|---|---|---|---|
| `prose-keep` | 5 | 0 | 1.0 | 1.0 | 137.0 | 0.12 | $0.1588 | 0.021 |
| `prose-drop` | 5 | 0 | 1.0 | 1.0 | 127.6 | 0.05 | $0.1576 | 0.011 |

| | interleaved grid (`--max-inflight 6`) | sequential (`--max-inflight 1`) |
|---|---|---|
| wall-clock | 159 s → 109 s (**−32%**) | 137.0 s → 127.6 s (**−6.9%**, t=1.17) |
| cost | **−4.8%** | **−0.8%** (t=0.72) |

Neither is resolvable at n=5+5. The first grid's numbers were **scheduling noise** — the
arms interleaved on a shared stack.

Worth stressing that this is a **strong** null, not an underpowered one: the sequential CVs
are 3–8× tighter than anything else in §7 (cost CV 0.011–0.021 against 0.25–0.63 elsewhere),
so an effect of the size the first grid suggested would have been unmissable. And it is the
*expected* result given the mechanism — `plan_shards` ignores prompt overhead entirely
(#775), so there was no path by which fewer prompt tokens could produce fewer dollars or
seconds.

The safety finding reproduces: recall and cell accuracy 1.0 in both arms, 0 failures,
forced tool honored 1.0, and `prose_schema_modes` shows 10 sections rendered `names` with
**0 kept** — the arm genuinely applied.

## Where that leaves this PR

**Every dimension is now measured, and the answer is "no effect of any kind":** no
completeness cost, no accuracy cost, no cost saving, no latency saving. The book is closed
on measuring the *current* design.

So the case for merging rests **entirely on #775**. Until the shard budget subtracts
measured prompt overhead there is no mechanism for either knob to pay, and shipping two
config knobs whose documented payoff is unreachable is worse than holding them. Recommend
this stays held until #775 lands, then re-measure on a document engineered to sit just over
a shard boundary — the test this feature has never been able to pass.

Committed: `docs/benchmarking/config-guidance.md` §7 now carries both the sequential table
and the "the first grid's gaps were queueing" correction, with the `--max-inflight 1`
invocation and the reason it matters in Appendix A. Scored data under
`benchmarks/results/v0.6.7/prosecost/`.

rstrahan:
Closing after assessment (2026-09-09). The PR's own A/B shows no measurable completeness, accuracy or cost effect (completeness and cell accuracy 1.0 on every arm; the cost pair is inside sampling error at n=9 and the dedicated cost suite was not run), and the one defensible gain — ~1,100–1,300 fewer prompt-prefix tokens per request — has no mechanism to act through while #775 is open (the shard planner absorbs prompt overhead in a blanket buffer) and could plausibly raise cost: #779 measured that Sonnet 4.6/5 skip caching below a 1,024-token prefix and two shipped `ocr-benchmark` classes already sit just under it, so removing prose from the prefix can turn cached reads into full-price input. Against that, it adds two configuration knobs (three states), a preview renderer that must track the backend, and a request-ordering change the PR itself calls correctness-critical, and it now conflicts with develop in ten files. If de-duplication is revisited it should be knob-free (drop the prose unconditionally on the Simple forced-tool path, where the toolSpec is shown to carry every description) and gated on #775 plus the cache minimum. The one finding worth keeping — the Advanced toolSpec drops the class root description on every class — is filed separately as https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/836.

Issue #778, opened 2026-09-04
[Bug]: page images over 3.75 MiB fail the whole document — Bedrock's 5 MiB image limit is enforced post-base64 and nothing checks it (3.8% of a shipped corpus)


rstrahan:
## Summary

Amazon Bedrock enforces its 5 MiB per-image Converse limit against the **base64-encoded**
payload, not the raw image bytes. The pipeline attaches page images with **no size check
of any kind**, and `extraction.image.target_width` / `target_height` default to `''`
(= "no resize"). So the effective budget for a stored page image is

```
5 * 1024 * 1024 / (4/3)  =  3,932,160 bytes  ==  3.75 MiB
```

not 5 MiB — and any page image above it fails the **whole document** with a hard
`ValidationException` at the extraction step. No downscale, no retry, no degradation.

## Measured, on a shipped seeded corpus at v0.6.7

`ocr-benchmark` (auto-seeded by `OcrBenchmarkDeployer`), 293 documents, stack `IDPBench`
at the v0.6.7 tag, `extraction.mode: simple`, Sonnet 4.6:

**11 of 293 documents (3.8%) failed**, all with the same error:

```
ValidationException: The model returned the following errors:
messages.0.content.2.image.source.bytes: image exceeds 5 MB maximum:
7167852 bytes > 5242880 bytes
```

Concentrated in one class — **8 of 31 `GLOSSARY` documents (26%)** — plus 2 `REAL_ESTATE`
and 1 `BANK_CHECK`.

### The base64 factor is exact

Comparing the stored page image (`.../pages/1/image.png`) to the byte count Bedrock
reported, across six failures:

| document | stored page image | bytes Bedrock counted | ratio |
|---|---:|---:|---:|
| `GLOSSARY_243.png` | 4,003,471 | 5,337,964 | **1.333** |
| `GLOSSARY_325.png` | 4,032,722 | 5,376,964 | **1.333** |
| `GLOSSARY_920.png` | 4,233,780 | 5,645,040 | **1.333** |
| `GLOSSARY_480.png` | 5,375,889 | 7,167,852 | **1.333** |
| `REAL_ESTATE_32.png` | 4,876,669 | 6,502,228 | **1.333** |
| `BANK_CHECK_17.png` | 4,904,953 | 6,539,940 | **1.333** |

Exactly 4/3 in every case. **This is why four of the failures have source images comfortably
[...]

PR #779, opened 2026-09-04
feat(benchmarks): measured prompt-cache behaviour, #744 re-measured, and a padding proposal

rstrahan:
Answers "what should we do with #744?" — and the answer changed, because the previous verdict came from six synthetic runs.

## Headline: the "buys nothing" verdict does not survive a real corpus

322 paired documents, two real labeled corpora, via Test Studio, on a stack at the **v0.6.7 tag**. Arms verified to differ in **exactly one stored config key** — decompressed from `Config#<profile>` in DynamoDB and diffed, not trusted from the upload.

| | `ocr-benchmark` (9 classes, Sonnet 4.6) | `realkie-fcc-verified` (1 class, Sonnet 5) |
|---|---|---|
| paired documents | 293 launched / **282 scored** | **40 / 40** |
| accuracy Δ | **−0.0003** (t=−0.34), sign **p=0.42** | **+0.0039** (t=+0.27), p=0.17 |
| cost/doc | $0.0286 → **$0.0279** = **−2.3%**, t=**−4.06** | +$0.007 (t=+0.78, ns) |
| forced tool honored | **282/282 sections** | **108/108 sections** |
| new failures | **0** (failure set byte-identical) | 0 |

**Recommendation: forcing is accuracy-neutral and cost-neutral-to-slightly-cheaper. There's no measured reason to avoid it — but I'm not proposing a default change,** because the win is small and corpus-specific. Enable it if you want the structural guarantee (malformed-JSON parse failure becomes impossible for declared fields).

## The cost win is NOT caching, and the write-up says so

| token class | Δ/doc | $/MTok | Δ$ | share of win |
|---|---:|---:|---:|---:|
| `outputTokens` | −34 | 16.50 | −0.000561 | **87%** |
| `inputTokens` | −113 | 3.30 | −0.000373 | 58% |
| `cacheReadInputTokens` | +1,221 | 0.33 | +0.000403 | −63% |
| `cacheWriteInputTokens` | −27 | 4.12 | −0.000111 | 17% |

**87% is fewer output tokens** (a tool call is terser than prose JSON with preamble and fences). The whole input+cache shift nets 13% — the +1,221 cache reads at 0.33/MTok nearly cancel the −113 uncached input at 3.30/MTok. Worth stating plainly: the cache mechanism below is the more interesting finding and it would have been easy to let it take credit for a saving it didn't produce.

## The cache dynamic — checked, not assumed

Several cost claims in this repo rest on one unmeasured sentence: the duplicated schema copies "sit inside the prompt-cache prefix, so they are cache reads at roughly a tenth of input price." Measured directly against `bedrock-runtime`, outside `idp_common`:

**1. Below the model's minimum, a cachePoint silently does nothing.** Sonnet 4.6 sweep: prefixes of 585 / 847 / 985 / 1,009 / **1,032** tokens all return `cacheWrite=0` **and** `cacheRead=0`; 1,059 writes. Boundary `(1,032, 1,059]`, consistent with the published 1,024 minimum. No error, no metric. The client inserts a cachePoint wherever `<<CACHEPOINT>>` appears with **no length check**.

**2. Two of nine classes in the shipped `ocr-benchmark` preset are under it** — `GLOSSARY` 949, `SHIFT_SCHEDULE` 1,000 — measured with each class's real prefix and the pipeline's own cachePoint placement.

**3. Confirmed live as a prospectively registered prediction**, on 293 documents:

| class | forcing off | forcing on |
|---|---:|---:|
| `GLOSSARY` (23 docs) | cacheRead **0 — never cached** | **1,839/doc — caching** |
| `SHIFT_SCHEDULE` (18 docs) | cacheRead **0 — never cached** | **1,901/doc — caching** |

Extraction read share **28.4% → 48.1%**. **14% of the corpus was paying full price on its entire prompt prefix, every request, with nothing reporting it.**

**4. A one-off document pays +24.9%** on the prefix for a cache it never reuses (write 1.25× vs read 0.1×), measured against a no-cachePoint counterfactual. Break-even is exactly the second request in the TTL.

**5. Nova is exempt.** Nova 2 Lite caches at ≤355 tokens and its write multiplier is 1.0×, so it has neither failure mode. Both mechanisms are **Claude-specific** — and `ocr-benchmark` as shipped pins Nova, so my arms pin Sonnet 4.6 (documented deviation; the cliff is a Claude question and Claude is what we recommend for extraction).

**So the "schema is cheap because it's cached" claim needs the class's prefix length and the workload's arrival rate attached, or it's unfalsifiable.** That has direct bearing on #710/#774: on a sub-minimum class a token saving is worth its **full** face value, ~10× what we've been saying.

## Two instrument bugs

- **`detection_ab_teststudio._tokens()`** sums every metering key matching `"token"`+`"input"`, folding `cacheRead` and `cacheWrite` into `inputTokens`. A toggle moving tokens *between* those classes reports ≈0 delta, hiding a 12.5× price difference. **Every prior real-corpus A/B in this repo used it** — that's why nobody noticed.
- **One was mine.** My sweep initially appeared to show Nova failing reads on 5 of 8 repeats. `filler()` builds by appending, so consecutive sweep points share cache state. Six *identical* calls: both models write once, then read reliably. Recorded in the doc so a re-run isn't misled.

## Separate bug found → #778

11 of 293 documents (3.8%) failed **identically in both arms**: `image exceeds 5 MB maximum`. Bedrock enforces that limit on the **base64** payload — ratio exactly **4/3** on all six failures checked — so the real budget is **3,932,160 raw bytes (3.75 MiB)**, and nothing in the pipeline checks it while `extraction.image.target_width/height` default to no resize. The threshold separates pass/fail across **all 293 documents with zero exceptions** (largest success 3,839,952; smallest failure 4,003,471). Not a forcing effect — filed as [#778](https://github.com/aws-solutions-library-samples/accelerated-intelligent-document-processing-on-aws/issues/778).

## Contents

- `benchmarks/harness/cache_threshold_probe.py` — sweep the boundary for any model
- `benchmarks/harness/cache_audit.py` — classify any existing run as *caching* / *write-only (paying 1.25× for nothing)* / *never cached*, per phase and per class
- `benchmarks/harness/real_corpus_ab.py` — paired real-corpus A/B with the four token classes kept apart and arbitrary section-metadata counters
- `docs/benchmarking/prompt-caching.md` (new, in the sidebar) + `config-guidance.md` §7 rewritten, superseded synthetic paragraph kept and labelled historical
- `benchmarks/results/v0.6.7/{prompt-cache,forcing-real-corpus}/`

Docs-and-harness only — **no runtime code changed.**


rstrahan:
## Added: the padding question, measured and answered

Asked whether it would be worth **padding prompts past Claude's minimum cacheable prefix**. Measured it end to end; the proposal is in `docs/benchmarking/prompt-cache-padding-proposal.md` and the actionable subset is filed as **#780**.

**Scope of the problem, 32 classes across 5 shipped presets** (real prefixes counted by Converse, not estimated):

| minimum | models | never cache |
|---:|---|---:|
| 512 | Opus 5, Fable 5 | **0 of 32** |
| 1,024 | Sonnet 5/4.6, Opus 4.8 | **8 of 32 (25%)** |
| 2,048 | Opus 4.7 | **31 of 32 (97%)** |
| 4,096 | Opus 4.6/4.5, **Haiku 4.5** | **32 of 32 (100%)** |

**On Haiku 4.5 every `<<CACHEPOINT>>` in the product is inert.**

**It works.** A 201-token block applied to only the two sub-minimum `ocr-benchmark` classes (surgically, via `x-aws-idp-extraction-task-prompt`; the other seven are an untouched control), 293 documents paired against the baseline:

| | treated (41 docs) | control (241 docs) |
|---|---|---|
| cacheRead Δ/doc | **+1,094 / +1,129** | noise |
| uncached input Δ/doc | **−943 / −994** | **exactly +0** |
| cost Δ/doc | **−$0.00274, t = −17.4** (−20% of extraction input) | −$0.00015, t = −1.35 |
| accuracy Δ | **−0.0081**, 95% CI **[−0.0240, +0.0078]** | +0.0001, t = +0.14 |

**And it should still not be automated**, for two reasons the arithmetic doesn't show:

1. **The sign flips on workload shape.** One document of a class per 5-min TTL makes padding **35% worse**; two makes it better. A config can't observe arrival rate — auto-padding would penalise the deployments least able to notice, to save a batch deployment ~$335/yr per class.
2. **The accuracy bound can't be tightened.** Bounding a 1pp regression needs n ≥ 105 treated documents; the corpus has 49.

**On the one document that moved** — worth reading as a method note. `GLOSSARY_185` went 1.0000 → 0.6667, `pageNumber` 3 → null, and that looked causal because the pad says *"where a field is absent return null"* and *"prefer the body over a header or footer"*. **It does not reproduce**: 3/3 calls return `3` in both arms at temperature 0 with the page image attached. My first repro returned `null` in *both* arms because it omitted `{DOCUMENT_IMAGE}` — the page number is read from the image, not the OCR text. I'd stated the causal story before running the check; the check overturned it. 36 of 41 documents scored identically.

**Recommended instead** (→ #780): warn at config-validate when a class is under the minimum; surface cache read/write per class (already metered, never shown); add `extraction.prompt_cache: auto|off` because the unconditional cachePoint costs a one-off document a measured **+24.9%** with no way to decline; document the non-monotonic table. Prefer model choice (Opus 5/Fable 5 at 512 caches all 32) or enriching thin classes over padding. Never pad at the 2,048/4,096 tiers where the pad is 74–246% of the prompt.

New tooling: `cache_prefix_survey.py`, `per_class_ab.py` (a treatment on 2 of 9 classes is diluted ~7× in a corpus mean).

Issue #726, opened 2026-09-02
[Bug]: sectionSplitting: llm_determined over-splits a single multi-page document into several sections (82% of runs)


rstrahan:
## Summary

Now that the #686 fix makes `sectionSplitting: llm_determined` perform real boundary detection for **single-class** configs, it **over-splits**: one 3-page bank statement is fragmented into 2–3 sections in **14 of 17 measured runs (82%)**. `disabled` produces 1 section in 5/5.

This is the mirror image of #565 (under-splitting), and it is silent — **row completeness is 100% in every run**, so nothing fails and no metric drops.

## Measurement

Live on IDP1 (`v0.6.7.dev3`, us-west-2), `bank-statement-sample` class (single-class), `small_narrow` = one 3-page statement, 100 transaction rows, exact ground truth.

| cell | runs | sections per run | rows recovered | recall |
|---|---|---|---|---|
| `split-disabled` | 5 | `1,1,1,1,1` | 100 ×5 | 1.0 ×5 |
| `split-llm` | 5 | `1,2,3,2,2` | 100 ×5 | 1.0 ×5 |
| `enforce-off` (also `llm_determined`) | 6 | `3,3,3,2,2,3` | 100 ×6 | 1.0 ×6 |
| `enforce-warn` (also `llm_determined`) | 6 | `3,2,1,1,2,3` | 100 ×6 | 1.0 ×6 |

Reproduce: `make_configs.py --suite splitcost --class bank_statement` then `run_matrix.py --stack <S> --suite splitcost`.

## Mechanism — confirmed directly, not inferred

The per-page `document_boundary` signal persisted by #694 shows the classifier calling *every* page the start of a new document:

```
3-section run : boundaries={1:'start', 2:'start',    3:'start'}    -> sections=3
1-section run : boundaries={1:'start', 2:'continue', 3:'continue'} -> sections=1
disabled      : boundaries={1:None,    2:None,       3:None}       -> sections=1
```

So this is not a splitting-logic bug — the logic faithfully honours the signal. It is **boundary-detection accuracy on a homogeneous multi-page document**: with a single-class config there is no type change to anchor on, and page 2 of a repeating transaction table looks exactly like the start of a new statement.

## Why it matters even though completeness is preserved

- Output shape changes: 3 `result.json` files, 3 evaluation sections, 3 reporting rows, 3 sections in the UI, for **one** statement.
[...]

Issue #686, opened 2026-08-28
[Bug]: A single-class config can never produce more than one section (sectionSplitting silently ignored)


rstrahan:
### What happens

When a configuration defines exactly **one** document class, `ClassificationService`
short-circuits the entire classification path and emits **one section containing every
page** of the document. The LLM is never invoked, no `document_boundary` is ever
produced, and the configured `sectionSplitting` strategy is never consulted — so even
`sectionSplitting: page` (one section per page) has no effect.

For any packet that concatenates multiple records of a single type — the common
"one record type per packet" configuration — this collapses the whole packet into one
section by construction. Extraction then sees a multi-record section and returns only
the first record (see the companion issue on multi-record sections), so every record
after the first is lost. This is a code-level guarantee, not a model-quality issue.

### Where

`lib/idp_common_pkg/idp_common/classification/service.py`

- `:134` — `self.has_single_class = len(self.document_types) == 1`
- `:2302` — page-level path (`classify_document`) early-returns a single all-pages section
- `:2945` — holistic path early-returns a single all-pages section

Both early-returns happen **before** `_apply_section_splitting_strategy()`, which is why
`sectionSplitting` is bypassed.

### Reproduction

1. Configure one class only (any schema) with
   `classificationMethod: multimodalPageLevelClassification` and
   `sectionSplitting: llm_determined` (or `page`).
2. Process a multi-page document containing several separate records of that class.
3. Observe one section spanning all pages.

Classification log line, and note the elapsed time — no Bedrock call is made:

```
Only one document class 'patient_service_record' is defined. Automatically classifying
all pages as this class without calling backend.
...
Time taken for classification: 0.05 seconds
```

Verified on v0.6.6 (4/4 runs, 3-page packet → 1 section). Adding a second class to the
same config restores correct splitting (3 sections, 5/5 runs).

### Expected

The single-class optimization should skip the *class decision* (which is genuinely
predetermined) without skipping *boundary detection*. Concretely, one of:

1. Run the LLM for boundary detection only when `sectionSplitting: llm_determined`,
   with the class fixed — the page's `class` is known, only `document_boundary` is
   needed; or
2. At minimum, honor `sectionSplitting: page` and `disabled` in the single-class path
   rather than hard-coding a single all-pages section; and
3. Warn when a single-class config is combined with `sectionSplitting: llm_determined`,
   since that combination cannot currently do what it says.

### Impact

[...]

Issue #677, opened 2026-08-28
[Bug]: --govcloud deploy rolls back — Function URL is removed but its LWA-dependent handler is not


karolytanczos-govcio:
### Summary

`GovCloudTemplateTransformer._remove_lambda_function_urls` correctly deletes
`ChatStreamProcessorUrl` because `AWS::Lambda::Url` does not exist in GovCloud, but leaves
`ChatStreamProcessorFunction` in place. That function layers in the AWS Lambda Web Adapter,
published only in the commercial partition, so it cannot be created in GovCloud and the
stack rolls back.

### Version

Present on `develop`, and in `v0.6.4` / `v0.6.5`. Neither
`template_transform.py` nor the GovCloud transform mentions the LWA layer or that function.

### The failure

```
Resource handler returned message: "User: arn:aws-us-gov:sts::<account-id>:assumed-role/...
is not authorized to perform: lambda:GetLayerVersion on resource:
arn:aws-us-gov:lambda:us-gov-west-1:753240598075:layer:LambdaAdapterLayerX86:25
because no resource-based policy allows the lambda:GetLayerVersion action
(Service: Lambda, Status Code: 403)" HandlerErrorCode: AccessDenied
```

`ROLLBACK_COMPLETE` after ~15 resources create, 30 cancelled.

### Root cause: accounts are partition-scoped, the ARN substitutes only the partition

`template.yaml` builds the layer ARN as:

```yaml
- !Sub arn:${AWS::Partition}:lambda:${AWS::Region}:753240598075:layer:LambdaAdapterLayerX86:25
```

`${AWS::Partition}` correctly becomes `aws-us-gov`, but `753240598075` is the LWA publisher
in the **commercial** partition and AWS account IDs do not exist across partitions. So the
ARN can never resolve — and the error reads as a missing *resource-based policy*, which
looks like a permissions problem rather than a partition mismatch. No identity policy on the
deploy role can fix it.

### The documented override appears to have no target

`template.yaml` documents a `LambdaWebAdapterLayerArn` parameter:

```
# Override this parameter for GovCloud (GovCloud publishes LWA under a different
# account; see the LWA releases) or to pin a specific version.
```

We could not find a GovCloud LWA publication. The LWA README documents only
`arn:aws:lambda:${AWS::Region}:753240598075:layer:LambdaAdapterLayerX86:<n>` (commercial
[...]

PR #781, opened 2026-09-05
docs(v0.6.7): settle the advanced-mode cost question, refresh the guidance paper

rstrahan:
Release-validation pass over the **published v0.6.7 template**, plus the documentation
and harness fixes the work needed. No product code changes — everything here is docs,
benchmark results, the benchmark harness, and SRT tooling.

## 1. Full stack-test set — 7/7 PASS

All against the published `idp-main_0.6.7.yaml`, us-west-2. Each self-deployed its own
throwaway stack and tore it down; all 12 stacks confirmed `DELETE_COMPLETE`.

| Test | Result |
|---|---|
| `stacktest-zap` (OWASP ZAP DAST) | **PASS** — 117 PASS · 1 WARN · 0 FAIL · 0 IGNORE; High=0 Medium=1 Low=0 Info=0 |
| `stacktest-waf` (WAF IP allow-list) | **PASS** — WebACL associated |
| `stacktest-hosting-global` (APIGateway GLOBAL) | **PASS** |
| `stacktest-hosting-private` (APIGateway PRIVATE, VPC) | **PASS** |
| `stacktest-jobsapi` (Jobs REST API, VPC) | **PASS** |
| `transform-deploy-test-headless` | **PASS** — 218 resources; UI/Cognito/CloudFront absent; processing core present; **sample document processed** (1/1, 173.9 s) |
| `transform-deploy-test-govcloud` | **PASS** — 304 resources; no `AWS::CloudFront::*` / `AWS::Lambda::Url`; LWA chat-stream family removed (#677); UI retained on API Gateway; **sample document processed** |

The VPC tests used the account's only VPC carrying an `execute-api` interface endpoint,
attached to exactly the two private subnets passed, with the endpoint SG as the Lambda SG
— the same wiring as the v0.6.6 set. Afterwards, 4 orphaned `available` ENIs left by the
`apigwpriv` teardown were deleted; the VPC now has 0.

> ⚠️ **The `govcloud` result is from a COMMERCIAL region.** It proves the CloudFront-free,
> API-Gateway-hosted transform deploys and processes documents. It does **not** prove
> GovCloud behaviour — partition-correct ARNs, GovCloud model availability and the BDA
> project rejection are all invisible outside `us-gov-*`, and two of the three defects
> this tier exists for would not have been caught by a commercial run. The Knowledge Base
> was disabled (the runner's default), so the KB path is **not** exercised.

## 2. The advanced-mode cost question is settled — and it is not the agentic path

The v0.6.7 `corefast` gate saw advanced cells +25/+40/+25% with each cell inside sampling
spread at n=9. A matched **same-stack in-place A/B** settles it: one stack deployed from
the published v0.6.6 template, measured, `update-stack`ed to the published v0.6.7
template, then the **same config files** replayed. Same pricing hash both sides; each arm
run with no other grid in flight.

| cell | v0.6.6 | v0.6.7 | delta |
|---|---|---|---|
| `core-tt-simple-sep` *(control)* | $0.5356 ± 0.0009 | $0.5326 ± 0.0083 | **−0.6%** |
| `core-tt-adv-sep` | $1.4005 ± 0.1104 | $1.7119 ± 0.1520 | **+22.2%** |
| `core-tt-adv-int` | $1.4883 ± 0.0838 | $1.8177 ± 0.1420 (n=3) | **+22.1%** |
| `core-tl-adv-sep` | $0.9902 ± 0.0749 | $1.2209 ± 0.0624 | **+23.3%** |
| `core-bda-adv-sep` | $1.3562 ± 0.0764 | $1.6957 ± 0.2314 | **+25.0%** |

Cost CV on this document is 0.056–0.079, not the 0.28–0.39 `corefast` saw (that spread
was averaging three very differently-sized documents). At n=5 that is ~3% SE per mean, so
+22% is roughly **seven standard errors**.

**Mechanism, from the raw metering.** All of it is Sonnet 4.6 *input* tokens in the
Extraction phase, with output flat and recall 1.000 on both sides:

| `core-tt-adv-sep`, Extraction phase | v0.6.6 | v0.6.7 |
|---|---|---|
| inputTokens | 96,423 | 152,098 **(+57.7%)** |
| outputTokens | 44,107 | 44,938 (+1.9%) |
| `Extraction/lambda/requests` | **1** | **4** |
| `sections` (truth: 1) | 1,1,1,1,1 | **4,4,2,4,4** |

`med_narrow` is **one** bank statement. v0.6.6 made **zero classification inference
calls** on it — #686/#705 skipped splitting entirely for a single-class config, so its
`sections=1` was a short-circuit, not a decision. v0.6.7 fixes that, and the now-active
classifier over-splits (#726). Each spurious section is a full agentic loop.

**Confirmed causally, within one release** (`advsplitcost`, same stack, `repeats: 5`):

| arm | sections | Extraction invocations | `core-tt-adv-sep` | `core-tt-simple-sep` |
|---|---|---|---|---|
| v0.6.6, `llm_determined` | 1,1,1,1,1 | 1 | $1.4005 | $0.5356 |
| v0.6.7, `llm_determined`, cls Nova 2 Lite *(default)* | 4,4,2,4,4 | 4 | $1.7119 (+22.2%) | $0.5326 (−0.6%) |
| v0.6.7, `sectionSplitting: disabled` | 1,1,1,1,1 | 1 | **$1.2969 (−7.4%)** | $0.5289 (−1.3%) |
| v0.6.7, `llm_determined`, **cls Sonnet 5** | 1,1,1,1,1 | 1 | $1.5521 (+10.8%) | $0.7764 **(+45.0%)** |

So: **nothing regressed in the agentic path** — with the section count corrected it is
7.4% *cheaper* than v0.6.6. The whole premium is the price of fixing #686 while #726 is
open. The over-split is **model-dependent** (Sonnet 5 classification: 1 correct section
5/5; Nova 2 Lite: 0/5), which also reconciles this with config-guidance §7, whose
`sections_correct` 1.00 was measured on Sonnet 5.

Neither mitigation is free: `sectionSplitting: disabled` is measured *wrong by
construction* on packets, and Sonnet 5 classification costs simple-mode users +45%
because classification is charged per page.

## 3. Security snapshot → `security/test-results/0.6.7/`

| Test | Gate | Detail |
|---|---|---|
| SRT (Bandit/Semgrep/Checkov/security-matrix) | **PASS** | 0 open, 0 reopened HIGH (44 suppressed, 60 resolved) |
| RBAC static | **PASS** | 0 fail, 2 known-gap warn (GAP-01, GAP-02) |
| RBAC dynamic (live, `IDPRel067`) | **PASS** | **596 checks, 0 hard fail**, 1 known-gap warn (stateless-JWT post-logout); TLS1.0/1.1 correctly refused, TLS1.2 accepted, port 80 not served |
| ZAP DAST (live) | **PASS** | High=0, Medium=1 (Cross-Domain Misconfiguration, the same known finding as 0.6.6) |

One note on the SRT expectation: LAMBDA-012 in `nested/bedrockkb/template.yaml` is
**suppressed** in the committed `scripts/srt/issues.json` baseline; the live scan did not
re-raise it, so it lands as `resolved` rather than `suppressed`. Either way the gate is
0 open/reopened HIGH.

## 4. Benchmarks: guidance paper refreshed (§2–§4 were v0.6.5 throughout)

- **§2** — new `coresynth` grid, 19 cells × 7 synthetic docs, **133 runs, 0 failures**.
  Every Textract and BDA cell is recall 1.000 *and* per-row cell accuracy 1.000.
- **§3** — new `scaling` grid. **The simple-mode completeness cliff has moved out**, from
  ~1,000 rows to above 1,600 (0.009 → **0.724** at 3,200 rows). Over-splitting appears to
  be why, which makes `sectionSplitting: disabled` a size-dependent trade rather than a
  free win — flagged as a mechanism, not a proven cause.
- **§4** — new `cost` grid at the shipped default model. Part of the famous agentic cost
  variance is **classification** non-determinism: the same document is classified 1 to 5
  ways across five identical runs, and the simple control's cost does not move at all.
- **§7** left as measured, with a note reconciling its `sectionSplitting` result against
  §2–§4 (it held classification at Sonnet 5).

Header, pricing hash, stack and release labels all updated.

## Two corpus / instrument defects found and fixed

- **`longdesc_100` was rendering unreadable amounts.** Long descriptions were plain
  strings in a reportlab table cell, which does not wrap, so they overprinted the Amount
  column. The OCR of that document contained **zero** amount-shaped lines, and one row
  read `"...recurring monthly charge00ference invoice 0"` — the amount `0.00` stamped over
  the word `reference`. Anything measured about that column was measuring the defect.
  Now wrapped; results for `desc_len: long` documents are **not comparable across the
  fix** (page count 3 → 4).
- **The metric that would have caught it was not being emitted.** `cell_accuracy` needs
  `rows_typed` in the truth file, which only the value-noise documents had — so a run
  returning all 100 rows with `Amount: null` scored recall 1.000 / accuracy 1.000. It is
  now emitted for every generated statement. Because scoring re-reads S3/DynamoDB, this
  re-scores already-completed runs at no cost.

## Harness / tooling fixes

- **`resolve_stack` paginates `ListTables`.** It took the unpaginated first 100 sorted
  names, so in this 220-table account `IDPRel067-TrackingTable-*` resolved to `None`
  while `IDPAB066-*` resolved fine. Failing closed was the lucky outcome — a decoy
  (`Bootstrap`/`Discovery`/`Chat…Table`) inside the first page while the real table is
  outside it would have scored a grid against the wrong table.
- **Scored rows carry `cost_by_key` and `tokens_by_phase`.** Doc-level token sums mix
  every model the pipeline called, so a +31% input-token move can be entirely Nova and
  worth a rounding error. These are what localized the cost rise.
- **Runs record their `--set` overrides** in `meta.overrides`; `RETENTION.md` gains the
  `<suite>__<override-slug>/` convention, because `(suite, class)` does not identify a
  measurement when one release runs `cost` twice at two models.
- **`srt-clean` prunes SRT scan dirs whose template no longer exists.** Stale dirs for
  deleted templates (`nested/alb-hosting`) and for build artifacts made every scan print
  *"this scan cannot prove the tree is clean"* — and the curated security snapshot
  published the warning. Now 0 incomplete scanners.

## Concerns worth reading, not buried

1. **A confidence-pass recovery ladder lost a document.** `simple` @800 rows extracted
   all 800 rows, then the Assessment step **timed out at 900 s and was retried five
   times** (6,205 s) and never converged; the document ends `ABORTED`. Nova Lite hits its
   10,000-token output cap on a 25-row confidence batch, the recovery halves the batch and
   retries, and each retry is a fresh ~77 s Bedrock call — paid every attempt. This is the
   only non-completing run in ~380 document runs, and it is a **confidence** failure, not
   an extraction failure.
2. **A second, pre-existing unbounded ladder.** `invoke_agent_with_retry` is
   `max_retries=50, max_delay=1800` inside a 900-second Lambda — 30 minutes of backoff in
   a 15-minute function. Byte-identical in v0.6.6 and v0.6.7, so **not** a regression, but
   it is why `core-tt-adv-int` shows 2/5 aborted above. Both ladders should be bounded by
   remaining Lambda execution time.
3. **`extraction.forced_tool.enabled: true` produces schema-invalid output on every
   section.** `valid_rate` 0.000, 19 errors in 19 sections, all the same cause: the nested
   object field comes back as a **JSON string** (`'{"Street_Number":"100",…}' is not of
   type 'object'`). `forced_tool.honored` is `true`, so the toolSpec path works and the
   serialization of nested groups is what is broken; coercion refuses it as a type-family
   mismatch. The earlier WS-05 study concluded "no measurable effect" because neither
   recall nor per-row cell accuracy covers a nested group. The flag is off by default and
   should stay off.
4. **Advanced mode fabricates where simple mode abstains.** Given the unreadable column
   above, simple extraction returned `null` for all 100 rows — which the new default
   validation reports loudly — while advanced returned a fabricated **`0.0`**, which is
   schema-valid and passes silently. A plausible zero in a financial field is the worse
   failure.
5. **Simple mode's validation failures are mostly over-splitting, not extraction.** Simple
   cells validate at 0.54–0.71 and advanced at 1.000 — but `split-disabled`, same schema
   and same simple extraction with one section per document, validates at **1.000 with 0
   errors**. The failures are `'Account Number' is a required property` on continuation
   sections that legitimately do not carry it. So #726 now manufactures ProcessingIssues
   as well as cost.

## Not done / limits

- `baseline.json` **not** promoted (unchanged from the v0.6.7 audit's reasoning).
- The `corefast` accuracy gate's `+0.167 scalar on four cells` remains **untested as to
  cause**; the DPI hypothesis is unchanged and the `ocr_dpi` axis was not run.
- `images/benchmark-scaling.png` still plots the **v0.6.5** curve; §3's table supersedes
  it and says so. Figure not regenerated.
- `meta.overrides` is absent/`None` on every set committed here — those runmaps predate
  the field. The directory name and `rows[].resolved` carry the information.
- The `core-tt-adv-int` 0/5 → 2/5 failure-rate change is **not** attributed to the release
  (n=5, and the trigger is a Bedrock read-timeout, not a code path).

## Reproduction

Every command is in `docs/benchmarking/config-guidance.md` Appendix A and
`docs/benchmarking/releases/v0.6.7.md` § *Reproducing*. Measured spend for the paper:
**$234.99** over 172 document runs; the supporting A/B and mitigation runs add **$81.17**
over 70 runs.


rstrahan:
### Follow-up: both extraction findings are now root-caused and filed

Pushed `e17e3ab0b`, which replaces the symptom descriptions in the paper with mechanisms.

**#782 — the agentic path cannot express "I could not read this."**
A `required` attribute is rendered as a **non-nullable** Pydantic field, so `null`, `""` and
*omitted* are all rejected — and `invoke_agent_with_extraction_retry` then feeds the
validation error back with *"Please fix the extraction using the tools."* A fabricated value
is the only accepted answer. Deterministic, offline repro (no AWS): for a required `number`,
of `{null, "", omitted, 0.0}` **only `0.0` validates**. It bites `number`/`integer`/`boolean`
hardest because a plausible `0`/`false` is unverifiable, and it directly contradicts the
table-tool prompt's own instruction to *"set that single cell to null"*.

**#783 — forced tool use, and it is our bug, on Sonnet 5 only.**
`sanitize_tool_schema` renames *property* names for Bedrock's `^[a-zA-Z0-9_.-]{1,64}$` rule
but deliberately leaves **`$defs` definition names** alone, so a group named
`Account Holder Address` ships a `$ref` whose pointer contains spaces. Isolated to four
Converse calls:

| toolSpec shape | Sonnet 5 | Sonnet 4.6 |
|---|---|---|
| `$ref` → **spaced** `$defs` key *(what we ship)* | **`str`** | `dict` |
| `$ref` → unspaced `$defs` key | `dict` | `dict` |
| `$ref` + explicit `"type": "object"` | `dict` | `dict` |
| inlined object, no `$ref` | `dict` | `dict` |

So the fix is small and on our side: sanitize `$defs` names and rewrite the pointers. Sonnet
4.6 resolving it is why the earlier WS-05 A/B never saw this — and why neither
`completeness_recall` nor per-row `cell_accuracy` would have, since neither covers a nested
group.

Neither issue is a v0.6.7 regression, and #783's feature is off by default. I have **not**
attempted either fix in this PR — it stays docs + results + harness. Happy to take either.

PR #798, opened 2026-09-09
fix(extraction): coerce minItems in the completeness check

### LINKED ISSUE #797: Advanced extraction: completeness check raises TypeError when a class's minItems was saved through the Web UI
When a document class is authored in the Web UI with `minItems` set on an array field, running extraction in agentic mode crashes the completeness check:

```
TypeError: '>' not supported between instances of 'str' and 'int'
```

at `lib/idp_common_pkg/idp_common/extraction/service.py:1742` (`min_items > 0` in `_check_completeness_detailed`). The `completeness_check` section is lost from the extraction metadata entirely, so schema-constraint violations for that document go unreported.

**Why `minItems` arrives as a string.** `ConfigurationRecord._stringify_values` (`config/models.py:3707`) converts every numeric scalar to a string before writing to the Configuration table — its docstring says "Convert numbers to strings (avoids Decimal conversion issues)" — and nothing coerces schema properties back on read: `classes` is typed `List[Dict[str, Any]]` (`config/models.py:3031`), so validation does not reach inside. A class saved through the UI therefore round-trips `minItems: 100` into `minItems: "100"`.

**Scope.** Only the agentic path is affected: the check runs when `extraction_method == "agentic"` and `schema_analysis` or `ocr_analysis` is present (`service.py:5323`, `5381`). Two of the three `minItems` readers in the same file already coerce defensively and say so in their comments (`_analyze_schema_for_table_requirements` at 1443, `_build_extraction_issues` at 2670); this is the third reader missing the same guard.

Reproduced against `develop` @ `3654387a6`. Happy to submit the fix — it is a 7-line application of the same idiom the other two sites already use, plus two unit tests.

art-aws:
Fixes #797.

**What.** `_check_completeness_detailed` reads `minItems` straight into a numeric comparison. A schema whose `minItems` arrives as a string raises `TypeError: '>' not supported between instances of 'str' and 'int'` at the `min_items > 0` guard (service.py:1742), and would raise again at the percentage division below it. The section then vanishes from the extraction metadata.

This applies the same coercion idiom — with the same comment — that the two other `minItems` readers in this file already use (1443 and 2670): coerce with `int(... or 0)`, and treat a value that cannot be read as a number as absent.

**Why it is reachable through this repo's own config path.** `ConfigurationRecord._stringify_values` stringifies every numeric scalar on save; `classes` is `List[Dict[str, Any]]` so nothing coerces schema properties back on read; `_class_schema` is loaded from that table and handed to this method on the agentic branch (service.py:5382). Verified empirically: `{"minItems": 100}` round-trips to `"100"` through `_stringify_values` + `IDPConfig.model_validate`.

**Scope.** Agentic mode only, and only when `schema_analysis` or `ocr_analysis` is present (service.py:5323, 5381). Simple mode reads `minItems` through the already-guarded site.

**Tests.** Two unit tests appended to the existing `tests/unit/extraction/test_empty_list_recovery.py`:

- a stringified `minItems` (`"100"`) is still enforced — before the fix this raises the TypeError above;
- an uncoercible value (`"many"`) disables that one constraint instead of failing the whole check, matching the two already-guarded sites.

Both fail before the fix with the exact TypeError at service.py:1742 and pass with it (26/26 in the file). `ruff check`, `ruff format --check`, `make lint-cicd`, and the `typecheck-pr` script all pass locally on the changed files.


rstrahan:
Merged into `develop` as 248b44e8e — thank you, @art-aws. This is an exemplary bug report/fix pair: the issue traced the string all the way from `ConfigurationRecord._stringify_values` to the comparison at `service.py:1742`, and I was able to reproduce both halves independently (`_stringify_values` does return `'100'` for a nested `minItems`, and `_class_schema` reaches `_check_completeness_detailed` uncoerced via `_get_class_schema`). Both new tests fail on `develop` with the exact TypeError and pass with the fix.

One follow-up thought, not a criticism of this PR: the repo now has two *central* coercers for this same root cause (`_coerce_numeric_keyword` in `extraction/validation.py` and `_coerce_string_types` in `evaluation/stickler_backend/mapper.py`) plus three per-site guards in `extraction/service.py`. Coercing once inside `_get_class_schema` would cover every present and future reader. That's a bigger change than a bug fix should carry, so merging this as-is was right — worth a separate issue.

Note for future PRs: this repo's CI lives on an internal GitLab mirror, so GitHub PRs never show checks. I ran the affected suites, `ruff`, and `basedpyright` locally before merging, and the full pipeline is running on the mirror now.

PR #800, opened 2026-09-09
fix(ocr): declare xlrd so legacy .xls workbooks are actually read

### LINKED ISSUE #799: Legacy .xls workbooks convert to a single blank page — xlrd is never declared
`.xls` is advertised, accepted, and routed — but silently unreadable:

- advertised: CHANGELOG (0.3.6 section) — "Excel workbooks (.xlsx, .xls) with multi-sheet support"
- accepted: `src/ui/src/components/common/constants.ts:9` (`SUPPORTED_UPLOAD_EXTENSIONS` includes `.xls`)
- routed: `lib/idp_common_pkg/idp_common/ocr/service.py:2711-2712` (`ext in ["xlsx", "xls"]` handler)
- unreadable: the `[ocr]` extra declares `openpyxl` (the `.xlsx` engine) but not `xlrd`, the engine pandas requires for the BIFF format.

There is **no visible error**. `pd.read_excel` raises `Missing optional dependency 'xlrd'. Install xlrd >= 2.0.1 for xls Excel support`, `document_converter.py:219-221` swallows it and returns one empty page reading "Error reading Excel file". The document then completes classification and extraction with no content, so the failure surfaces downstream as inexplicably empty results.

Reproduced against `develop` @ `3654387a6` in a clean env installing exactly `idp_common[ocr]`: a two-sheet `.xls` converts to a single blank page; adding `xlrd==2.0.2` to the extra, both sheets' content comes back. Happy to submit the fix (dependency declaration + a `.xls` regression test — there are currently no Excel-converter tests).

art-aws:
Fixes #799.

**What.** Declare `xlrd==2.0.2` in the `[ocr]` extra (and mirror it in `[test]`, as `openpyxl` and `python-docx` already are), regenerate `uv.lock`, and add a regression test that reads a small synthetic two-sheet `.xls` through `convert_excel_to_pages`.

**Why a real-file test.** The converter swallows the ImportError (`document_converter.py:219-221`), so an import assertion cannot catch this class of bug — only converting an actual `.xls` exercises the missing engine. The fixture is synthetic (generated with `xlwt`: two sheets, six cells, ~5.5 KB).

**Verified red/green in isolated envs** (installing exactly `idp_common[ocr]`): without the declaration the test fails with the exact swallowed error above; with it, both sheets' content survives conversion.

**Note on the lock.** `uv lock` also picks up the package's own `0.6.7` version bump, which the committed lock predated (it still said `0.6.6`).


rstrahan:
Merged into `develop` as 7bf7a9aca (merge commit, because of a trivial CHANGELOG conflict with #798/#802) — thank you, @art-aws.

Your reasoning for a real-file test rather than an import assertion is exactly right, and it's why this bug survived: `document_converter.py:219-221` swallows the ImportError and returns a page of error text, so nothing short of converting an actual BIFF file exercises the engine. I verified red/green the same way you did (test fails in an env with `openpyxl` but no `xlrd`, passes once `xlrd==2.0.2` is importable), and checked the dependency itself: BSD-licensed, no OSV advisories, and 2.x is `.xls`-only so it cannot affect the `.xlsx` path. The `uv.lock` version correction from `0.6.6` to `0.6.7` matches `VERSION`, so that's a welcome fix to a stale lock.

Two things left over, both outside what you changed:

1. The `[all]` extra in `pyproject.toml` is a flat duplicate list rather than a reference to the other extras, and it still lacks `xlrd` — so `pip install -e "lib/idp_common_pkg[all]"`, which the README, docs and notebooks all describe as "everything", still gets the blank-page behaviour. No runtime impact (the OCR Lambda installs `[ocr]`), and no test enforces extras parity. One-line fix, happy to take it as a follow-up PR if you want it.
2. Adjacent instance of the same class of bug: `feature-platform/pii-anonymizer/hook/handler.py:242` routes `.xls` to `openpyxl.load_workbook`, which cannot read BIFF at all — `xlrd` wouldn't fix that one since it also writes a workbook back. Worth its own issue if you're interested.

PR #802, opened 2026-09-09
fix(pipeline): retry the Lambda code-artifact provisioning errors

### LINKED ISSUE #801: No retry policy covers the Lambda code-artifact provisioning errors — cold functions fail documents outright
`Invoke` answers `Lambda.CodeArtifactUserPendingException` (HTTP 409, runtime cause "Lambda is initializing your function. It will be ready to invoke shortly.") while a function's code artifact is being provisioned. No `Retry` entry in any of the five state machines names it, so a task whose function has gone cold fails outright on a condition that clears itself.

Observed on a stack that had been idle for weeks: consecutive documents each died at a different pipeline stage as one function after another woke up cold. Once the errors are retried, the condition cleared within two retries — between 40 and 106 seconds of wall time — in every observed execution.

The transient retry ladders are the wrong home for these errors: 13 blocks in the unified workflow retry transient faults for ~2.8 hours worst case, which is deliberate for a throttle but holds a concurrency slot pointlessly for a function that becomes `Active` in seconds (an over-long budget on the wrong error class once held documents ~5.2 hours — 210306f73). `Lambda.CodeArtifactUserFailedException` should not be retried at all: the API reference says it needs the function's code package updated.

Happy to submit the fix: a dedicated first-position `Retry` entry (5 s / 5 attempts / 2x ≈ 2.6 min worst case) on every Lambda task not already covered by a `States.TaskFailed` wildcard, plus a structural test across all five state machines. Reproduced against `develop` @ `3654387a6`.

art-aws:
Fixes #801.

**What.** `Invoke` answers `Lambda.CodeArtifactUserPendingException` (HTTP 409) while a function's code artifact is being provisioned. No Retry entry in any of the five state machines named it, so a task whose function had gone cold failed outright on a condition that clears itself — observed on a stack idle for weeks, where consecutive documents died at a different stage each time as the pipeline warmed one function at a time. Retried, the condition cleared within two retries (40–106 s) in every observed execution.

**Design.** The errors get their own first-position entry (5 s / 5 attempts / 2x ≈ 2.6 min worst case) rather than joining the transient list: 13 blocks in the unified workflow retry transients for ~2.8 h worst case — the right shape for a throttle, the wrong one for a function that becomes `Active` in seconds (an over-long budget on the wrong error class once held documents ~5.2 h, see 210306f73). This is the distinction `test_workflow_evaluation_resilience.py` already draws: deterministic conditions come off the generous ladder, genuinely transient faults keep it.

- 28 of 41 Lambda Task states get the dedicated entry.
- 13 already retry these errors through a `States.TaskFailed` wildcard and are deliberately unchanged — inserting ahead of a wildcard silently moves every error it used to catch onto a different budget, which would be a separate proposal.
- One of those 13 is new coverage: confbench's `MarkFailed` had no Retry at all, so a cold FailFunction meant a job failure was never recorded; it now carries the same transient entry as its sibling states.
- `CodeArtifactUserDeletedException` rides the same entry per the API reference; it has not been observed in the field. `CodeArtifactUserFailedException` is deliberately not retried (needs a code-package update).

**Guard test** (`scripts/tests/test_state_machine_provisioning_retry.py`, runs in CI via `make test-packages-cicd`): asserts over Lambda Task states across all five machines — including the two an `*.asl.json` glob misses (a `definition.json` and an inline YAML `Definition`) — so a task with no Retry fails; dedicated entries must list both errors and nothing else, must not sit behind a wildcard, and must span 120–600 s worst case computed with the real ASL defaults (1 s / 3 / 2.0, `MaxDelaySeconds` honored). Validated against six regression mutations, each turning the suite red.


rstrahan:
Merged into `develop` as b36c272e5 (merge commit, because of a trivial CHANGELOG conflict with #798/#800) — thank you, @art-aws.

I checked the load-bearing claims rather than taking them on trust, and every one I could verify was exact:

- `CodeArtifactUserPendingException` is documented as "wait for the function's `State` to become `Active` and try the request again", and the Step Functions best-practices page confirms Lambda errors surface as `Lambda.ErrorName` — so the error strings are right, and excluding `CodeArtifactUserFailedException` is right for the reason you give.
- 41 Lambda Task states, 28 dedicated entries, 13 wildcard-covered: exactly as claimed, by running your own parser over the five definitions.
- The dedicated entry's worst case is 155 s, inside the test's 120–600 s window; the 13 unified-workflow transient entries really do sum to 10,166 s (2.82 h) each.
- The five definitions are exhaustive: every `AWS::Serverless::StateMachine` in `template.yaml` and `patterns/unified/template.yaml` uses `DefinitionUri` pointing at one of the four files, plus confbench's inline one.
- Your guard test passes (20 cases), the whole `scripts/tests` directory passes (125), the existing `test_workflow_evaluation_resilience.py` and `test_retry_deadline_bounds.py` still pass, `cfn-lint` is clean on the confbench template, and `scripts/tests` is genuinely in the CI target so the guard will run.

Leaving the 13 wildcard-covered blocks alone, rather than inserting ahead of a `States.TaskFailed` retrier, was the right call — and the test module is a genuinely useful piece of documentation for *why* the budget shape matters.

One observation on the confbench `MarkFailed` hunk: the sibling wildcard entry it inherits (10 s / 3 / 2.0) sums to 70 s of waiting, which is shorter than the ~106 s you observed as the worst clearing time — so a very cold `FailFunction` could still fail to record the job failure, which is the case that hunk exists to cover. The `Catch: States.ALL → Failed` keeps the original failure visible either way, and consistency with its siblings is defensible, so I merged as-is. Worth revisiting if you see it in the field.

The one thing I could not verify is the field data itself — the idle-stack observation and the 40–106 s clearing window come from your runs, and the test's `budget >= 120` lower bound rests on them. No objection, just noting which part of the change is measured here versus measured there.

PR #803, opened 2026-09-09
fix(dynamodb): stop transact_write_items writing into the caller's payload

art-aws:
**What.** `transact_write_items` copies each transact item with a shallow `item.copy()`, so the nested `Put`/`Update`/`Delete` dicts stay shared with the caller — and the `TableName` this method injects lands in the caller's own dict. A payload the caller builds, logs, retries, or reuses against another client gains a key it never put there. Deep-copy instead.

**Why the tests also pin the retry path.** The nested `Item` values are protected today only by boto3's `copy_dynamodb_params` handler, which deep-copies parameters before the resource's wire-format serialization; switching this call to a low-level client would lose that protection and a retried payload would double-serialize (`{"M": {"S": {"S": "doc#1"}}}` where `"doc#1"` belonged). The second test locks the observable contract (a retried transaction stores the same item) regardless of which layer provides it.

**Typecheck note.** `client.py` predates the incremental typecheck gate; touching it surfaces 32 pre-existing `reportTypedDictNotRequiredAccess` findings on the `except ClientError` handlers (botocore types those keys as not-required; they are present when `ClientError` is raised). Silenced with a scoped file-level rule comment to keep this change about the payload sharing — happy to fix them properly instead if you prefer.

**No CHANGELOG entry** — the only in-repo caller passes a payload it then discards, so no shipped behaviour changes. Happy to add one if you'd rather.

### DIFF
--- a/lib/idp_common_pkg/idp_common/dynamodb/client.py
+++ b/lib/idp_common_pkg/idp_common/dynamodb/client.py
@@ -5,6 +5,12 @@
 DynamoDB client for direct table operations.
 """
 
+# botocore types ClientError.response with not-required keys, but every read of
+# Error/Code/Message in this file happens inside an `except ClientError` block,
+# where botocore populates them. File pre-dates the typecheck gate.
+# pyright: reportTypedDictNotRequiredAccess=false
+
+import copy
 import logging
 import os
 from typing import Any, Dict, List, Optional
@@ -231,10 +237,13 @@ class DynamoDBClient:
             DynamoDBError: If the DynamoDB operation fails
         """
         try:
-            # Convert table references to use the table name
+            # Convert table references to use the table name. Deep-copied:
+            # a shallow copy shares the nested Put/Update/Delete dicts with
+            # the caller, so injecting TableName below would write into the
+            # caller's own payload.
             processed_items = []
             for item in transact_items:
-                processed_item = item.copy()
+                processed_item = copy.deepcopy(item)
                 if "Put" in processed_item:
                     processed_item["Put"]["TableName"] = self.table_name
                 elif "Update" in processed_item:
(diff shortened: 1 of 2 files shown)


rstrahan:
Merged into `develop` as 1a3de366d — thank you, @art-aws.

The bug is real and your first test is a proper regression test: reverted to `develop` it fails with the table name sitting in the caller's own `Put` dict. I also confirmed the subtler point you raised — the second test passes on `develop` too, because boto3's `copy_dynamodb_params` handler (registered on the resource whose `meta.client` this method uses) deep-copies parameters before wire-format serialization. Documenting that as a contract-lock test rather than claiming it as a regression test is the honest framing, and the docstring explaining what would break if this ever moved to a low-level client is worth keeping.

Agreed on skipping the CHANGELOG: the only in-repo caller (`dynamodb/service.py:786`) discards the payload immediately, so nothing shipped changes.

One thing I merged as-is but would suggest removing in a follow-up: the file-level `# pyright: reportTypedDictNotRequiredAccess=false` is inert in this repo's environment. I ran `basedpyright` on `client.py` at both your head and the `develop` baseline and got **0 errors, 0 warnings in both** — the installed `botocore` (1.42.97) ships no `py.typed`, and no `botocore-stubs`/`boto3-stubs`/`types-boto*` is declared anywhere in the repo, so `ClientError.response` types as `Any` and there is no TypedDict to complain about. The 32 findings you saw must come from a stubs-equipped environment. So it silences nothing here while permanently disabling the rule for the whole file. Dropping it (or narrowing to per-line ignores on the `except ClientError` handlers) would be my preference.

Minor, while you're in that loop: the `if/elif` chain handles `Put`/`Update`/`Delete` but not `ConditionCheck`, which `TransactWriteItems` also accepts — such an item would go out with no `TableName`. Pre-existing and unused in-repo, so entirely optional.

PR #806, opened 2026-09-09
fix(workflow): retry transient Lambda failures only, under one stable error name (#787)

### LINKED ISSUE #787: ExtractionStep and AssessmentStep omit States.TaskFailed, so a function error is never retried (ShardExtractionStep has it)
## The asymmetry

In `patterns/unified/statemachine/workflow.asl.json`:

- **`ShardExtractionStep`** (line ~464) lists `States.TaskFailed` in its `Retry.ErrorEquals`, so any function error re-runs the task — and because `extract_one_shard` is idempotent skip-if-complete with `S3ShardPersistence`, only the incomplete shards are redone. Added with the SFN shard runtime in `707858bfb`, deliberately.
- **`ExtractionStep`** (line ~517, the *default* in-process runtime) and **`AssessmentStep`** (line ~588) do **not**. Their `ErrorEquals` covers `States.Timeout`, `Sandbox.Timedout`, `Lambda.*`, and the Bedrock modeled error names (`ThrottlingException`, `ServiceUnavailableException`, …), with no `States.TaskFailed`, no `States.ALL`, and no `Catch`.

## Why it matters

Step Functions matches the Lambda-reported `errorType`, i.e. the Python exception class name. The Bedrock *modeled* errors match because botocore names the dynamic class after the code. But the transient failures that actually reach these handlers as Python exceptions do **not**:

- `ReadTimeoutError`
- `ConnectTimeoutError`
- `EventLoopException` (Strands' wrapper)
- `ModelThrottledException` (Strands' throttle wrapper)

None of those appear in either list. So on the default extraction runtime, any of them fails the document outright with no retry, while the same condition on the shard runtime is retried 8 times with completed work preserved.

Perverse consequence: **being killed by the Lambda timeout is currently a better outcome than raising**, because a timeout surfaces as `Sandbox.Timedout`/`Lambda.Unknown` and *is* retried. Code that fails fast to save time is therefore penalised.

This came up in review of #786: an early-raise design for retry backoff had to be abandoned in favour of clamping the sleep, specifically because raising would have converted a retryable `Sandbox.Timedout` into an unrecoverable task failure on the default runtime. #786 works around it; this issue is the underlying gap.

## Options

1. Add `States.TaskFailed` to `ExtractionStep` and `AssessmentStep`, matching `ShardExtractionStep`. Simple and consistent. **Risk:** a genuinely deterministic failure (bad schema, unparseable document) would now be retried 8× at 2.5× backoff, multiplying cost and latency on documents that can never succeed. `ShardExtractionStep` accepts that trade because shard resume makes a retry cheap; the in-process runtime also wires `S3ShardPersistence` at the shipped default `max_concurrent_batches: 10`, so resume applies there too — but the blast radius is wider.
2. Add only the specific transient names (`ReadTimeoutError`, `ConnectTimeoutError`, `EventLoopException`, `ModelThrottledException`). Narrower, but a list of exception class names is brittle — it drifts as Strands and botocore change, and the drift is silent.
3. Add a `Catch` that routes a function error to a state which records the failure and decides, rather than blanket-retrying.

I'd lean to (1) for consistency plus an explicit note about the cost trade, but it changes failure behaviour for every document and deserves a deliberate decision rather than being folded into an unrelated fix.

## Verifying

```bash
grep -n "States.TaskFailed" patterns/unified/statemachine/workflow.asl.json   # one hit, in ShardExtractionStep
sed -n '517,545p' patterns/unified/statemachine/workflow.asl.json            # ExtractionStep
sed -n '588,620p' patterns/unified/statemachine/workflow.asl.json            # AssessmentStep
```

Related: `Runtime.ImportModuleError` is also absent from `AssessmentStep`, so a bad import in that function fails every document with no retry — worth considering in the same change.

rstrahan:
Closes #787.

## The gap

Step Functions retries a Lambda task by matching the reported `errorType` — the Python exception class name. `ExtractionStep` and `AssessmentStep` listed the Bedrock *modeled* codes (`ThrottlingException`, `ServiceUnavailableException`, …), which match because botocore names its dynamic classes after the code. The transient failures that reach a handler as plain Python exceptions did not match: botocore `ReadTimeoutError` / `ConnectTimeoutError`, a Strands `EventLoopException` wrapping one, a reset connection. Those failed the document outright on the default runtime with no retry, while the same condition that ran into the Lambda timeout *was* retried (`Sandbox.Timedout`) — so code that failed fast was penalised relative to code that hung. (#786 had to be designed around this.)

## The fix — transient only, never hard errors

The issue's option 1 (retry `States.TaskFailed`) would also retry a bad schema or an unparseable document eight times at 2.5× backoff for a document that can never succeed. The decision was to retry transient failures only.

- **`idp_common.utils.transient_errors`.** `is_transient_error(exc)` walks the `__cause__` / `__context__` chain (so a wrapper cannot hide a transient root) and classifies by botocore error code — the pipeline's own `DEFAULT_RETRYABLE_ERRORS` minus `ValidationException` and `ModelErrorException`, which are deterministic at the task level — by exception type (botocore timeouts and connection errors, `ConnectionResetError`, `TimeoutError`, `RemoteDisconnected`), by class name (`ModelThrottledException`, …) and by message markers. A bare `EventLoopException` is transient only if its root or message is. `TransientError(cause, where)` is the **one stable name** Step Functions matches; `raise_if_transient(e)` is the handler idiom.
- **Handlers.** The extraction handler, the shard-runtime handler (plan/shard/merge) and the assessment handler re-raise a transient cause as `TransientError`. Hard errors keep their own name; the assessment handler still marks those FAILED without retry, as before.
- **`workflow.asl.json`.** `ExtractionStep`, `AssessmentStep` and `ShardExtractionStep` retry `TransientError`. `ShardExtractionStep`'s blanket `States.TaskFailed` is **removed** — the same transient-only rule now applies everywhere; completed shards are still preserved by `S3ShardPersistence` for the retries that do happen.
- **Tests.** 31 unit tests on the classifier (throttling / availability / timeout codes incl. streaming casing, botocore and stdlib network types, a transient root under a wrapper, a deterministic root under a wrapper, `ValidationException` whose text mentions tokens, cycle termination, the re-raise keeps the cause) and a structural test pinning the three retry lists and that the three handlers raise the name.
- **Docs.** `docs/configuration.md`'s retry section showed a generic `States.TaskFailed` example; rewritten around the transient-only rule. CHANGELOG entry.

## Verification

Full unit suite **5,238 passed**; `patterns/unified/tests` 15 passed; ruff and type check clean on changed files. Not exercised live: a transient failure cannot be provoked on demand; the classifier is unit-tested and the ASL is structurally tested.

## Review follow-ups (second commit)

An adversarial review found one blocker and three should-fix items, all addressed. The shard-runtime handler re-raises `TransientError` for all three of its modes, but only `ShardExtractionStep` listed the name, so a modeled `ThrottlingException` in plan or merge mode would have gone from retried to not retried; `ExtractionPlanStep` and `ExtractionMergeStep` now list it and the structural test pins all five states. The classifier no longer follows `__context__`: `BedrockClient` retries by re-invoking from inside its `except`, so attempt 2's `ValidationException` carried attempt 1's throttle as context and was judged transient — reproduced against the real client. It now judges an exception by its own definite code first and follows only explicit `raise … from` links. The transient vocabulary gained the S3 spellings (`SlowDown`, `ServiceUnavailable`, `InternalError`), the stream error and the botocore/urllib3/stdlib connection types the removed blanket retry used to cover, while throttling *phrases* were dropped from the message markers because they appear inside coded `ClientError`s and in deterministic wrappers. The extraction and assessment handlers now wrap their whole body so the S3 and DynamoDB loads are classified too. Full unit suite: **5,253 passed**.


rstrahan:
Superseded by #820: identical commits under a new branch name because GitHub Actions stopped running on this branch after the first push (no check runs for any later head, including after close/reopen).

PR #809, opened 2026-09-09
fix(ocr): the [all] extra gets xlrd too, so "everything" reads .xls

art-aws:
Follow-up you offered to take in the #800 review (item 1): `[all]` is a flat duplicate list and still lacked `xlrd`, so the "everything" install kept the blank-page behaviour. One line + `uv.lock`.

Verified by installing exactly `idp_common[all]` in a clean env and converting the `.xls` fixture — both sheets come back where it previously produced the "Error reading Excel file" page.

No CHANGELOG entry: the Unreleased xlrd entry from #800 already announces `.xls` support; this completes it for the `[all]` install path before it ships.

### DIFF
--- a/lib/idp_common_pkg/pyproject.toml
+++ b/lib/idp_common_pkg/pyproject.toml
@@ -231,6 +231,7 @@ all = [
   "requests==2.33.0",
   "pyarrow==23.0.1",
   "openpyxl==3.1.5",
+  "xlrd==2.0.2",
   "python-docx==1.2.0",
   "strands-agents==1.14.0; python_version>='3.10'",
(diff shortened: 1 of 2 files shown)


rstrahan:
Merged as f7376f86d — thank you, @art-aws. I checked completeness rather than just the diff: parsing `pyproject.toml` and listing every extra that carries an Excel or Word engine gives exactly three (`ocr`, `test`, `all`), and all three now hold `openpyxl==3.1.5`, `python-docx==1.2.0` and `xlrd==2.0.2` — no fourth flat list left behind. Agreed on skipping the CHANGELOG: #800's Unreleased entry hasn't shipped, so this completes that entry rather than needing its own.

Issue #813, opened 2026-09-09
Batched assessment: row indexes in confidence-alert paths are slice-local, so alerts point at the wrong rows


art-aws:
In the batched assessment flow, per-row confidence alerts come out of the per-slice inference cores, and the row index in the alert path is local to the slice that produced it: with `batch_size` 25, global row 28's alert reads `transactions[3]`.

Verified empirically against `develop` (120 rows, batch 25, only global row 28 below threshold — the single alert reads `transactions[3]`; the merged assessment itself is correctly aligned, so only the alert surface is wrong). The retry path is worse than a constant offset: `_splice_missing_rows` re-assesses `[rows[i] for i in idx_chunk]` for arbitrary non-contiguous missing indexes, so its alerts come back numbered 0..N over an arbitrary index mapping.

Consequences: an alert can point a reviewer at a healthy row, and alerts for different global rows can collide onto the same path. (The dedupe in the fix PR deliberately skips indexed paths for exactly this reason.)

The correct fix appears to be latent in the codebase already: `enrich_assessment_with_thresholds` run over the MERGED assessment produces correctly-global alert paths — `extraction/service.py` calls it after the integrated missing-row retry and currently discards those alerts (`merged_assessment, _ = enrich_assessment_with_thresholds(...)`). Regenerating alerts from the merged assessment (or rebasing paths with each slice's chunk offset at merge time) would fix both the labels and the collisions. Happy to take a swing at it if useful.

art-aws:
Design study done, including the #831 interaction — posting before building so you can veto cheaply.

**Approach: regenerate the alert surface from the MERGED assessment, replacing per-slice accumulation.** After `assess_results_batched`'s merged assessment is final (post retry/splice), derive alerts via `enrich_assessment_with_thresholds` over it and return THOSE, instead of extending per-core emissions. Row paths become globally indexed by construction (`_enrich_leaf_container` enumerates the full merged list), which fixes both the mislabeling and the cross-slice collisions in one move — no offset arithmetic, and the retry path's arbitrary index mapping stops mattering because its alerts are no longer part of the surface.

[...]

PR #815, opened 2026-09-09
fix(assessment): one confidence alert per finding, not one per slice

### LINKED ISSUE #814: Batched assessment raises one confidence alert per slice, not per finding — large documents lose their tracking write
The batched assessment path raises one confidence alert per attribute below threshold **per slice**. Every slice is assessed with the same scalars/context (by design — the module docstring says so), so a non-list attribute is re-assessed once per slice and re-alerts once per slice. The merge already treats those repeats as redundant for the assessment itself (`"scalars": left["scalars"] or right["scalars"]` keeps the first), but the alert list is a plain concatenation (`"alerts": left["alerts"] + right["alerts"]`, same return dict in `_assess_slice_adaptive`; the `assess_results_batched` accumulators have the same shape).

**Why it matters:** the list is stored inline in the tracking item (`dynamodb/service.py`, `serialize_confidence_threshold_alerts`), and nothing bounds its size, so a large document can push the item past DynamoDB's 409,600-byte ceiling — the write fails and the run shows no result at all. Measured on a 512-row workbook: 2,603 alerts over 16 distinct group paths (~163 copies each), the dominant share of an item that breached the ceiling; two real documents were lost this way. The repo has hit this ceiling before — 263d03205 compressed the *config* table for the same reason — but the tracking-item path has no mitigation.

The existing test suite encodes the behaviour: `test_large_list_is_batched_and_fully_covered` asserts `len(result["alerts"]) == 5` under the comment "One alert per batch accumulated", with a fake that re-alerts the same `attribute_name` on all five calls — five copies of one finding.

Fix incoming (diagnosed, measured and fixed in production by Diego Galaviz; ported to develop): collapse repeats of the same non-indexed attribute path at the merge and at the section boundaries, keeping the lowest-confidence copy. Indexed paths are deliberately left untouched because their row indexes are slice-local and can collide across slices — that mislabeling is its own pre-existing issue (#813).

art-aws:
Fixes #814. Related: #813 (why indexed paths are exempt).

**What.** Batched assessment sends every slice the same scalars/context, so a non-list attribute is re-assessed once per slice — and re-alerts once per slice. The merge already discards the repeated *assessments* (`"scalars": left["scalars"] or right["scalars"]`) but concatenates the repeated *alerts* (`"alerts": left["alerts"] + right["alerts"]`). The list is stored inline in the tracking item, so on a large document the duplicates can push the item past DynamoDB's 409,600-byte ceiling: the write fails and the run shows no result at all. Measured: 2,603 alerts over 16 distinct group paths (~163 copies each), the dominant share of an item that breached the ceiling; two real documents lost.

**Fix.** `dedupe_alerts` collapses repeats of the same **non-indexed** attribute path, keeping the lowest-confidence copy (the alert asserts "scored below threshold"; the worst score is the strongest true form of that claim — and it makes the collapse order-independent where float jitter across slices would defeat a whole-record comparison). A missing confidence coerces to 1.0, so a scoreless copy never displaces a scored one. First-appearance order is preserved; unkeyable entries pass through; idempotent.

**Why indexed paths are exempt.** In the batched flow, row indexes in alert paths are local to the slice that produced them (the core enumerates the rows it was handed), so two slices can emit the same indexed path for *different* rows — collapsing them would delete a genuine finding. That slice-local labeling is itself a pre-existing defect, filed separately as #813; this PR neither depends on it nor worsens it.

**Where.** At the merge (where the diagnosis lives) and at the boundaries (where the failure lives): the `_assess_slice_adaptive` merge and both `assess_results_batched` returns in `batching.py`; the section boundary in `assessment/service.py` before the list leaves for the tracking item; and the integrated-path boundary in `extraction/service.py`, which accumulates alerts from the enrich pass, cross-shard merges and the missing-row retry — none of which pass through the batched merge.

**Tests.** New `test_alert_dedupe.py` (11 cases: collapse semantics, the measured corpus shape, indexed paths never collapsed even when byte-identical, lowest-wins order-independence, scoreless-copy handling, ordering, pass-throughs, idempotence). One existing assertion changed meaning rather than value: `test_large_list_is_batched_and_fully_covered` asserted `len(alerts) == 5` under "One alert per batch accumulated", with a fake that re-alerts the same `attribute_name` on all five calls — that expectation encoded the defect; it now asserts 1 and says why. Red/green verified against `develop` (the flipped assertion fails there; 280/280 assessment tests pass with the fix). `ruff`, `typecheck-pr` and `make lint-cicd` pass locally.

**Credit.** Diagnosed, measured and fixed in production by **Diego Galaviz**; this ports his fix to `develop` with the narrower no-indexed-paths dedupe rule (his production corpus duplicated only group paths, so the narrowing does not change what his fix collapsed there).


rstrahan:
Merged as 3bfef08a6 — thank you, @art-aws, and please pass on the credit to Diego Galaviz.

The safety question here is whether "same non-indexed path" really means "same finding", because if it doesn't this silently deletes findings. I traced the producers rather than taking the argument on trust, and it holds: `enrich_assessment_with_thresholds` walks the assessment **dict** and every list row carries an `[i]` suffix, so a single pass cannot emit the same non-indexed path twice — duplicates can only come from repeated passes over the same scalars, which is exactly what you collapse. I checked the standalone path too, since your `assessment/service.py` boundary also covers it: `service.py:1193-1225` iterates the attribute dict and labels rows `attr[i]`, so the dedupe really is the no-op your comment claims. Coverage is complete — the merge at 1250 and both `assess_results_batched` returns; the remaining `"alerts":` at 1203 is the single-slice base case where duplication can't arise. HITL is unaffected because `processresults_function` gates on `len(...) > 0` and the collapse never empties a non-empty list.

The failure mechanism is structurally confirmed: `serialize_confidence_threshold_alerts` writes the list inline into `section_data`, and at ~110 bytes per serialized alert your 2,603 copies is ~285 KB — consistent with being the dominant share of a 409,600-byte breach. Tests: 280/280 assessment, 1298 across extraction+assessment, and I verified the flipped assertion is genuinely red on `develop` (1 failed / 10 passed with `batching.py` reverted). Flipping an existing assertion always deserves scrutiny; this one was asserting five copies of one finding, so the change is right and your comment explains why.

Three follow-ups, all filed rather than blocking:

1. **The list is still unbounded** (#821). This removes the duplicate-scalar component, but indexed per-row alerts are deliberately never deduped, so a wide table with many sub-threshold cells can still breach the ceiling by itself — 512 rows x 8 columns is ~4,096 alerts, ~450 KB. Worth saying the CHANGELOG entry promises slightly more than the fix guarantees.
2. **`assessment_alert_count`** at `extraction/service.py:5070` still measures the pre-dedupe list while the line above stores the deduped one, so the recorded count can say 2,603 for a section carrying 16 (#822).
3. **Docstring precondition** (#822). The key assumes "same non-indexed path implies same finding". True for every producer that reaches these boundaries today, but *not* for the BDA path: `bda_processresults_function/index.py:1036+` iterates pages and uses the raw KV key, so the same key on two pages legitimately yields two same-path alerts. It never routes through your three sites, so no impact today — but given your stated aim of guarding boundaries against future accumulation paths, one sentence would stop a future author routing BDA alerts through it.

#813 stays open, and I've left it to you since you offered to take it — say if you'd rather I did.

art-aws:
Thank you for tracing the producers independently — the per-alert weight arithmetic (~110 B × 2,603 ≈ 285 KB) is a better statement of the harm than the one I shipped, and you're right that the CHANGELOG entry promised slightly more than the fix guarantees; #831's bound closes that gap properly.

I'll take **#813** as offered. Sketch, so you can veto early: regenerate the alert surface from the MERGED assessment via `enrich_assessment_with_thresholds` (whose paths are globally indexed by construction) instead of rebasing slice-local paths hop by hop — the retry path's arbitrary index mapping makes offset-rebasing fragile, and the enrich pass already exists and is discarded today at the integrated boundary. I'll check the interaction with #831's bound before writing anything.

Credit passed on to Diego — the diagnosis, the measured corpus and the production fix are his.

PR #816, opened 2026-09-09
chore(dynamodb): drop the inert pyright suppression from client.py

art-aws:
The follow-up you suggested in the #803 review. Reproduced your observation both ways: with the project venv (botocore installed, no `py.typed`, no stubs declared) basedpyright reports **0 errors on this file with and without the comment** — and without a venv it falls back to its bundled typeshed stubs and reports the 32 findings that motivated the comment in the first place. That environment is not the one the gate runs in, so the comment only pre-disabled the rule for a future in which stubs are added deliberately. Removed.

No behavior change; `ruff` and the dynamodb unit tests pass.

### DIFF
--- a/lib/idp_common_pkg/idp_common/dynamodb/client.py
+++ b/lib/idp_common_pkg/idp_common/dynamodb/client.py
@@ -5,11 +5,6 @@
 DynamoDB client for direct table operations.
 """
 
-# botocore types ClientError.response with not-required keys, but every read of
-# Error/Code/Message in this file happens inside an `except ClientError` block,
-# where botocore populates them. File pre-dates the typecheck gate.
-# pyright: reportTypedDictNotRequiredAccess=false
-
 import copy
 import logging
 import os


rstrahan:
Merged as 7a455f664 — thank you, @art-aws.

Confirmed the state that matters: with the comment gone, `basedpyright` reports 0 errors on `client.py` in the project venv, which is the environment the gate uses (`pyrightconfig.json` pins `venvPath: "."` / `venv: ".venv"`). Tests (63) and `ruff` are clean.

One honest note: I could not reproduce the no-venv side of your explanation — my run without the venv also reported 0 errors, because pyright printed the "venv .venv subdirectory not found" warning and then resolved against the ambient interpreter, which has an equally unstubbed `botocore`. Your account is plausible and I can't contradict it; it doesn't change the decision either way, since the gate's own environment is 0 errors with and without the comment.

For the record, if `botocore-stubs`/`boto3-stubs` is ever added deliberately, those 32 `reportTypedDictNotRequiredAccess` findings will surface on this file. The right response then is to fix the accesses — the keys really are present inside `except ClientError` — rather than reinstating a file-wide disable.

art-aws:
For the record on the part you couldn't reproduce: the divergence is whether the ambient interpreter has botocore installed at all. Mine didn't — and with **no** botocore importable, basedpyright falls back to its *bundled typeshed stubs* (which include botocore's TypedDicts), producing the 32 findings. Yours has an unstubbed botocore ambient, which resolves to `Any` → 0 errors, same as the venv. Both observations are consistent; the gate's environment is what matters and we agree on its state. And agreed on the future response if stubs are ever added deliberately: fix the accesses, not the file.

PR #837, opened 2026-09-10
fix(assessment): regenerate the alert surface from the merged assessment so batched row indexes are global

### LINKED ISSUE #813: Batched assessment: row indexes in confidence-alert paths are slice-local, so alerts point at the wrong rows

art-aws:
Fixes #813.

Implements the approach posted to the issue on 2026-09-09 (no veto raised): **regenerate the alert surface from the merged assessment, replacing per-slice accumulation.**

## What changed

- `assess_results_batched` gains `default_confidence_threshold`; when it and `class_schema` are present (both real callers now pass them), the returned `alerts` are rebuilt by `enrich_assessment_with_thresholds` over the **final merged assessment** — at both returns (single-call fallback and batched tail). Its enumeration of the full merged list makes every row index global by construction, so `transactions[6]` from the second slice becomes `transactions[16]`, and the collision with the first slice's genuine `transactions[6]` (the case #815's dedupe deliberately left alone) disappears with it.
- The integrated missing-row retry returns the alerts its final re-enrichment builds (previously discarded at that exact line) and `_attach_explainability` **replaces** its surface with them instead of extending — the retry's slice-relative alerts are no longer part of the surface, so its arbitrary index mapping stops mattering.
- Since #831 declared the stored list a derived copy of `explainability_info`, this makes it exactly that: a pure projection, flowing into `set_section_alerts` unchanged. `dedupe_alerts` stays at the boundaries as the guard it already is.

## Parity points verified while building (from the design study)

1. **Threshold resolution is identical** — both paths use `resolve_array_item_thresholds` over the dereferenced subschema with a raw property-threshold read. Emission semantics differ at three edges (a non-numeric confidence does not alert where the per-core checker coerced to 0.0 and did; non-dict list items; traversal depth) — these are the regime the inline integrated path **already** lives under, since it derives its surface from this same helper. The commit message records them.
2. **Idempotence** — enrichment over an already-enriched assessment verified stable (identical output and alerts), so the standalone path's pre-enriched leaves are safe.
3. **Alert shape on rows** is per-cell (`transactions[16].date`), matching the standalone path's per-sub-field alerts; the new producer test documents this.
4. Without a schema/default, the per-core accumulation remains (there is nothing to resolve thresholds against) — no behavior change for callers that don't opt in.

## Tests

- Producer: one low-confidence row planted in the second batch must surface at its merged-list index (fails on the old surface with `transactions[6]`, as captured in the assertion message).
- Retry: the low row planted among the spliced subset must surface at its final-list index; **mutation-checked** — reverting the return to the accumulated alerts fails it.
- Full unit suite: **5,335 passed**. `ruff check` and `ruff format --check` clean.


rstrahan:
## Review

Thanks — the design is right and the batched-list half is correct and well tested: regenerating from the merged assessment makes row indexes global by construction, and I verified threshold resolution in `enrich_assessment_with_thresholds` matches `_check_confidence_alerts_with_field_thresholds` (same property-threshold lookup, same `resolve_array_item_thresholds`). All four parity points from #813 hold.

One blocking regression, which I have fixed on this branch (maintainer edits were enabled):

- **Integrated-mode alerts were wiped whenever nothing needed retrying.** `_attach_explainability` now *replaces* its alert surface with what `_retry_missing_integrated_rows` returns, but that helper's nothing-to-retry early return handed back `[]`. Every integrated section whose inline pass had scored all rows — the common case, and always for a scalar-only class — therefore stored no confidence alerts at all (`assessment_alert_count: 0`). Reproduced against this branch vs `develop` with the `test_integrated_alert_count.py` harness and integrated mode genuinely on. The early return now rebuilds the surface from the merged assessment exactly as the retry path does. Two tests added: a scalar class keeps its `InvoiceTotal` alert, and a fully-scored 30-row list yields `Transactions[28].Amount` (global) superseding a slice-local incoming `Transactions[3]`.

Also pushed: `ruff format` on `test_batched_assessment.py` (the CI lint gate would have failed), the stale "indexes are slice-local" rationale in `test_integrated_alert_count.py`, a CHANGELOG sentence noting the whole standalone (`separate`) surface is now derived by the enrichment too — with its two edge differences from the per-call emission (a non-numeric/missing confidence no longer alerts; nested groups walked to any depth) — and a merge of current `develop` (CHANGELOG conflict with #841 resolved, both entries kept).

Two things worth knowing, not blocking: alerts on rows now regenerate per **cell**, so the #831 per-section cap of 50 covers roughly 50/columns distinct rows instead of 50 rows (the true count is still stored beside the list); and the `new_alerts`/`merged_alerts` accumulators on the regeneration path are now written but never read.

Local test run (fork PRs get no CI here): `tests/unit/assessment` + `tests/unit/extraction` → 1,319 passed; the 2 failures are `test_image_downscale_metadata.py` cases that fail identically on `develop` when the assessment suite runs first in the same process (test-order pollution from #841's test, being fixed separately — not this PR). Merging.

PR #840, opened 2026-09-10
fix(cognito): parameter-gated mutable email so federated users can sign in more than once (#835)

### LINKED ISSUE #835: [Bug]: Federated users can sign in only once — second sign-in fails with `user.email: Attribute cannot be updated` (email schema is Mutable: false)
### Describe the bug

On a User Pool built from `template.yaml` with an external identity provider configured, **a federated user can sign in exactly once.** The second sign-in is rejected outright by Cognito with:

```
error_description=user.email: Attribute cannot be updated.
```

No token is issued and no session is created. It is not a group or token-content problem — the user cannot get back in at all. Because the app client's refresh token lasts 30 days, it presents as "I can't log back in" some time after first use rather than as an immediate failure.

**Cause.** Cognito re-applies the identity provider's `AttributeMapping` on *every* federated sign-in, rewriting each mapped attribute from the incoming assertion. `template.yaml` maps `email`, `given_name`, `family_name` and `custom:idp_groups`, and the pool schema declares `email` as `Mutable: false`. The first sign-in creates the user and writes `email` (allowed as the initial write); the second attempts to write it again, the immutable flag forbids it, and Cognito aborts the sign-in. It does not matter that the value is identical.

**Confirmed by controlled experiment**, not just reasoning: two throwaway pools identical except that one flag, the same user signed in twice through each via a real OIDC provider.

| `email` schema flag | 1st sign-in | 2nd sign-in |
|---|---|---|
| `Mutable: false` (shipped) | OK | **FAIL** — `user.email: Attribute cannot be updated` |
| `Mutable: true` | OK | OK |

So `Mutable: true` is a genuine fix — but see "Why it hasn't been fixed" below.

### To Reproduce

1. Deploy a stack with `ExternalIdPType=OIDC` (or SAML) and the other `ExternalIdP*` parameters set. `scripts/security/live_checks/oidc_provider/deploy.py up` stands up a throwaway provider and prints the parameters.
2. Sign in through the hosted UI as a federated user. Succeeds.
3. Sign out (or clear the session) and sign in again as the same user. Fails with the error above.

`make verify-idp-federation` (see `.claude/skills/live-auth-checks.md`) reproduces this as a recorded `NOTE` in section 3b.

### Expected behavior

A federated user can sign in repeatedly, and a changed group claim from the IdP is applied on the next sign-in.

### Why it hasn't been fixed already

This was fixed once. Commit `11fd05dbd` flipped `email` to `Mutable: true`. Commit `3bed47097` reverted it, because a Cognito schema attribute is fixed at pool creation: applying the change to an **existing** stack surfaced as a misleading `Required custom attributes are not supported` error followed by `UPDATE_ROLLBACK_FAILED` — a wedged stack, worse than the bug. The revert's message says it "re-introduces the federated sign-in lockout for new deployments" and calls for a follow-up that doesn't break existing-stack updates. Nothing was tracked, hence this issue.

### Impact / who is affected

- Only deployments with an external IdP configured. Native users write `email` once at account creation and are unaffected on every deployment.
- Per user record: each new federated user gets exactly one sign-in.
- IdP-driven role changes cannot reach an existing federated user, because the sign-in that would apply them fails. To change a federated user's role on an affected pool, change the Cognito group directly rather than the attribute.
- The documented escape hatch (`admin-delete-user`, `docs/external-idp.md`) buys one more sign-in but yields a new Cognito `sub`, orphaning that user's `UsersTable` preferences and `allowedConfigVersions`.

### Proposed fix

**Short term (new deployments):** gate the schema flag on a parameter, so new stacks get `email: Mutable: true` while existing pools keep `false` and their updates keep working. Small change; stops the bleeding for every future deployment without touching anyone's existing stack. This is the "parameter-gated schema flag" option the revert commit suggested.

**Long term (existing deployments):** a migration to a new pool with `Mutable: true` — users re-federate, and the affected DynamoDB rows are re-keyed from the old `sub` to the new one. Needs a product decision.

**Ruled out:** dropping `email` from the attribute mapping (it is `Required: true`, so a federated user record cannot be created without it), and making it non-required (the app keys users by email — the `UsersTable` `EmailIndex` lookup behind `allowedConfigVersions` depends on it).

### Related

- Verified while testing #793. The changes there do **not** cause or worsen this — the second sign-in was equally broken before — but they do remove the old workaround: the trigger no longer re-syncs groups on token refresh, so on an affected pool an admin should change the Cognito group directly.
- `docs/external-idp.md` documents the lockout and the one-shot bridge.

rstrahan:
## Problem

On a User Pool built from `template.yaml` with an external identity provider configured, a federated user can sign in exactly once. Cognito re-applies the IdP `AttributeMapping` on every federated sign-in, the pool's `email` attribute is `Mutable: false`, and an immutable attribute cannot be rewritten even with an identical value, so the second sign-in fails with `user.email: Attribute cannot be updated`. Confirmed by controlled experiment in #835.

`Mutable: true` was shipped once (11fd05dbd) and reverted (3bed47097) because a Cognito schema flag is fixed at pool creation: the change failed every existing stack's update and left stacks in `UPDATE_ROLLBACK_FAILED`.

## Change

- **New parameter `ExternalIdPEmailMutable`** in the *External Identity Provider (Federation)* parameter group, with the console label *"Federated sign-in: mutable email attribute (set 'true' for ALL new deployments using an external IdP; fixed at stack creation, do not change on an existing stack)"*. Drives `UserPool.Schema[email].Mutable` through one condition.
- **Default stays `"false"`** so every existing stack resolves to the value its pool was created with. CloudFormation cannot tell a create from an update, so the template cannot default it to `true`.
- **`idp-cli deploy` sets it to `true` automatically when creating a new stack with `ExternalIdPType` set** — never on an update, a headless deploy, or when the caller set it explicitly.
- **Compensating control:** the pool has no `UsernameAttributes` (no email uniqueness), `email` is in the app client's `WriteAttributes` because it is IdP-mapped, SRP tokens carry `aws.cognito.signin.user.admin`, and the configuration resolver keys `allowedConfigVersions` on the token's email claim. On a mutable pool a native user could adopt another user's address and inherit their config scope. Pools created with the flag therefore also set `UserAttributeUpdateSettings.AttributesRequireVerificationBeforeUpdate: [email]` (gated on the same condition, so an existing pool sees no property change).
- Headless transform removes the parameter and its condition.
- `verify_federated_signin.py` reads the pool's actual schema flag: on a mutable pool section 3b is a real PASS/FAIL (second sign-in succeeds, re-asserted claim applied); on an immutable pool it remains a NOTE.
- Docs: `docs/external-idp.md` parameter table, all six deploy examples, rewritten mutability section, troubleshooting row; skill + live-checks README; CHANGELOG.

## Verification

- `cfn-lint template.yaml`: only pre-existing findings (artifact-bucket placeholders, nested-parameter E3043), none on the edited regions.
- New `scripts/sdlc/tests/test_userpool_email_mutability.py` (5 tests): default, group + label, condition wiring, gated verification setting, headless removal. Full `scripts/sdlc/tests`: 551 passed.
- `lib/idp_cli_pkg/tests/test_deploy_params.py`: 29 passed (6 new for the create-time defaulting).
- `lib/idp_sdk/tests/unit -k "transform or headless"`: 38 passed (the dangling-reference test caught the condition on the first run).
- `make check-arn-partitions`, ruff, basedpyright: clean on changed files.

**Not verified live.** Needs a fresh stack created with `ExternalIdPEmailMutable=true` and `make verify-idp-federation` (see `.claude/skills/live-auth-checks.md`). That run also confirms the verification-before-update setting does not interfere with the federated email sync, which AWS docs describe only for user-initiated `UpdateUserAttributes`.

## Existing deployments

Unchanged and unaffected on update. They still need a new pool (new stack) to get repeat federated sign-in; `docs/external-idp.md` describes the one-shot `admin-delete-user` bridge and the `sub` re-keying caveat.

Closes #835


rstrahan:
## Live verification (2026-09-10, us-west-2)

Built this branch (`idp-835` prefix), created a fresh stack **IDPVerify** with `ExternalIdPEmailMutable=true` against the throwaway OIDC provider from `scripts/security/live_checks/oidc_provider/`, and ran `make verify-idp-federation`.

**Pool as created:** `email` → `Mutable: true, Required: true`; `UserAttributeUpdateSettings.AttributesRequireVerificationBeforeUpdate: ["email"]`.

**Result: 23/23 checks passed**, including the two this PR adds:

- **3b — second hosted-UI sign-in by the same federated user: PASS.** The sign-in that fails today with `user.email: Attribute cannot be updated` completes, and the re-asserted `IdP-Admins` claim is applied (role set `{'Admin'}`, stale `Viewer` removed).
- **3c — IdP-side email change applies on the next sign-in: PASS.** With the mock IdP asserting a different email, the third sign-in succeeded, the Cognito record's `email` became the new address, and the new token carries it. So `AttributesRequireVerificationBeforeUpdate` does **not** gate Cognito's attribute-mapping write — the interaction the review flagged as unanswerable from AWS documentation.

**Compensating control probed directly on the same pool** (not part of the harness): a native user's own access token called `UpdateUserAttributes` to set `email` to another address. The call was accepted and a verification code was dispatched to the *new* address, but the stored `email` and the `email` claim in a fresh ID token both remained the original value. The self-service change does not take effect until the new address is verified, as intended.

The `#835 review` follow-up commits on this branch address the review's findings (update guard in the SDK deploy path, doc examples trimmed, the scope-widening limitation documented, `== {"Admin"}` assertion, behavioural transform test, the 3c check). IDPVerify and the provider have been torn down.

PR #851, opened 2026-09-10
fix(config): lending Payslip descriptions name the ADP header codes so EmployeeNumber/PayrollNumber extract (#839)

### LINKED ISSUE #839: [Bug]: simple-mode extraction on the lending sample payslip returns null (or a hallucinated value) for EmployeeNumber / PayrollNumber that are plainly in the OCR text
## Summary

On `samples/lending_package.pdf`, the Payslip section (page 1) contains, as OCR'd Textract table rows:

```
| Employee Number | 00000000 |
| Payroll Number  | 126543   |
```

Simple-mode extraction with the shipped `lending-package-sample` config (Claude Sonnet 5, `temperature 0`) returns `null` for both `EmployeeNumber` and `PayrollNumber` most of the time, and occasionally a **wrong** `EmployeeNumber` (`'12345'`). Measured 2026-09-10 by running the real `ExtractionService.process_document_section` on the identical page data (same OCR output, same image) repeatedly:

| code | runs | both null | wrong `12345` | correct |
|---|---|---|---|---|
| v0.6.7 (`v0.6.7` tag) | 4 | 3 | 1 | 0 |
| develop (`580bfb5d1`) | 3 | 2 | 1 | 0 |
| live smoke runs on a deployed stack (v0.6.7 then upgraded to develop) | 4 | 3 | 0 | 1 |

So 1 correct answer in 11 attempts, identical across the two code versions; the merged default configs differ only in the classification boundary prompt. This is model behaviour on this page, present at v0.6.7, **not** a regression — it was noticed only because the pre-upgrade smoke run of the upgrade test happened to be the lucky sample and the three post-upgrade runs were not.

Each run is schema-valid with no processing issue (the population check passes because most fields are filled), and the separate confidence pass then explains the null as `"Not provided in document"` — so the miss is invisible unless you compare against ground truth. Coercion did not fire (`metadata.coercion` absent), so the nulls come from the model's response, not from post-processing.

## Why it matters

The lending package is the sample every first-time user processes, and a field that is literally a labelled table cell is the easiest possible extraction. Two candidate causes worth checking before blaming the model outright: (1) the Payslip class's attribute descriptions for these two fields (does `EmployeeNumber`'s description match the label "Employee Number"? is `PayrollNumber` described at all?), and (2) the OCR table row rendering — the value sits in a two-column key/value table that is rendered with `| key | value |` pipes; a key/value-table hint in the prompt or a non-table rendering may recover it. The forced-tool path (`extraction.forced_tool.enabled: true`) is a cheap third experiment: on the benchmark statement it changed nothing, but this shape (labelled scalar in a table) is different.

## Reproduce

Run the section once with the shipped config on the sample; or use the probe pattern from the 2026-09-10 release validation (`ExtractionService.process_document_section` on a `Document` whose page 1 points at an existing run's `pages/1/{image.jpg,rawText.json,result.json}`).

rstrahan:
## What the page actually says

`samples/lending_package.pdf` page 1 (the Payslip) has **no "Employee Number" or "Payroll Number" label anywhere** — confirmed on the page image and in the Textract LINE blocks. Its only identifiers are the ADP header row:

```
CO.   FILE     DEPT.    CLOCK   NUMBER
ABC   126543   123456   12345   00000000   1
```

So the issue's premise (labelled `| Employee Number | 00000000 |` rows in the OCR text) does not match the page; the OCR markdown renders exactly the header row above. With the shipped descriptions — "unique identifier assigned to the employee by the employer" and "payroll batch or sequence number" — the model has nothing to map the two fields to. A null is a defensible abstention and the occasional `12345` is the CLOCK column. Not a code defect; a description gap in the preset.

## Measured (real Simple-mode `ExtractionService.process_document_section`, real OCR + image from IDP1, Sonnet 5, temperature 0, 5 runs per arm)

| descriptions | both correct | `EmployeeNumber` values | `PayrollNumber` values |
|---|---:|---|---|
| shipped | **1/5** | 12345, 12345, 00000000, 00000000, 00000000 | null, 00000000, null, null, 126543 |
| + alternative labels and ADP column names | 3/5 | 12345, 00000000, 00000000, 00000000, 12345 | 126543 ×5 |
| + "NUMBER is the LAST of the five codes, never CLOCK" (**this PR**) | **5/5** | 00000000 ×5 | 126543 ×5 |

## Change

Both `lending-package-sample` and `lending-package-sample-govcloud` presets: the `EmployeeNumber` and `PayrollNumber` descriptions now name the alternative labels (`Employee No.`, `Emp #`, `File`, `File No.`, …) and, for ADP-style statements, which of the five header codes each field is. Config-only. Existing deployments pick it up by re-applying the preset or editing the two descriptions in the Web UI; the default record on a new stack carries it. CHANGELOG entry included.

Not measured: whether the added text changes any *other* Payslip field (the descriptions are field-local, so no mechanism for it), or the govcloud preset live (identical text, Bedrock GovCloud not exercised).

Closes #839


rstrahan:
## Converted to draft: the field-to-column mapping needs a ground-truth decision

Review found the mapping the first commit encoded — `EmployeeNumber` = NUMBER (`00000000`), `PayrollNumber` = FILE (`126543`), i.e. the values #839 expects — is the reverse of what ADP's header codes mean. On ADP earnings statements **FILE** is the employee's file number (ADP's per-employee identifier), **CLOCK** the badge number, and **NUMBER** the check/advice number for that payment (the specimen's `00000000` is a placeholder check number). No ground truth for this page exists anywhere in the repository, so the expected values in the issue were most likely an earlier model output adopted as baseline.

Measured on the real page (Sonnet 5, temperature 0, 5 runs each) — the model follows whichever mapping the description states:

| descriptions | `EmployeeNumber` | `PayrollNumber` |
|---|---|---|
| shipped | 12345 / 00000000 (mixed) | null ×3, 00000000, 126543 |
| issue's mapping, made explicit (first commit) | 00000000 ×5 | 126543 ×5 |
| **ADP's mapping, made explicit (this branch now)** | **126543 ×5** | **00000000 ×5** |

So the extraction defect is fixed either way; what ships must be the mapping we believe is *correct*, and that is a data decision, not a model one. This branch now encodes ADP's own semantics (`EmployeeNumber` = FILE, `PayrollNumber` = check/advice NUMBER). @rstrahan please confirm which mapping the lending sample's ground truth should carry before this merges — and if `00000000 / 126543` came from an authoritative source, say so and I will swap it back (one commit).

Also applied from the review: ASCII hyphens (the Simple-mode prompt renders the schema with `ensure_ascii`, so the model saw `—`), and the CHANGELOG now states the upgrade path (a stack still on the `default` profile receives the text via the stack-deployment revision; other profiles re-apply the preset). The `notebooks/examples/config/classes.yaml` teaching copy still carries the old text; left alone pending the mapping decision.

Issue #764, opened 2026-09-04
[Bug]: Re-running Discovery on an existing class silently erases its class-level x-aws-idp-* settings


rstrahan:
### What happens

Re-running Discovery on a document class that already exists in a configuration
profile **replaces the class dict wholesale**, so every class-level
`x-aws-idp-*` authoring key on that class is silently erased.

`lib/idp_common_pkg/idp_common/discovery/classes_discovery.py` builds a
`classes_by_id` map from the existing version's classes and then does:

```python
classes_by_id[new_class_id] = new_class   # line 540
```

`new_class` is what the discovery LLM produced. The only extension keys it ever
emits are `$id` and `x-aws-idp-document-type` (normalized at line 586). It does
not read, preserve, or merge anything the author had set on the existing class.
Note the surrounding code *does* carefully preserve unrelated classes (the
stale-id normalization at lines 526-538) — it is only the re-discovered class
itself that loses its hand-authored settings.

### What is lost

Any class-level key an author configured in the Schema Designer or by editing
YAML. Concretely, and each of these changes behavior:

| key | effect of losing it |
|---|---|
| `x-aws-idp-extraction-model` | class reverts to the profile-wide model |
| `x-aws-idp-extraction-escalation-model` | escalation stops happening |
| `x-aws-idp-extraction-system-prompt` / `-task-prompt` | per-class prompt tuning reverts |
| `x-aws-idp-confidence-threshold` | class reverts to the global threshold |
| `x-aws-idp-confidence-escalation-model` | confidence escalation stops |
| `x-aws-idp-document-name-regex` / `-document-page-content-regex` | deterministic classification hints gone → class is LLM-classified again |
| `x-aws-idp-page-types` / `-source-page-types` | page-type routing gone |
| `x-aws-idp-exclude-from-processing` / `-exclusion-reason` | an excluded class starts being processed again |
| `x-aws-idp-examples` | few-shot examples gone |
| `x-aws-idp-multi-instance` / `x-aws-idp-instance-array` | records 2..N silently stop being extracted (this is #715 / #565 re-opening itself) |

### Why it matters

The failure is silent and delayed. Discovery reports success, the class looks
right in the UI (its properties are there, possibly improved), and the
regression only shows up in the *next* document processed — as a different
model, a missing escalation, a re-included class, or dropped records. There is
no warning, no diff, and nothing in the saved config records that keys were
dropped.

[...]

Issue #715, opened 2026-08-31
[Feature]: Multi-instance sections — synthesize a List-of-Class schema for a flagged class (schema transform half of #565)


rstrahan:
Follow-up to #565. The **visibility** half shipped in PR #694; this is the automatic **schema transform** half, deliberately deferred to its own PR because its blast radius is completely different.

**Scope note (split as of 2026-09-03):** detecting the silent single-object multi-record case is **no longer part of this issue** — it is #753, and it is what #565 will be closed on. This issue is now purely the ergonomic transform: letting a user opt a class into a List-of-Class shape without hand-authoring the wrapper, and making every downstream stage work with it.

## What already works (shipped in #694)

- `Section.instance_count`, surfaced in the UI as a badge for `> 1`, plus `extraction_multi_instance_detected` (warning) and CloudWatch metrics.
- **Designate mode** — `x-aws-idp-instance-array: <propertyName>` lets a class that is *already* modelled as a packet of records name its own instance axis. Zero schema change, zero downstream impact.
- Recovery of a top-level JSON array instead of discarding it (#687).

Verified live: a forced-merge section over 3 pages produced `InstanceCount=3` with all three records extracted, per-item coercion, and clean validation.

## What this issue covers

**Synthesize mode** — `x-aws-idp-multi-instance: true` on a class, which replaces the *effective* class schema with:

```json
{ "type": "object", "title": "<Class> Instances",
  "properties": { "instances": { "type": "array", "items": { …original class schema… }, "minItems": 1 } },
  "required": ["instances"] }
```

so a user does not have to hand-author the wrapper and degrade a single-record class's schema.

**This is opt-in per class, NOT auto-detection.** Auto-detecting would move #565's detection problem one stage later and make every single-document section pay for a wrapper level. Detection is #753 and stays a warning there; it never flips this flag on a user's behalf.

## Why it is worth doing (and why it is not the urgent half)

Two independent live runs against a deployed stack (Sonnet 5, `sectionSplitting: disabled` forcing one section over a multi-record document, single-record class):

| Run | Result |
|---|---|
| 3-patient document | model returned **one object** — Alice Anderson only; Brian Baker and Carol Chen **silently lost**, `instance_count=1`, no warning |
| 2-patient document | model returned **one object** — Dana Doyle only; Evan Ellis **silently lost** |

[...]

PR #863, opened 2026-09-11
feat(discovery): suggest — never set — several documents per section when a sample holds several records of one class (#765)

### LINKED ISSUE #765: [Feature]: Discovery should suggest (never silently set) multi-instance when a sample holds several records of one class
### Background

#715 added multi-instance sections: a class flagged `x-aws-idp-multi-instance:
true` gets its schema wrapped so one section can carry several records of that
class, and `extraction.multi_instance_detection` (off by default) makes the
silent loss visible when the flag is *missing*. See
`docs/extraction-and-confidence.md` and
`docs/benchmarking/feature-multi-instance.md`.

Both are opt-in and neither is discoverable from the UI: nothing in the product
tells you that the document you just ran Discovery on is one of the documents
this feature exists for. You find out from missing rows.

### What Discovery could do

Discovery already reads the document. When the sample it analyzes contains
**several records of the same class** — three pay statements in one PDF, several
checks on one sheet, two bank statements back to back — it is in the best
position in the whole product to notice, because it is the only stage that sees
the pages and is *authoring* the schema at the same time.

So: **detect the candidate and surface it. Do not set it.**

### Why suggest and not set

Getting this wrong in either direction is costly, and the two directions are not
symmetric:

- Setting `x-aws-idp-multi-instance` when the class is genuinely single-record
  changes the shape of every result for that class (`{"instances": [...]}`),
  which invalidates committed evaluation baselines and any downstream consumer
  of the section's `inference_result`. There is a migration script
  (`scripts/migrate_multi_instance_baselines.py`) precisely because this is not
  a free change.
- Not setting it costs nothing until the user hits a multi-record document —
  and then #753's detection warning is there to catch it.

A wrong auto-set is also invisible in exactly the way #764 describes: the config
changes, nothing warns, and the effect shows up as a shape change in the next
run. Discovery inferring intent from one sample is not a strong enough signal to
silently rewrite a class's output contract.

There is also a real ambiguity Discovery cannot resolve from one document: three
pay statements in one file might mean "this class is multi-instance" *or* "this
packet should have been split into three sections". Those want different fixes
(`x-aws-idp-multi-instance` vs. section splitting), and only the user knows
which.

### Proposal

1. During Discovery, note when the analyzed sample appears to hold more than one
   record of the class being discovered — either by asking in the discovery
   request (the #753 probe wording in
   `idp_common/extraction/instance_probe.py` is already tuned for this: "count
   documents, not pages", which matters because repeated per-page headers read
   as multiple documents) or by the model reporting repeated identity blocks.
2. Return it as a **suggestion on the discovery result**, not a config write:
   "This sample appears to contain N Pay-Statement records. If sections of this
   class can hold several records, enable *Documents per section → Several* on
   this class."
3. Surface it in the Discovery UI next to the discovered class, with a one-click
   action that applies the setting — so the user makes the decision and can see
   what it changed.
4. Optionally mention the alternative (section splitting) so the ambiguity above
   is the user's call rather than a silent guess.

### Prerequisite

This depends on #764: today re-running Discovery erases class-level
`x-aws-idp-*` keys, so a one-click "enable it for me" action would be undone by
the next Discovery run on that class. Fix the merge first.

### Not in scope for #715

Filed as a follow-up. #715 ships the transform and the detection probe; this is
about making the feature *findable* rather than something you have to already
know about.

rstrahan:
Implements #765 (unblocked since #764 closed).

## Why

`x-aws-idp-multi-instance` (#715) and the detection probe (#753) are both opt-in and neither is discoverable from the UI: nothing tells you the document you just ran Discovery on is one this feature exists for. Discovery is the one stage that sees the pages *and* authors the schema, so it is the best place to notice "this sample holds three Pay-Statements" — and the wrong place to *act* on it silently, because enabling the flag changes the shape of every result for the class (invalidating baselines), and one sample cannot tell "multi-record class" from "packet that should have been split".

## What

**Detect (backend, `classes_discovery.py`).** Both single-document flows append `_INSTANCE_COUNT_INSTRUCTION` to the prompt: the model adds one top-level integer `IDPDocumentInstanceCount` (same field name and question as the #753 probe — "count complete documents, not pages, sections or repeated headers"). `pop_instance_count()` strips it (and removes it from `properties` if the model put it there) before validation, so it can never become a class field. The result dict gains `multi_instance_hint`: `None` for one record, otherwise `{instance_count, class_name, already_multi_instance, message}` built by `build_multi_instance_hint()`. `already_multi_instance` reflects the class *after* the merge, so a re-run on a class you already flagged (kept by the #764 carry-forward) reports the flag as on rather than asking again.

**Carry (processor → job → API).** `discovery_processor` passes `multiInstanceHint` (JSON string) through the `updateDiscoveryJobStatus` mutation and the DynamoDB fallback, and mentions it in the status message. `schema.graphql`: `multiInstanceHint: String` on `DiscoveryJobListItem`, `DiscoveryJob` and the mutation; `ddb_direct._DISCOVERY_OPTIONAL_FIELDS`; the two UI operation documents; `npm run codegen`.

**Surface + one-click apply (Web UI).** `DiscoveryJobDetails`: an info Alert *"This sample appears to contain N 'X' records"* explaining both options (Documents per section → Several, or section splitting) with an **Enable several documents per section** button. The action reads the job's configuration version (`getConfigVersion`), sets `x-aws-idp-multi-instance: true` on that class and clears any `x-aws-idp-instance-array` designation (mirroring SchemaInspector's Synthesize mode), writes the full `classes` list back (`updateConfiguration` replaces lists wholesale), and then reminds you to run `scripts/migrate_multi_instance_baselines.py`. Already-enabled and not-found are reported; no version → points at Configuration. `DiscoveryPanel`: blue *"N records in sample"* badge. Pure helper `multiInstanceHint.ts` (`parseMultiInstanceHint`, `applyMultiInstance`).

**Docs.** `docs/discovery.md` new section; `docs/extraction-and-confidence.md` "two known gaps" rewritten (both now closed); `idp_common/discovery/README.md`; CHANGELOG.

## Scope / not in this change
- Single-document discovery only. Multi-document (cluster) discovery and policy discovery are unchanged.
- SDK `DiscoveryResult` / CLI do not yet expose the hint (the field is on the job record for anyone reading it).
- No new config toggle: the question costs a few tokens and changes no schema.

## Tests
- `tests/unit/discovery/test_multi_instance_hint.py` (17): pop/coerce incl. misplaced property and bool/zero/fraction rejection; hint text names the class, the setting and the alternative; the prompt carries the probe wording; end-to-end flow with a mocked model returns the hint and a clean schema, one record → `None`, reply without the count still succeeds, carried-forward flag → `already_multi_instance`. Whole discovery suite: 193 passed.
- `src/ui/src/components/discovery/__tests__/multiInstanceHint.test.ts` (3). `tsc --noEmit` clean, eslint clean, `scan_api_rbac.py` 0 FAIL (1 pre-existing WARN).

Closes #765 (issue to be closed manually after merge).


rstrahan:
## Live check (requested in review): does a real model honor the instruction?

Ran `ClassesDiscovery` locally against Bedrock (default discovery model `us.amazon.nova-pro-v1:0`, default temperature 1.0, `save_to_config=False`, config store mocked) on two shipped samples.

| sample | truth | runs | count returned | key leaked into schema |
|---|---|---|---|---|
| `samples/paystub_multi_instance.pdf` | 3 records (`expected_instance_count: 3`) | 4 | **3, 3, 3, 3** → hint produced | never |
| `samples/bank-statement-multipage.pdf` | 1 document, several pages | 1 | **1** → no hint | never |

So the model puts the count where asked (top level, not in `properties`), gets it right on the multi-record sample, and does not confuse pages with documents on the multi-page one.

**Schema-quality control, same paystub sample, 3 runs per arm:**

| arm | properties per run | names seen in fewer than 3 of 3 runs |
|---|---|---|
| with the instruction | 12, 11, 12 | `Earnings` / `EarningsSummary` / `EarningsDetails`, `EarningsBreakdown`, `PayDetails`, two spellings of the PTO group |
| control (instruction removed) | 12, 12, 12 | none |

Honest reading: with-instruction runs showed property-count and naming variance that the three control runs did not, at temperature 1.0 and n = 3 per arm. That is suggestive of a small effect, not proof of one; recorded as a caveat in `docs/discovery.md`. If it matters in practice the natural follow-up is a `discovery.*` toggle, which I deliberately did not add here to keep the change additive (it would touch the config model, the UI config-form schema and the parity test).

Scripts used: `/tmp/live765.py`, `/tmp/live765_rep.py` (throwaway; not committed).

PR #864, opened 2026-09-11
fix(dynamodb): decrement VersionCount only when a run was actually deleted

art-aws:
**What.** `delete_document_run` issued `DeleteItem` and then decremented the document's `VersionCount` unconditionally. `DeleteItem` on a missing key succeeds silently, so a retried mutation, a double click on the Version History panel, or an unknown `run_id` sent straight to the API each took the counter one below the number of runs actually recorded — and nothing recomputes it, so the drift is permanent.

**Fix.** Ask for the deleted item back (`ReturnValues=ALL_OLD`) and decrement only when there was one; return `False` otherwise so the caller can tell. `DynamoDBClient.delete_item` gains an optional `return_values` parameter for that; existing callers are unaffected. This mirrors the at-least-once discipline `create_document_run` already applies on the way in, where a duplicate `run_id` is a no-op rather than a second increment.

**How it was found.** Reviewing the run-record lifecycle while consolidating multiple processing runs of the same document from an external pipeline, where deletes and re-runs can arrive more than once.

**Tests.** Two unit tests in `tests/unit/dynamodb/test_service_document_runs.py`: a real delete decrements once and passes `return_values="ALL_OLD"`; a missing run neither decrements nor claims success. `132 passed` across the dynamodb, document-versions, versions-resolver and delete suites; `ruff` clean.

**Changelog.** One `Fixed` line under `[Unreleased]`.

**Not in this change.** The resolver still lists runs before deleting (to protect the newest); it does not additionally pre-check existence, since the service now reports it.

Co-authored-by: Diego F Galaviz <digalavi@amazon.com>


art-aws:
Closing this myself after a second look: the only caller, `delete_document_version` in the versions resolver, already does `get_document_run` and raises `Version not found` before reaching `delete_document_run`, so the missing-run path is not reachable from the product, and nothing reads `VersionCount` beyond exposing it in the schema. What remains is a theoretical double-decrement under two concurrent deletes of the same run, on a counter with no consumer — not worth a client API change the night before a release. Sorry for the noise.

Issue #782, opened 2026-09-05
Advanced extraction fabricates a value where it cannot abstain: a required number cannot be null, so an unreadable cell becomes 0.0


rstrahan:
## Summary

On the **Advanced (agentic)** extraction path, a `required` attribute cannot be returned as
`null`. The class schema is enforced through a generated Pydantic model, and
`datamodel-code-generator` renders a required property as a **non-nullable** field with no
default. So when the model genuinely cannot read a value, the extraction contract gives it
no way to say so — and the retry loop then explicitly instructs it to make the validation
error go away.

For a `type: number` attribute the result is a **fabricated `0.0`** that is schema-valid,
passes JSON-Schema validation, and is indistinguishable from a real zero.

**Simple mode is not affected** — it emits plain JSON, so `null` survives and
`extraction.validation` (on by default since v0.6.7) reports it loudly.

## Observed

Measured during v0.6.7 release validation, on a document whose Amount column was
physically unreadable (the corpus rendered the description text over it):

| mode | what came back for all 100 rows | validation | visible? |
|---|---|---|---|
| simple | `"Amount": null` | 100 × `'Amount' is a required property` | **yes, loudly** |
| advanced | `"Amount": 0.0` | 0 errors, section `valid` | **no** |

`completeness_recall` was 1.000 for both (it counts rows, and the row tag lives in the
Description field), and `scalar_accuracy` was 1.000 for both (it only reads document-level
fields). Nothing except a per-row typed-value check could see it.

## Root cause

`Transaction.Amount` is in `required` and typed `number`, so the generated model field is
`Amount: float` — `required=True`, `default=PydanticUndefined`, not `Optional`.

`agentic_idp.py::invoke_agent_with_extraction_retry` then does:

```python
result = data_format(**current_extraction)     # Pydantic enforcement
...
except Exception as e:
    prompt_content = [ContentBlock(
        text=f"The extraction failed validation with error: {str(e)}. "
             f"Please fix the extraction using the tools.")]
    continue                                    # up to max_extraction_retries (3)
```

So an agent that abstains with `null` is told its answer is wrong and asked to fix it. The
only value that satisfies `float` is a number.

This **directly contradicts the table-tool prompt**, which instructs the opposite
(`tools/table_parser.py`):

> *"If ONE column is unreadable, garbled, or cannot be mapped, still emit EVERY row. Set
> that single cell to null, or to the literal text you can see."*

[...]

PR #867, opened 2026-09-11
feat(extraction): multi-instance detection probe on the unsharded agentic path (#772, option 1)

### LINKED ISSUE #772: [Feature]: extend multi-instance detection (#753) to Advanced (agentic) extraction
### Summary

`extraction.multi_instance_detection` (#753) is skipped when Advanced (agentic)
extraction is enabled, so a section that silently drops records of the same class
raises **no warning** in that mode. This is a documented gap, not a deliberate
scope decision — and it lands in the mode that is recommended for exactly the
corpora most prone to the problem (large multi-page tables: bank statements,
brokerage statements, transaction logs).

**The #715 fix is unaffected.** `x-aws-idp-multi-instance` lives in
`_get_class_schema`, which both modes share, so a flagged class extracts every
record in agentic mode too. It is only the *detector* — the thing that tells you a
class needs the flag — that is missing.

```python
# extraction/service.py::_build_wire_schema
if self.config.extraction.agentic.enabled:
    return class_schema, False
```

### Why it isn't a one-line change

Two real obstacles, both in the agentic path:

1. **The path is Pydantic-model-driven, not schema-dict-driven.** Simple
   extraction puts the probe on the wire by adding one property to a copy of the
   JSON Schema, which flows into the prompt text and the forced-tool `toolSpec`.
   Agentic instead does
   `create_pydantic_model_from_json_schema(schema=self._class_schema, …)` and
   passes the resulting class as `structured_output_async(data_format=…)`. A
   property added to the schema copy never reaches the wire. The probe would have
   to be generated **onto the model** and then stripped off the validated object
   before it reaches `inference_result`.

2. **Agentic shards by page range**, so each shard would answer the count for its
   own pages. Neither obvious reconciliation is correct:
   - **sum** double-counts a document that spans a shard boundary;
   - **max** under-counts a section whose records are spread across shards.

   So the reconciliation rule is a genuine design decision. Note the sharding is
   by *page range* — an earlier version of the `_build_wire_schema` docstring said
   "by field", which was wrong.

### Options, cheapest first

1. **Probe only the unsharded agentic path.** When `_build_shard_payloads`
   produces no shards (the common case for a section short enough), add the probe
   field to the generated model and strip it. Sidesteps obstacle 2 entirely and
   covers most agentic sections.
2. **Probe every shard and report a lower bound.** Take the max across shards and
   label it as such, so the warning says "at least N" rather than "N". Weaker
   claim, full coverage, no wrong arithmetic.
3. **Probe once with a page-range-independent question.** Ask each shard for the
   identity of the first and last record it sees rather than a count, and
   reconcile by de-duplicating identities across boundaries. Most accurate, most
   work.

(1) is the recommended first step; it is measurable against the same corpora used
in the original study.

### Testing note

The existing benchmark suite already has the control for this: `migate` includes
`core-tt-adv-sep` precisely because the probe is *not* on the agentic path today,
so a delta there means something leaked. Whichever option is taken, that cell
becomes the arm to measure rather than the control, and the accuracy check from
the original study should be repeated — detection cost ~1.3 accuracy points on a
corpus with nothing to find, and there is no reason to assume agentic behaves the
same way.

### Docs to update when this changes

- `docs/extraction-and-confidence.md` — the blast-radius table row currently reads
  "Advanced (agentic) extraction | the wrapper applies, but the #753 detection
  probe does not".
- `lib/idp_common_pkg/idp_common/extraction/README.md` — "Simple extraction only
  (prompt and forced-tool paths) — Advanced (agentic) extraction is not covered".
- `docs/benchmarking/config-guidance.md` §7.
- `base-extraction.yaml` — the `multi_instance_detection` comment says "Simple
  extraction only (prompt and forced-tool paths)".

rstrahan:
Implements **option 1** of #772 — probe only the unsharded agentic path — which the issue recommends as the first step because it sidesteps the shard-reconciliation decision and covers the common case.

## Why

`extraction.multi_instance_detection` (#753) was skipped whenever Advanced (agentic) extraction was enabled (`_build_wire_schema` returned early), so a section that silently dropped records of one class raised **no warning** in exactly the mode recommended for the corpora most prone to it. The #715 wrapper was never affected — only the detector was missing.

## What

**`_agentic_probe_model(dynamic_model, class_label, resuming)`** — for the single-agent call, when detection is enabled and the run is not a resume, builds the transport model from a probe-augmented copy of the effective class schema: the *same* `augment_schema_with_probe` the Simple path uses, fed through `_transport_model` (so the #782 nullable-leaves rule applies unchanged). Returns the plain model otherwise. Records `_instance_probe_requested` like the Simple path.

**Pop before anything downstream.** Right after `structured_data.model_dump()`, `_read_instance_probe` removes the answer and sets `instance_probe`, which flows into the existing `_resolve_instance_reporting` → `extraction_multi_instance_suspected` warning. The probe key is also removed from `metering["_integrated_field_assessment"]` so it cannot acquire a confidence entry. Validation/escalation and the existing shared probe read (which now finds nothing to pop) are unchanged.

**Not probed:** sharded sections (each shard would answer for its own pages; sum double-counts a document spanning a boundary, max under-counts records spread across shards — options 2/3 of the issue remain open) and resumed runs (their `existing_data` was validated against the plain model).

**Docs.** The four sites the issue lists (`docs/extraction-and-confidence.md` blast-radius row, `extraction/README.md`, `base-extraction.yaml` comment, `config-guidance.md` §7) plus the `config_matrix.yaml` `migate` comment: `core-tt-adv-sep` was the control *because* the probe was not on the agentic path; it is now an arm. `_build_wire_schema`'s docstring updated. CHANGELOG entry.

## ⚠️ Measurement not done — decision for the reviewer

The issue asks for the accuracy A/B to be repeated: Simple-mode detection cost **−1.3 accuracy points** on RealKIE-FCC-Verified (a corpus with nothing to find), and there is no reason to assume agentic behaves the same. Running it needs this code deployed to a stack (a branch build on IDP1 or a throwaway stack) and ~40 documents × 2 arms under an agentic profile. I did **not** deploy a branch build to the shared stack unasked. The CHANGELOG and §7 say explicitly that the agentic cost is unmeasured; if you want the A/B run before merge, say so and I will do it (`detection_ab_teststudio.py` with an agentic profile pair, or the `migate` suite before/after on one stack).

## Tests
`tests/unit/extraction/test_agentic_instance_probe.py` (5): the probe is an optional integer on the generated model and a reply without it validates; the answer is popped from the dumped fields and read as the count with declared fields intact; disabled and resume keep the plain model; the configured question is the probe description; the Simple-path helper still returns the schema unchanged under agentic. Whole extraction suite: 1126 passed. ruff clean.

Closes #772 (option 1; issue to be closed or re-scoped manually after merge).


rstrahan:
Status: review findings addressed in `18e278bc2` (in-loop validator now ignores the probe on strict schemas, with a test; stale comments updated; mock-based wiring test through `_invoke_extraction_model`). CI green.

**Deliberately not merged.** The issue's own status note defers this to a later release, and the agentic accuracy cost is unmeasured (Simple-mode detection cost −1.3 points on a corpus with nothing to find). Detection stays `enabled: false` by default, so merging is safe in the narrow sense — but whether it goes in now, and whether the A/B (`detection_ab_teststudio.py` with an agentic profile pair, or `migate` before/after on one stack) runs first, is your call. Say which and I will do it.

rstrahan:
## Agentic detection A/B — measured on a throwaway stack (IDP772AB, this branch's build)

Test Studio runner, two profiles per corpus derived from the corpus's own shipped config with `extraction.mode: advanced`, differing only in `multi_instance_detection.enabled`; 40 paired documents per arm (`numberOfFiles` takes the first 40, so both arms see identical documents). **Verified from section metadata**: every section in every arm ran `extraction_method: agentic`, unsharded; the probe was answered in the on-arms (RealKIE 76/100 sections, OCR benchmark 40/40) and absent in the off-arms.

| corpus (extraction model) | accuracy off → on | paired delta | better / worse / same | sign test | extraction tokens (on vs off) | extraction cost/doc |
|---|---|---|---|---|---|---|
| RealKIE-FCC-Verified (Sonnet 5) | 0.7716 → 0.7567 | **−1.5 pts** | 7 / 11 / 22 | p = 0.48 | uncached in −30%, cache read −5% | $0.473 → $0.416 (−12%) |
| OCR benchmark (Nova 2 Lite) | 0.9543 → 0.9458 | **−0.9 pts** | 3 / 9 / 28 | p = 0.15 | uncached in +25%, cache read +25%, out +11% | $0.0085 → $0.0096 (+12%) |

- **Accuracy:** the same size as the Simple-mode cost measured for #753 (−1.3 pts), but at n = 40 neither corpus's delta is statistically distinguishable from zero (the Simple-mode result was p = 0.001 on the same corpus).
- **False positives:** `extraction_multi_instance_suspected` raised 0 of 80 documents across both corpora (neither corpus has a multi-record document).
- **Positive control:** `samples/paystub_multi_instance.pdf` (three records) under the lending default + advanced + detection on: probe answered 2, warning raised ("appear to contain 2 separate 'Payslip' documents, but extraction returned 1"). Undercounts 3 as 2, but flags.
- **Tokens:** on the OCR benchmark the agent's extraction input grew ~25% — the probe rides in the tool schema and its system-prompt restatement on every agent turn, so the overhead multiplies with turns. On RealKIE the on-arm used *fewer* tokens; agent turn-count variance at n = 40 dominates there and I would not read a direction into it.

Recorded in the branch's CHANGELOG entry, `config-guidance.md` §7 and the `migate` matrix comment (`9c2470f6e`). Raw analysis: `/tmp/ab772/agentic-results.txt` on the build host.

**Two develop-level gaps hit on the way, not caused by this branch** (worth issues): (1) a profile uploaded with `idp-cli config-upload` has no revision body in S3, so a run pinned to its revision 1 (the harness's default) fails in OCR with an S3 AccessDenied that is really a missing object; (2) test-run ids are `<set>-<timestamp to the second>`, so two arms launched within one second collapse into one run — the first RealKIE pair did exactly that and was relaunched with a 3 s gap.

rstrahan:
## Off hold — merging

The hold rested on two things, and both are now resolved.

1. **Release timing.** Issue #772 said this work was "not targeted at the current release" (0.6.7). 0.6.7 has since been tagged and `develop` is at 0.6.8.dev4, so this lands in exactly the subsequent release the issue deferred it to.
2. **Unmeasured agentic accuracy cost.** Measured 2026-09-11 on a throwaway stack (comment above): −1.5 pts on RealKIE-FCC-Verified (Sonnet 5, p = 0.48) and −0.9 pts on the OCR benchmark (Nova 2 Lite, p = 0.15), neither distinguishable from zero at n = 40; 0/80 false positives; the three-record paystub positive control flagged. Same size as the Simple-mode cost already accepted for #753. Detection ships `enabled: false`, so nothing changes for anyone who does not opt in. The numbers are recorded in the CHANGELOG entry, `config-guidance.md` §7 and the `migate` matrix comment.

**Merge check.** The only conflict with current `develop` is CHANGELOG ordering (both sides add entries at the top of Unreleased → Added). `service.py` auto-merges despite ~120 lines of unrelated change there since the branch point (#856 shard budget, #871 prompt cache). Verified on a scratch merge of this branch onto `develop` `679a6703a`: extraction unit suite 1141 passed / 2 skipped, ruff lint + format clean.

**Follow-ups after merge.** #772 stays open and is re-scoped to options 2/3 (probing sharded sections) rather than closed — option 1 covers only the unsharded call. The ~25% agent input-token overhead on the OCR benchmark (probe restated every agent turn) is why the docs say to run detection as a one-off diagnostic, not leave it on; no default or guidance change on that basis. The two develop-level gaps the A/B surfaced were filed and fixed as #878 and #879.

Issue #759, opened 2026-09-03
The review-effort estimate always reads the test set's combined confidence curve, never a per-configuration one


rstrahan:
Split out of #698, which turned out to be a refinement of a distinction the estimate does not yet make. Pre-dates Configuration Profile revisions (#689/#691) by a month — it arrived with the confidence-curve feature itself in `96a49ba1c` (2026-07-31).

## The problem

Test Studio's review-effort estimate rests on a confidence→accuracy curve. Curves are *written* per configuration:

```python
# lib/idp_common_pkg/idp_common/evaluation/curve_store.py
def curve_sk(config_version): return f"{CURVE_SK_PREFIX}{config_version or AGGREGATE_KEY}"
```

`add_observations` and `add_ece_bins` both write the config-specific key *and* the set's aggregate. `get_curve` then reads the config-specific key, falling back to the aggregate — a fallback the module docstring describes as the cold-start path:

> An unknown config version falls back to the test set's aggregate curve, then to the global prior, so a cold start still gets an answer.

**In practice the fallback is the only path**, because the reader never has a config version to pass:

- `estimate_review_effort` takes it from `meta.get("boundConfigVersion")` (`test_set_resolver/index.py:1908`), and **`boundConfigVersion` has never been written in any commit in the repo's history**. `git log -S` finds it four times: a *read* added in `4d67e5bd0` (test-set versioning), an ERD line in `0e5aa51dc`, a second *read* in `96a49ba1c`, and prose in the config-profiles commits. It is a designed-but-never-populated field.
- No UI caller supplies one: `TestSetDetail.tsx:781` renders `ReviewEffortModal` without the prop, and `AnnotationWorkspace.tsx:177` / `TestSets.tsx:169` pass only `testSetId`. The GraphQL argument exists and is always omitted.

So every estimate reads `curve#_aggregate` — the mixture of every configuration the set has been labeled or scored under.

## Why it matters

The per-config curve exists precisely because confidence means different things across models and prompts. Blending them is the failure the keying was built to prevent, and it is silent: the aggregate curve is genuinely populated, so `estimateConfidence` can report `measured` and `quality_tier` can return **gold** for a relationship averaged over configurations that no longer exist. A wrong `docsToReview` here means a team either under-reviews a set they believe is calibrated, or over-reviews one that is.

[...]

PR #869, opened 2026-09-11
feat(test-studio): key confidence curves by revision fingerprint at all three sites (#698)

### LINKED ISSUE #698: Enhancement: key confidence curves by revision fingerprint, not by profile alone
Follow-up to #691, and the one item in `docs/proposals/config-profiles-and-revisions.md` §6 that was consciously **not** implemented with the rest of Phase 2.

## The problem

Test Studio's review-effort estimate measures a confidence→accuracy curve per configuration, because confidence means different things across models and prompts. Curves are keyed by **profile**:

```python
# lib/idp_common_pkg/idp_common/evaluation/curve_store.py
CURVE_SK_PREFIX = "curve#"
def curve_sk(config_version): return f"{CURVE_SK_PREFIX}{config_version or AGGREGATE_KEY}"
```

Now that a profile has revisions, all of its revisions share one curve. That is **right** for a prompt tweak — observations stay comparable and keep accumulating — and **wrong** after a model swap or an assessment-configuration change, which changes what a confidence number means. The estimate would then report a relationship measured under a configuration that no longer exists, with no signal that it had happened.

## Why partial wiring is worse than none

Three call sites touch a curve, and they do **not** have equal information:

| Call site | Knows the pinned revision? |
|---|---|
| `test_execution_aggregation_function` (scoring run) | **yes** — the run pins it |
| `complete_section_review` (reviewer saves labels) | no — only `metadata.config_version` |
| `test_set_resolver::estimateReviewEffort` | no — only `boundConfigVersion` |

Fingerprinting only the site that has the revision would send scoring observations to `curve#<profile>@<fingerprint>` while review observations kept landing on `curve#<profile>`. The estimate reads one key, so it would **stop seeing scoring data** — a regression against today, where both sources land together. Hence: all three, or none.

## What already exists

Every revision records a `confidenceFingerprint` — a hash of only the confidence-relevant configuration (extraction model and sampling parameters, assessment settings), computed by `idp_common/config/revisions.py::confidence_fingerprint`. A revision that edits only a classification prompt keeps the same fingerprint; a model swap does not. So the data needed to branch curves correctly is already being captured on every save.

## Proposed shape

1. Carry the revision into **label metadata** when labels are written, so a reviewer's save knows which revision produced the values it is correcting.
2. Carry it into the **estimate's** inputs (the test set's bound profile plus revision, or an explicit argument).
3. Switch `curve_sk` to `curve#<profile>@<fingerprint>`, with the plain `curve#<profile>` key retained as a **read fallback** so existing curves are not orphaned and a cold start inherits what was measured before.
4. Validate against a test set with **real** review observations and at least one scoring run — this is statistical behaviour, and a green unit test proving the key changed does not prove the estimate got better.

## Interim mitigation (already documented)

`docs/test-studio.md` and `docs/configuration-profiles.md` tell operators to reset the affected curve (`CurveStore.reset`) after changing the extraction model or assessment configuration. That is a manual step, which is exactly why this issue exists.

rstrahan:
Implements #698, unblocked by #853 (#759).

## Why

Test Studio's review-effort estimate rests on a confidence→accuracy curve keyed by Configuration Profile. Now that a profile has revisions, all of them shared one curve: right for a prompt tweak (observations stay comparable), wrong after a **model swap or assessment change**, which changes what a confidence number means — the estimate then reported a relationship measured under a configuration that no longer existed, with no signal.

The issue's constraint: **all three call sites or none.** Fingerprinting only the site that knows the pinned revision would send scoring observations to `curve#<profile>@<fp>` while review observations kept landing on `curve#<profile>`, and the estimate would stop seeing half the data.

## What

**Key scheme (`curve_store.py`).** `curve_sk(config_version, fingerprint)` → `curve#<profile>@<fingerprint>`; `parse_curve_sk`. `_merge_all` folds every observation into the revision curve (when the fingerprint is known), the profile's **pooled** curve `curve#<profile>`, the set aggregate and the global prior, so the fallbacks keep accumulating. `get_curve(test_set_id, config_version, fingerprint)` reads revision → pooled → aggregate and reports `served_from` (`revision` is new; `config` now means "pooled across revisions"). `list_curves` and `reset` understand the fingerprint. `ConfidenceCurve.confidence_fingerprint` added.

**Stamp (test runner).** `_confidence_fingerprint_of(config)` computes `confidence_fingerprint` from the configuration body the runner already captures and stamps `ConfidenceFingerprint` on the run item. The runner is the only Lambda in the loop with configuration access; the value is recomputed from the body, never read from the revision index, so the pre-normalization (#758) caveat does not apply. Best effort — never fails a run.

**Carry (harvest).** `_write_draft_labels_for_doc` sets `metadata.confidence_fingerprint` beside `config_version` from the run item (setdefault, pipeline value wins).

**Read — the three sites.**
- `test_execution_aggregation_function`: `run["ConfidenceFingerprint"]` → `add_ece_bins(..., fingerprint=)`.
- `complete_section_review`: `previous["metadata"]["confidence_fingerprint"]` → `add_observations(..., fingerprint=)`.
- `test_set_resolver.estimate_review_effort`: new `_resolve_set_fingerprint(test_set_id, meta, config_version)` reads the drafting runs (every `labeljob#` item, legacy `labelJobId` fallback): one fingerprint → `drafting-run`; several → `mixed-revisions` (pooled curve served); none → runs predate the stamp. `get_curve(..., fingerprint)`; the pooled/aggregate fallback is logged and returned.

**API/UI.** `ReviewEffortEstimate.confidenceFingerprint` / `confidenceFingerprintSource`; `curveSource` documents `revision`. `api_validation_spec.json` regenerated; UI query + codegen; `ReviewEffortModal` notes: measured for the current settings / pooled after a model swap / drafted under several revisions.

**Docs.** `docs/test-studio.md` and `docs/configuration-profiles.md` blockquotes rewritten (the "treat the estimate as unreliable after a model swap" limitation is gone; the estimate now says what it used), evaluation README, config README, CHANGELOG.

## Not in this change
- No IAM changes (by design — the fingerprint rides on records the sites already read).
- Runs recorded before the stamp carry no fingerprint and keep contributing only to the pooled curve; no backfill.
- No reset surface (`CurveStore.reset` still has no caller); unchanged from today.
- No live statistical validation on a set with real review observations and a scoring run (issue step 4). The wiring is unit-tested end to end at each site; the behaviour on a real set is what a reviewer may want to see before relying on the tier.

## Tests
- `tests/unit/evaluation/test_curve_store.py` (+6): key scheme/parse, observation lands on revision + pooled + aggregate + prior, model swap → new curve with pooled fallback reported, reads without a fingerprint unchanged, `list_curves` reports the fingerprint, `reset` targets one revision curve.
- `tests/unit/test_curve_fingerprint_plumbing.py` (3): runner stamps the fingerprint of the captured config; prompt-only edit keeps it, model swap changes it; missing/broken config → none, no failure.
- `tests/unit/test_test_set_resolver.py` (+3): estimate serves the revision curve when stamped and reports the pooled fallback before; `mixed-revisions`; harvest copies the fingerprint onto the label.
- `tests/unit/test_review_curve_feedback.py` (+2), `tests/unit/test_test_execution_aggregation.py` (+1, fakes updated).
- Affected suites: 725 passed. `tsc --noEmit` clean; test-studio vitest 173 passed; ruff clean on changed files (two pre-existing findings in the resolver untouched).

Closes #698 (issue to be closed manually after merge).


rstrahan:
Review addressed in `1ce7b7069`, and the branch is **rebased onto `develop`** — it had been cut from the #772 branch by mistake, so the earlier diff carried PR #867's commits; it now holds only the #698 work (22 files in the base commit).

- `_resolve_set_fingerprint` now returns `(None, "partial")` when some drafting runs of the profile carry a fingerprint and some predate the stamp (the upgrade shape: 199 pre-stamp documents plus one re-extract). A missing fingerprint reads as *unknown*, never as "the same family"; the pooled curve is served and the API/modal say so. Test added.
- `parse_curve_sk` splits from the right; the profile-name comment points at the actual validators; `list_curves` docstring fixed; redundant modal guard removed.
- Docs: one-line "treat the estimate as provisional while the pooled note shows" in both doc tiers; proposal doc marked implemented (pooled key still written).
- Tests: Decimal-ised body hashes like the float body; scoring and review observations for one revision land on one key (the all-three-sites invariant), in the curve-store suite against a real moto table.

The `artfork` remote with the embedded token has been removed from this checkout — thank you for flagging it.

Issue #937, opened 2026-09-17
Refresh docs/well-architected.md and give customers a review checklist they can fill in


rstrahan:
## Problem

`docs/well-architected.md` is 188 lines, sidebar-linked, and contains **zero numbers**. Its six
"Consider implementing..." recommendations cannot be tracked or falsified, and at least two of them are
already satisfied by the current codebase:

- "Add automated integration tests" — an integration suite exists and runs in GitLab CI
- "Implement distributed tracing" — X-Ray is instrumented across the pipeline Lambdas

It also contains at least one stale factual claim: `:43` states that WAF protects the AppSync API, and
there are zero `AWS::AppSync` resources remaining anywhere (see #929).

As published it reads as a design review rather than an operational-readiness review, and it is one of
the weakest documents in the tree while being one of the more prominent.

## Constraint on the fix

This document is a **guide for customers to use** on their own deployments. It must not be filled in with
this projects owners, dates, or internal review status — those belong to the customer, not to us.

## Fix

1. Correct every stale factual claim so the six-pillar assessment describes what the solution actually
   does today, with concrete references to the mechanisms that implement each pillar (the alarm set, the
   dep-audit and SRT gates, the KMS CMK and TLS-enforcing bucket policies, the permissions-boundary
   threading, the retry and DLQ posture, the cost instrumentation).
2. Replace the untrackable "Consider implementing..." bullets with a **customer-fillable checklist
   template** — per pillar, a table of review items with empty columns for the customers own comment,
   owner and date, plus guidance on how to use it in their own Well-Architected review.
3. State plainly which items the solution provides out of the box versus which are the customers
   responsibility in their account (for example: subscribing an address to `AlertsTopic`, setting a
   budget, enabling MFA on the user pool, configuring WAF ranges).

rstrahan:
Fixed on `develop` in #951. `docs/well-architected.md` has been corrected and now carries a customer-fillable review table. Every numeric claim on the page is re-derived from the templates by `scripts/tests/test_well_architected_doc.py` rather than restated, which has already caught the figures going stale three times as other changes landed.

Verified against the merged tree at `0b25b3ec2` rather than from the pull request description. Still open here only because a `Closes` keyword fires on merge to the default branch (`main`) and this merged to `develop`.

PR #959, opened 2026-09-18
docs: stop documenting AWS AppSync as current, and gate it (#929)

### LINKED ISSUE #929: Documentation asserts AppSync is present in ~26 places after it was fully removed
## Problem

There are **zero** `AWS::AppSync` resources anywhere in `template.yaml`, `patterns/unified/`, or any
`nested/*` template — the REST migration removed the service entirely, and
`docs/migration-appsync-to-rest.md` states "There is no toggle and no AppSync footprint remaining."

30 documents still carry 144 mentions, roughly 26 of them present-tense assertions that the service is
part of the running system, including:

- `docs/architecture.md:78,157` — "AppSync GraphQL API connects the UI to backend services"
- `docs/web-ui.md`, `docs/rbac.md`
- `docs/aws-services-and-roles.md:51,113`
- `docs/well-architected.md:43` (claims WAF protects the AppSync API)
- `docs/circuit-breaker.md:186` — instructs operators to set an `AppSyncVisibility` parameter **that no
  longer exists in any template**

`CLAUDE.md` contradicts itself within one file: line 448 says AppSync was removed, while lines 268, 357,
381 and 525 list it as present.

`docs/architecture.md` is additionally stale on the pattern consolidation: it still sells "multiple
processing patterns" at `:44-49`, says "all 12 Lambda functions" at `:115` where the unified pattern has
19, and documents only the superseded `PostProcessingLambdaHookFunctionArn` EventBridge mechanism at
`:221-235` with no mention of the pipeline-hooks dispatcher that is now the primary extension point.

Residual non-doc debt: `iam-roles/cloudformation-management/IDP-Cloudformation-Service-Role.yaml:126`
still grants `appsync:*`, and `scripts/sdlc/cfn/codepipeline-s3.yml:453,512-514` still references it.

## Why this matters more than ordinary doc rot

`docs/architecture.md` is the first document a new engineer or a customer architect reads. Three
extension mechanisms now exist — the EventBridge `PostProcessingLambdaHookFunctionArn`, the
config-driven pipeline-hooks dispatcher, and `model_id: "LambdaHook"` inference substitution — and the
architecture doc describes only the deprecated one. An extension author will build against the wrong
mechanism.

## Fix

1. Add a **retired-service gate**, generalising the existing
   `scripts/sdlc/tests/test_retired_models_not_offered.py` pattern: fail when a published doc asserts a
   removed service or references a CloudFormation parameter that no template declares. Seed it with
   AppSync and `AppSyncVisibility`.
2. Fix everything it flags, including the four self-contradicting lines in `CLAUDE.md`.
3. Rewrite `docs/architecture.md` around the current topology, and add a section naming all three
   extension mechanisms with guidance on which to use when.
4. Remove `appsync:*` from the service role and the stale CodePipeline references.

rstrahan:
Fixes #929

## Problem

The web UI used to talk to AWS AppSync. AppSync was removed in 0.6.0 — it is unavailable in GovCloud and not FedRAMP-compliant — and replaced by an API Gateway REST API (`HttpApi`, a REST API v1 chosen because only REST supports the `PRIVATE` endpoint type and WAFv2 on the stage) with a single `POST /op/{field}` route, a `COGNITO_USER_POOLS` authorizer for authentication, and an `AWS_PROXY` integration to a dispatcher Lambda. GraphQL subscriptions were replaced by polling. The GraphQL schema survives only as a codegen'd typed contract: `nested/api-resolvers/src/api/schema.graphql` feeds UI codegen and the build-time `api_validation_spec.json` the dispatcher validates against. No template declares an `AWS::AppSync::*` resource, and the `AppSyncVisibility` / `UsePrivateAppSync` parameters were renamed to `ApiGatewayVisibility` / `UsePrivateApi`.

The prose describing all of that was never swept. On the tree this PR forks from, the new gate reports **172 documentation lines across 47 documents** that still presented the GraphQL endpoint, its subscriptions or the old parameter names as things a deployment has. `CLAUDE.md` listed AWS AppSync under "Key AWS Services Used" a few hundred lines above its own correct explanation that AppSync had been removed, so a reader could get either answer from one file depending on where they looked. Nothing caught this, because removing a service is a template change and the prose describing it lives elsewhere.

## Counts, and how to reproduce them

Every number below is measured by the gate itself over the surface it scans: 215 Markdown files, the `scannedPaths` globs minus `excludedPaths`. The unit is a **line** that matches the registry's AppSync pattern, which is what the gate reports; a line with two mentions counts once.

```bash
mkdir -p /tmp/pre-929
git archive c33e0725d^ | tar -x -C /tmp/pre-929
python scripts/sdlc/check_retired_services.py --root /tmp/pre-929   # exit 1
python scripts/sdlc/check_retired_services.py                       # exit 0
```

| Tree | Lines matching the pattern | Reported as presenting AppSync as current | Exempt: allowlist entry | Exempt: historical marker |
|---|---|---|---|---|
| pre-fix (`c33e0725d^`, i.e. `fac1c120b`) | 246 in 54 files | **172 in 47 files** | 67 | 7 |
| this branch | 137 in 29 files | **0** | 71 | 66 |

Both rows are measured with the *final* registry, so the 172 are precisely the lines that are neither marker-covered nor allowlisted — the exemptions are held constant and only the prose changes. Of those 172, 109 lines lost the mention altogether (246 − 137), and the remaining 63 were rewritten so the mention became either explicitly historical (marker coverage rises by 59) or a named identifier whose retention the surrounding prose now explains (allowlist coverage rises by 4, without any new allowlist entry — the list is 11 entries on both rows). One caveat if you run the command above: against the pre-fix tree the gate also reports one *stale* allowlist entry — the `docs/architecture.md` entry whose `linePattern` is the phrase "historically named nested/appsync/" — because the text that entry excuses is text this PR adds. That is the dead-entry check working, not a second class of finding.

A find-and-replace would have broken two of the three categories, so every mention was classified before being edited:

| Bucket | Meaning | Disposition |
|---|---|---|
| (a) Wrong — asserts AppSync is part of the current architecture | fixed to describe the API Gateway + dispatcher path | 172 lines fixed here; 2 more deferred to #937 and allowlisted with an expiry (1 bucket-(a) allowlist entry) |
| (b) Correct historical reference — describes what the architecture used to be, or names the retained GraphQL schema as a typed contract | left in place, checked that each is unambiguously marked as historical | covered by the 5 enumerated markers, plus 5 bucket-(b) allowlist entries where a marker would be too blunt |
| (c) Legitimate current use — a logical id, path, env var or IAM grant that keeps the old name for compatibility | identifier left alone, nearby prose now explains the name is vestigial | 5 bucket-(c) allowlist entries |

Bucket (c) is things like `APIRESOLVERSTACK` (formerly `APPSYNCSTACK`, for `nested/api-resolvers/`, formerly `nested/appsync/`), the `appsync` install extra in `idp_common`, `APPSYNC_API_URL` (still set on roughly a dozen Lambdas, always to the empty string, which the consuming code reads as "write DynamoDB directly"), and the inert back-compat shim in `docs_service.py` where `is_appsync_mode()` is hardwired to `False` and `mode="appsync"` is accepted and ignored.

Repo-wide (`git grep -ic appsync`, all tracked files, not just documentation) the number of files containing the word falls from 236 to 216. The line count barely moves, 1182 to 1172, and it is worth being explicit about why rather than quoting a flattering figure: this PR *adds* 88 mentions of its own in `check_retired_services.py`, `retired_services.json` and their test file, because a gate for a retired service necessarily names it. Excluding those three new files the count is 1182 → 1084. The ~1000 that remain are source identifiers — `api_adapter.py`, `template_transform.py`, `rerun_processor.py`, the feature-platform test suites, `pyproject.toml` extras — all bucket (c) and deliberately outside the gate's documentation scope.

## Fix

Beyond the mechanical corrections, three things needed more than substitution.

`docs/architecture.md` described the old request path end to end, so it is rewritten rather than patched: the REST route and its authorizer, the dispatcher and its input-shape validation, where authorization actually happens (per-resolver on `identity.claims['cognito:groups']`, `PermissionError` becoming a 403), the polling intervals that replaced subscriptions, the nested-stack list (five stacks, of which three — `PATTERNSTACK`, `APIRESOLVERSTACK`, `MULTIDOCDISCOVERYSTACK` — carry no `Condition` and so exist in every deployment, and two are conditional), and the Lambda Function URL with `InvokeMode=RESPONSE_STREAM` that chat tokens stream from. While there, it now documents the three extension mechanisms the codebase actually has — the pipeline-hooks dispatcher and its seven stages, of which four (`preprocessing`, `postRuleValidation`, `postSummarization`, `postprocessing`) fire in both BDA and pipeline modes because `preprocessing` is the state machine's `StartAt` state and the two branches rejoin at the HITL check; the feature-platform registration and hook mechanism; and the `idp_common` extras-based modular install — plus the two narrower substitution points (`PostProcessingLambdaHookFunctionArn` and `model_id: "LambdaHook"`) that are easily mistaken for them. An architecture document that omits the extension points is missing the part a reader most needs.

`CLAUDE.md`'s service list now names Amazon API Gateway, which resolves the self-contradiction; the accurate nested-stacks explanation is untouched.

`iam-roles/cloudformation-management/IDP-Cloudformation-Service-Role.yaml:126` **deliberately keeps its `appsync:*` grant**. Deleting it would look like completing the sweep but would break the in-place upgrade of a pre-0.6.0 stack, which needs the permission for CloudFormation to delete the GraphQL API that stack created. This is a knowing divergence from item 4 of #929; the README around it now explains that the grant exists only for that upgrade path and that an operator who only ever deploys 0.6.0 or later can drop it.

## Gate

`make check-retired-services` (`scripts/sdlc/check_retired_services.py`) fails when a document presents a retired service as part of the current architecture. It runs from `lint`, `fastlint` and `lint-cicd`, so both CI systems execute it.

The design is a registry plus a justified allowlist, following the precedent of `scripts/security/dep_audit_allowlist.json` and `scripts/srt/issues.json`. `scripts/sdlc/retired_services.json` holds one retired service (AWS AppSync, `removedIn` 0.6.0, with its replacement and its renamed parameters), eight `scannedPaths` globs, seven exclusions each with a justification, five enumerated `historicalMarkers` regexes, and an 11-entry allowlist in which every entry carries a `path`, a narrow `linePattern`, its bucket letter and a written justification. `load_registry` rejects a missing, empty or whitespace-only justification, an unknown bucket letter, and the legacy `expires` key that nothing ever read.

**Marker scope.** The markers exist so that correct historical prose does not need one allowlist entry per sentence. A marker governs **its own sentence**, and a sentence is reconstructed across the Markdown block it sits in, so a claim that wraps over two physical lines still counts as one sentence. Block boundaries — a blank line, a heading, a table row, a thematic break, a fence, a new list item, an unindented line under a bullet, a hard line break — stop the rejoining. Both halves of that rule are load-bearing, and the numbers come from running the previous implementation and this one side by side over the same fixtures:

| Rule | Stale claim next to a historical bullet (24 cases: 8 fixtures × alone / marker above / marker below) | Correct historical prose that wraps (3 cases) |
|---|---|---|
| one-line window either side (previous) | 16 of 24 missed | 2 of 3 falsely reported |
| marker scoped to its own reconstructed sentence (this branch) | 0 missed | 0 falsely reported |

The window was not a theoretical hole: consecutive bullets, one historical and one stale, is how these documents are written, so every genuinely historical sentence in the corpus was exempting the whole line above and the whole line below it. Nor can the rule be tightened to "same physical line" instead — 12 of the 66 marker-exempt lines on this branch have the mention and its retirement verb on different physical lines, and a strict same-line rule reports all 12.

Seven further lines claimed status updates arrive "via GraphQL subscriptions" without naming AppSync at all, so no pattern keyed on the service name could see them. Every call site was checked before rewording: `rest-client.ts` returns an inert subscription handle and the real mechanism is polling at 10s (discovery), 2s (error analyzer, chat) and 5s (quick-start upload), which the prose now states with the interval. The registry pattern matches the *mechanism-claim* forms only — the subscription operations legitimately survive in `schema.graphql`, the codegen output and fenced examples, and matching those would be wrong.

**Expiry is enforced offline.** An allowlist entry may carry `expiresAfterVersion`; `load_registry` compares it against the leading release number in the `VERSION` file and refuses to load once the tree is past it. No API call, so it works in CI and offline. A `.devN` suffix does not count as being past the release, so `0.6.9.dev3` still loads an entry that expires after `0.6.9`.

**Reintroduction.** The registry lists six `AWS::AppSync::*` types that must not reappear in a template, separated in the file into the four that are a measurement (they existed at `a9e318c53`, with the reproduce command recorded) and the two that are forward-looking guards. The check requires an actual `Type:` declaration rather than any occurrence of the string, in both YAML and JSON block form; matching bare occurrences meant the comment explaining the retained `appsync:*` grant tripped it, and only the wildcard spelling of that grant kept the repository green. The prose scan and the reintroduction scan now both always run — the reintroduction report used to return before the prose scan, so a reintroduced resource would have hidden every stale document.

`scripts/sdlc/tests/test_retired_services_not_documented.py` adds 39 tests. Most keep the gate honest rather than re-running it: the scanned surface cannot silently shrink (more than 100 files, five named ones that must be in it, `workshop/` and `CHANGELOG.md` asserted out); a dead allowlist entry fails, because an exemption nobody needs is a standing licence to reintroduce the claim it excused; for **every** non-whole-file allowlist entry, the entry's own `linePattern` is used to read the real matching line out of the real file and a stale claim is planted elsewhere in a rebuilt copy, so the test proves both that the pattern matches text that exists and that the entry does not blanket the file; exactly one entry is allowed to be whole-file and it must say so with `wholeFile: true` (`docs/migration-appsync-to-rest.md`, which is entirely about the migration), and no other entry may use `.*`, `.+`, `""` or a bare service name; correct wrapped historical prose is not reported; a marker on the same line but in a different sentence does not exempt; every bucket-(a) entry must carry an expiry; eleven structural mutations of the registry that would neuter the gate are each rejected by `load_registry`; and the five markers are checked against eight real pre-fix stale lines, copied verbatim from the tree before the fix, so that widening a marker to something like `GraphQL` becomes a reviewable act instead of an invisible one. Each of those eight is driven three ways — alone, with the historical bullet above it, and with the historical bullet below it — which is the 24 cases in the table above.

### Expected interaction with #937

`docs/well-architected.md` is owned by the PR for #937, which is fixing its stale AppSync/WAF claim. I did not edit it. Its one allowlist entry is bucket (a), carries the `issue` link and `expiresAfterVersion: "0.6.9"`, and — importantly — its `linePattern` is pinned to the two literal stale sentences rather than the bare word `appsync`. That matters because PR #951 deletes both lines and adds three new ones that also contain "AppSync": under a bare pattern the exemption would have transferred to three unreviewed lines with the gate's output byte-identical. Pinned, the entry goes dead the moment those two sentences go, and the dead-entry check then fails the build until somebody deletes it.

### Expected interaction with #960

#960 adds a sentence to `docs/threat-model.md` recording that the document once described an AppSync API "that had been removed several releases earlier". The markers cover `replaced`, `superseded` and `retired` as well as `removed`, and tolerate Markdown emphasis on the verb (`were **removed** in v0.6` occurs in this branch's own prose), so that sentence passes on either branch and on the merge of the two. Verified by a local `git merge --no-commit --no-ff` of `github/docs/threat-model-linked-and-gated` into this branch: the gate exits 0 over 216 files, and the merge was then aborted, not committed. The registry's `security/threat-modeling/**` exclusion now records that it should be revisited once #960 lands, since a document with its own currency gate should not be exempt from this one.

### CI parity

`SHARED_GATES` in `scripts/tests/test_ci_gate_parity.py` did **not** need editing: it already contains `make lint-cicd`, so a gate invoked from that recipe runs in both CIs automatically. The list that did need the new target is the parametrize of `test_lint_cicd_covers_what_local_lint_covers`, which greps the `lint-cicd` recipe slice for each gate local `make lint` runs; without `check-retired-services` there it would have failed.

## Testing

| Command | Result |
|---|---|
| `make check-retired-services` | pass — 215 files scanned, zero findings |
| `python scripts/sdlc/check_retired_services.py --root <pristine c33e0725d^>` | **exit 1, 172 findings across 47 files** — the pre-fix failure proof |
| `python -m pytest scripts/sdlc/tests/test_retired_services_not_documented.py` | pass — 39 passed |
| `python -m pytest scripts/sdlc/` | pass — 605 passed |
| `python -m pytest scripts/tests/` | pass — 431 passed, 3 skipped |
| `make lint-cicd` | pass — All code quality checks passed, including the new retired-service check |
| `make typecheck-pr TARGET_BRANCH=github/develop` | pass — 0 errors, 0 warnings, 0 notes; confirmed it does select both changed Python files |
| `ruff check --no-force-exclude` / `ruff format --check --no-force-exclude` | pass — All checks passed; 2 files already formatted |

A caveat on the Python lint gates, because it would be misleading to present them as evidence. `ruff.toml` line 29 puts the bare name `scripts` in `extend-exclude`, and a bare name matches at any depth, so `make ruff-lint` and `make format` never reach anything under `scripts/` — the two files this PR adds included. `ruff check --show-files <explicit path>` does *not* reveal this, because an explicit path bypasses the exclusion; `--no-force-exclude` is the honest probe, and that is what the table reports. Of the repository's static-analysis gates only `make typecheck-pr` actually reads `scripts/`, so the gate that enforces documentation accuracy is itself written in the one directory the linters skip. That exclusion is pre-existing (287 of 1114 tracked Python files sit under some bare-name excluded directory) and fixing it has a far wider blast radius than this PR, so it is left alone here — but the two new files were reviewed by hand rather than on the strength of a green lint run.

## Follow-ups noticed but out of scope

Several stale claims surfaced during the sweep that are code rather than documentation, or are functional gaps rather than wording:

- `src/lambda/chat_stream_processor/app.py`'s docstring says the role is granted `lambda:InvokeFunctionUrl`, but `template.yaml:11193-11198` records that `InvokeFunctionUrl` alone returns 403 and the grant is `lambda:InvokeFunction`.
- `lib/idp_common_pkg/idp_common/dynamodb/__init__.py`'s docstring says "bypassing AppSync GraphQL API" in the present tense.
- `lib/idp_common_pkg/idp_common/synthesis/appsync_status.py` is an honest no-op shim, but it stands in for a real gap: synthesis and bootstrap job status is never reported to the UI.
- `schema.graphql` still declares `@aws_subscribe` fields and the UI still generates subscription documents that nothing serves.
- The `scripts` exclusion in `ruff.toml` described above.

## Scope

Documentation and comments only, plus the new gate and its tests. No AWS API was called, `publish.py` was not run, nothing was merged, approved or closed, and `workshop/` is untouched (it is referenced externally and is excluded in the registry with that justification).


rstrahan:
Heads-up from #951 (`docs/well-architected-refresh`, fixes #937), which is the change your
temporary allowlist entry is waiting on. **Please do not act on that entry's `expires` field
literally — deleting it will red-line this gate on `develop`.**

The entry in `scripts/sdlc/retired_services.json` is:

```
{
  "path": "docs/well-architected.md",
  "linePattern": "appsync",
  "bucket": "a",
  "justification": "KNOWN STALE, owned by a different change ... remove it once #937 lands.",
  "expires": "when #937 merges"
}
```

#951 does fix the stale bucket-(a) claim that entry was created for — the old line 43 saying
WAF protects the AppSync GraphQL API is gone. But the rewritten file still mentions AppSync,
now at `docs/well-architected.md:183`:

> The API Gateway REST transport replaced AWS AppSync entirely — there are no
> `AWS::AppSync` resources in any template — see [AppSync to REST migration](./migration-appsync-to-rest.md).

That is a legitimate **bucket (b)** historical mention, of exactly the kind your registry is
designed to allowlist, not a claim that the service is present.

I ran your gate from a scratch copy outside my worktree (`--registry` / `--root` overrides,
your files untouched) against #951's branch:

| Registry state | `docs/well-architected.md` findings |
|---|---|
| entry present, as it stands on your branch | **none** — the broad `"appsync"` `linePattern` excuses the whole file |
| entry removed, i.e. acting on `expires` | **1** — `docs/well-architected.md:183`, the historical sentence above |

So nothing breaks when #951 merges, and your stale-entry check won't fire either, because the
broad `linePattern` keeps matching. The failure appears only when someone later reads
`expires` and deletes the entry.

**Suggested action after #951 merges:** re-justify the entry into bucket (b) rather than
removing it. Alternatively, narrow its `linePattern` to that one sentence and add a
`historicalMarkers` pattern covering the "replaced AWS AppSync entirely" phrasing — worth
noting that line 183 currently matches **none** of your existing markers, which is why it
genuinely needs an allowlist entry rather than relying on a marker. Either way the
justification text and `expires` should stop saying "remove it once #937 lands".

I have deliberately not touched `scripts/sdlc/retired_services.json` — this PR owns it. For
the record, #959 changes no file that #951 touches (`docs/well-architected.md` is not in
#959's changed-file list), so the two do not conflict. The same note is in #951's description.

PR #1010, opened 2026-09-18
[PARKED — next release] test(lambda): make the offline suites independent of an ambient AWS region

### LINKED ISSUE #988: test: three src/lambda suites build boto3 clients at import time with no region, so they pass locally and fail in CI
## Summary

Three test suites under `src/lambda/` construct a boto3 client at import time without a region, so they pass on a developer machine — which has an ambient region, usually from `~/.aws/config` — and fail on any CI runner, which has none. They were invisible until now because none of the three ran in either CI; PR #944 adds them to `test-packages-cicd` and had to pin a region to make them pass.

This is **not** a defect in the handlers. The Lambda runtime always sets `AWS_REGION`, so the module-level clients are fine in production. It is a test-harness assumption, and the consequence is that a suite's result depends on the machine it runs on.

## Measurement

Each suite was run from its own directory with every `AWS_*` variable stripped and `HOME` pointed at an empty directory, so no ambient region or credential could leak in from `~/.aws/config` — i.e. the environment a GitHub runner actually has. Six of the nine directories PR #944 adds are clean; three are not.

| Suite | clean env, no region | + `AWS_DEFAULT_REGION` only |
|---|---|---|
| `src/lambda/external_idp_group_mapping` | 3 failed, 5 passed, **41 errors** | 3 failed, 46 passed |
| `src/lambda/save_reporting_data` | **1 collection error** (0 tests run) | 8 passed |
| `src/lambda/test_file_copier` | **7 failed** | 7 passed |
| the other six directories | all pass | all pass |

Every failure resolves to one exception, `botocore.exceptions.NoRegionError: You must specify a region`, raised from `botocore/regions.py:285`. No credentials are required — a region alone is sufficient — so this is purely about region resolution.

`external_idp_group_mapping` has a second, distinct instance of the same root cause. Its `TestGroupMapping` tests use `patch.dict(os.environ, env, clear=True)` and then `importlib.reload(index)`, which re-executes `index.py:41`'s `boto3.client("cognito-idp")`. Because `clear=True` wipes the entire environment, an `AWS_DEFAULT_REGION` supplied by the caller is removed before the client is built — so those three tests fail even when a region *is* provided externally, and pass only when the region comes from `~/.aws/config`, which `clear=True` cannot reach. That is why supplying a region fixes 41 of the errors but leaves 3 failures.

## What PR #944 did, and what it left

PR #944 needed these suites green to add them to CI at all, so it applied the narrow fix:

- `AWS_DEFAULT_REGION=$(TEST_AWS_REGION)` on the three `test-packages-cicd` recipe lines that need it, with `TEST_AWS_REGION ?= us-east-1` defined once in the `Makefile`.
- The region added back into the three `patch.dict(..., clear=True)` dictionaries in `external_idp_group_mapping/test_index.py`, which cannot affect what those tests assert (the mapping is built from the `*_GROUP_NAME` variables only).

What it did **not** do, and what this issue is for: the suites themselves are still not import-safe. Anyone running `pytest` directly in those three directories with no ambient region still gets `NoRegionError`, and `make test` (via `scripts/run_all_tests.py`) does not set the variable either — it works today only because developer machines have a region. The fix belongs in the suites, not in the caller.

## Suggested remediation

1. Make each of the three suites import-safe, so no caller has to know. A `conftest.py` doing `os.environ.setdefault("AWS_DEFAULT_REGION", "us-east-1")` before collection is the smallest version; making the handlers construct their clients lazily is the more thorough one and also helps anyone unit-testing them in future.
2. Once (1) lands, drop the `AWS_DEFAULT_REGION=` prefixes from the `test-packages-cicd` recipe and the `TEST_AWS_REGION` variable, so the fix lives in one place.
3. Audit the rest of the repository the same way. Only the nine directories PR #944 touched were measured, and only for this one variable. The generalisable check is to run a suite with a deliberately empty environment rather than an inherited one — a green run on a developer machine does not establish that a suite is hermetic. This is the same shape as #980 (a gate that is correct but never consulted) and #974.

## References

- `Makefile`, `test-packages-cicd` — the three pinned recipe lines and the comment recording why.
- `src/lambda/external_idp_group_mapping/test_index.py`, `TestGroupMapping` — the `clear=True` + `reload` case.
- `src/lambda/save_reporting_data/test_index.py`, `src/lambda/test_file_copier/test_index.py` — the collection-time cases.
- PR #944 — where this was found, while adding these suites to CI.
- #980, #974 — the CI-coverage gap that kept these suites out of CI, so the fragility was never exposed.

rstrahan:
Closes #988.

## What the issue said, and the wider cause it did not have

Three suites under `src/lambda/` failed under a clean environment with `NoRegionError`. Reproduced before changing anything, from each suite's own directory under `env -i` with an empty `HOME` — the environment a CI runner actually has: `external_idp_group_mapping` 8 passed / 41 errors, `save_reporting_data` 1 collection error with 0 tests run, `test_file_copier` 7 failed.

**One of the three was not a test-harness problem at all.** `save_reporting_data`'s traceback runs `index.py` → `idp_common.reporting` → `idp_common.s3` → `idp_common.utils` → `idp_common/utils/settings_helper.py:22`, which was `ssm_client = boto3.client("ssm")` at module scope. Because `idp_common.utils` is imported by `idp_common.s3`, **importing the shared library at all required a resolvable region — for every consumer in the repo**, not just this suite. That is fixed at the cause: the client is now built on first use and memoized, matching the sibling `idp_common/monitoring/settings_cache.py` which has always been lazy. After that change `save_reporting_data` passes in a clean environment with **no conftest at all**.

The other two keep their module-scope clients, which is correct for a Lambda handler — the client is reused across warm invocations — and get a `conftest.py` pinning fake credentials and a region, matching seven sibling conftests already in the tree. Converting those handlers would also have meant editing the duplicated InlineCode copy inside `template.yaml`, which has its own drift test, for no production benefit.

## The Makefile change, which is the part to review hardest

The three `AWS_DEFAULT_REGION` pins are **removed**, not kept. A pin means CI never exercises the hermetic path, so a regression in a suite's conftest would be invisible — the gate would be weaker than the environment it claims to simulate.

In their place the recipe does the opposite. `HERMETIC_AWS` unsets every `AWS_*` variable and points `AWS_CONFIG_FILE` and `AWS_SHARED_CREDENTIALS_FILE` at an empty file, and **all 34 pytest invocations** in `test-packages-cicd` go through `PYTEST_HERMETIC`. So the local run equals the CI run instead of being weaker, and a newly added recipe line is covered automatically. Stripping credentials as well as the region is deliberate: a suite in this target that reaches a real endpoint should fail loudly rather than transact against whichever account the developer happens to be signed in to.

## The sweep, complete rather than sampled

All 34 invocations (31 distinct directories) were run in full — not collect-only — each from its own working directory under `env -i` with an empty `HOME`. Total 277 s. Exactly three failed: the three known ones. The issue's "six of nine were clean" is confirmed by measurement (`api_handler` 35, `batch_pre_processor` 17, `complete_section_review` 6, `job_tracker` 11, `user_management` 17, `version_check_resolver` 7), and the sweep extends it — no other suite in the recipe is region-dependent.

## The guard

`scripts/tests/test_offline_suites_are_hermetic.py` parses `HERMETIC_AWS` out of the Makefile at test time and builds its probe environment from that, so the environment it tests under is by construction the one the recipe uses. No suite name is hardcoded anywhere in it. Five assertions:

1. the wrapper parses, unsets something, assigns something, and `PYTEST_HERMETIC` still routes through it;
2. **the wrapper is measured, not trusted** — a subprocess under it must fail to build a client with `NoRegionError`. This is the anti-vacuity check: a stripping wrapper that silently stopped stripping would turn the whole gate into a no-op;
3. no recipe line sets an `AWS_*` variable for the suite it runs (the old pins would fail this);
4. every recipe line running pytest goes through `$(PYTEST_HERMETIC)`;
5. every invocation the recipe names collects under the stripped environment — 34 parametrized cases.

**A finding that invalidates the obvious probe, recorded in a comment:** S3 cannot be used to detect a missing region. Under the full wrapper `boto3.client("s3")` resolves `us-east-1` from botocore's legacy global default, while `boto3.client("ssm")` raises. The first draft used S3 and failed for that reason.

**Stated scope limit,** in the module docstring rather than left implicit: collection catches a client built at import time, which is the shape all three instances had and the only one that can take out a whole suite with a collection error. A client built inside a test body is caught instead by the recipe running the suite under the same wrapper. Probing all 34 suites by running them would cost 277 s against 27 s, which is the only reason the probe stops at collection.

## Mutation probes — both halves of the control

- **Import-time:** reinstating the eager client in `settings_helper.py` failed exactly one case, `test_suite_collects_without_an_aws_region[src/lambda/save_reporting_data]` (1 failed, 38 passed), with the traceback and remediation hint. Restored → 39 passed.
- **Run-time:** deleting `src/lambda/test_file_copier/conftest.py` left the collect probe **passing** — the documented limitation, honestly reproduced — while the recipe's own hermetic run of that suite failed with 7 failed. So each half catches the case it claims and neither is doing nothing.

## Verification

- `make test-packages-cicd`: exit 0. The three suites now report 49, 8 and 7 passed.
- Clean-environment acceptance, from each suite's own directory under `env -i`: **49, 8, 7 passed**.
- `scripts/tests` + `scripts/sdlc/tests`: **1,756 passed, 7 skipped**
- `make test-cicd -C lib/idp_common_pkg`: **6,185 passed, 13 skipped** (run because the shared library changed)
- `make typecheck-pr`: 0 errors, 4 warnings — one on `settings_helper.py`, confirmed pre-existing against `git show HEAD:` of the same file
- `make ruff-lint`, `make format`: clean

## Flagged rather than claimed complete

`scripts/run_all_tests.py`, which backs `make test`, does **not** strip the AWS environment. The suites are self-contained now so `make test` works either way, but enforcement lives only on the `test-packages-cicd` path — the one whose result gates a merge. `run_all_tests.py` discovers a superset of roots that have not been measured under a stripped environment, so widening it there is a separate change.

Two measurement notes: under the Makefile wrapper `lib/idp_sdk` reports 490 passed where the `env -i` sweep reported 485 passed / 5 skipped, and `lib/idp_feature_sdk` 140 against 119 passed / 21 skipped. `env -i` removes more than the AWS variables, so those suites skip for unrelated environmental reasons under it; the wrapper is stricter for AWS purposes and looser for everything else, which is what is wanted. Which variable drives those skips was not chased.


rstrahan:
**Parked — not for this release.** Deferred to the next release cycle so `develop` can be frozen for stack testing. This is test-harness hygiene, not a functional or security fix, so nothing in it needs to ship now.

Two concrete defects from review are outstanding and both are small:

1. `scripts/tests/test_offline_suites_are_hermetic.py` — a loop variable named `token` compared to a string literal trips Bandit `B105` (`hardcoded_password_string`). SRT rates that HIGH, so the `SRT Security Review` gate fails. Renaming it to `word` clears the gate and leaves only the Low-severity findings already tolerated across other files in `scripts/tests/`. These are shell words, not security tokens, so the name was slightly wrong regardless.
2. `Makefile` — `HERMETIC_AWS` unsets `AWS_PROFILE` but not `AWS_DEFAULT_PROFILE`, which botocore also resolves its profile from. On a machine exporting that variable the wrapper produces `ProfileNotFound` rather than `NoRegionError`, which red-lines the merge-gating target. Add `-u AWS_DEFAULT_PROFILE`.

A third failure is **not attributed to this PR**. `Lint, Type Check, and Test` reports five failures in `lib/idp_common_pkg/tests/unit/test_test_set_resolver.py` with `AttributeError: 'NoneType' object has no attribute 'access_key'`. That file is untouched by this branch and unchanged since 2026-09-15. I could not reproduce the failures in five separate hermetic attempts — the file alone, the whole `tests/unit` tree, and CI's exact `pytest -m "not integration" ... tests/` invocation, each with every `AWS_*` variable unset and the config and credentials files pointed at `/dev/null`. All passed, including 6,185 passed / 0 failed under CI's own command. The remaining differences are environmental: CI does a fresh `pip install -e ".[test]"`, runs `typeguard` 4.6.0 against 4.5.2 locally, passes `--cov` flags, and collects a different set (17 deselected versus 23 locally). The same job also failed on an unrelated branch whose earlier run passed, which points away from this change.

Everything else the review checked held up: the sweep of all 34 pytest invocations, the clean-environment acceptance results (49/8/7 passed), both mutation probes, and the finding that the shared library's module-scope SSM client made *importing* `idp_common` require a resolvable region for every consumer — which is the substantive fix here and worth keeping.

No action needed from reviewers until this is picked back up.

